PR #23, opened 2016-01-24
Added remove textures stage

leerichard42:


lib/removeUnusedTextures.js
```diff
@@ -0,0 +1,56 @@
...
+var defined = Cesium.defined;
+
+module.exports = removeUnusedTextures;
+
+function removeUnusedTextures(gltf, stats) {
+    var usedTextureIds = {};
+    var materials = gltf.materials;
+
+    // Build hash of used textures by iterating through materials
+    if (defined(materials)) {
```
pjcozzi: In addition to checking materials, we also needs to check techniques since a technique can provide a default value for a parameter.  In object-oriented terms, think of a technique as a class, and as a material as an instance of the class.

See https://github.com/KhronosGroup/glTF/blob/master/specification/README.md#techniqueparameters-1

In practice, it is unlikely that a technique will define the default value for a texture (well, at least for a diffuse texture), but it could happen.  Here is an example of a default color (not texture) defined in a technique:

https://github.com/KhronosGroup/glTF/blob/master/sampleModels/duck/glTF/duck.gltf#L287-L294

leerichard42:
Updated - I also added two extra tests with techniques, but not sure if there's anywhere I can cut down duplicated code in the technique check?

pjcozzi:
Looks good!  I merged in master.  The duplicate is not too bad, I would not worry about it.

PR #24, opened 2016-01-24
Added remove meshes stage

leerichard42:


specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf
```diff
@@ -159,6 +159,14 @@
                     "mode": 4
                 }
             ]
+        },
+        "unusedMeshId": {
+            "name": "Mesh",
```
pjcozzi: `name` is optional so I would remove here and in any other new examples.

specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf
```diff
@@ -159,6 +159,14 @@
...
+            "primitives": [
```
pjcozzi: `primitives` is also optional so we could just make this an empty object.

See https://github.com/KhronosGroup/glTF/blob/master/specification/README.md#reference-mesh

pjcozzi:
Just those minor comments.

leerichard42:
Updated!

pjcozzi:
Thanks!

PR #26, opened 2016-01-26
Added remove animation samplers stage

leerichard42:


pjcozzi:
I'm not sure that this will be useful in practice.  I don't know glTF assets will often be generated with unused animation samplers, and even if they are, these are not heavy objects - or do not point to heavy objects - that could be removed like buffers, images, etc.

What do you think?

leerichard42:
That makes sense! Should we just delete this branch?

pjcozzi:
Yes, I think so.

PR #29, opened 2016-01-31
Code cleanup

### LINKED ISSUE #18: Cleanup
I can do these at the code sprint if we don't get to them sooner.

* [x] Move `OptimizationStatistics.prototype.print` to a function in `bin/gltf-pipeline.js` since it doesn't make sense in the context of the browserify build.
* [x] To reduce code duplication, create a helper function for at least this part of unused object removal:

https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/removeUnusedImages.js#L22-L42

This will require replacing properties like `stats.numberOfImagesRemoved` with something like `stats.numberRemoved['image']`.

pjcozzi:
@leerichard42 for your review.

Fixes #18.


pjcozzi:
@leerichard42 I added the tests as we discussed offline.  Let's merge this before #30.

pjcozzi:
Merged as part of #30.

PR #31, opened 2016-02-01
Added parseGLTF and loaded shaders

leerichard42:


bin/gltf-pipeline.js
```diff
@@ -38,7 +39,7 @@ fs.readFile(gltfPath, function (err, data) {
         throw err;
     }
 
-    var gltf = JSON.parse(data);
+    var gltf = parseGLTF(path.dirname(gltfPath), data);
```
pjcozzi: To follow camel case naming conventions and to better show that this is to create our in-memory runtime data structure for glTF, let's call this `loadGltfUris`.

Let's also do the `JSON.parse` before calling the function, so the function should take a glTF object, not a string.

lib/parseGLTF.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+var fs = require('fs');
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+
+module.exports = parseGLTF;
+
+function parseGLTF(dir, gltfData, callback) {
```
pjcozzi: Don't use abbreviations in public APIs, rename `dir` to `path`.

lib/parseGLTF.js
```diff
@@ -0,0 +1,38 @@
...
+    var gltf = JSON.parse(gltfData);
+    gltf.loadedShaders = {};
+
+    var shaders = gltf.shaders;
+    if (defined(shaders)) {
+        for (var shaderId in shaders) {
```
pjcozzi: For a short function like this where there is only one "id", it is generally better style to use `id` instead of `shaderId`.

lib/parseGLTF.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+function loadURI(gltf, dir, uri, loadId, callback) {
+    var uriPath = dir + '\\' + uri;
+    fs.readFile(uriPath, function (err, data) {
+        if (err) {
+            throw err;
+        }
+
+        gltf[loadId][uri] = data;
```
pjcozzi: I thought more about this: even though we don't have to, let's put this in an `extras` object, e.g.,
```
shader.extras.source
```
See https://github.com/KhronosGroup/glTF/blob/master/specification/README.md#reference-shader

specs/lib/parseGLTFSpec.js
```diff
@@ -0,0 +1,64 @@
...
+                    '"type": 35632,' +
+                    '"uri": "CesiumTexturedBoxTest0FS.glsl"' +
+                '},' +
+                '"CesiumTexturedBoxTest0VS": {' +
+                    '"type": 35633,' +
+                    '"uri": "CesiumTexturedBoxTest0VS.glsl"' +
+                '}' +
+            '}' +
+        '}';
+        var frag = 'precision highp float;\n' +
```
pjcozzi: Instead of storing these strings, load `CesiumTexturedBoxTest0FS.glsl` or `CesiumTexturedBoxTest0VS.glsl` (no need to test both) from disk just like the test you wrote on Sunday.

lib/parseGLTF.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    return gltf;
+}
+
+function loadURI(gltf, dir, uri, loadId, callback) {
+    var uriPath = dir + '\\' + uri;
+    fs.readFile(uriPath, function (err, data) {
+        if (err) {
+            throw err;
```
pjcozzi: Also add a test for this case.

specs/lib/parseGLTFSpec.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+var parseGLTF = require('../../lib/parseGLTF');
+
+describe('parseGLTF', function() {
```
pjcozzi: Also add a test for when the shader is a data uri, e.g., like this:

https://github.com/KhronosGroup/glTF/blob/master/sampleModels/box/glTF-Embedded/box.gltf#L202-L210

If `fs.readFile` does not support data uris (it might not given its name), see if there is a built-in way to handle them in node or a library on npm; otherwise, check out how we do it client-side: https://github.com/AnalyticalGraphicsInc/gltf-utilities

pjcozzi:
@leerichard42 good start, the next step is to add a new function
```
writeGltf
```
to replace the following code and write out the data in the `extras` object (and `delete` those objects and update the `uri` properties if needed).

https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/bin/gltf-pipeline.js#L60-L64

To start, this function could take a boolean for either writing all the `extras` data to separate external files (like `.glsl` for shaders) or to write data uris embedded in the glTF.  This is how, for example, a user can use glTF pipeline to convert an asset to/from embedded data uris.

Later, we'll replace the boolean with more fine-grained info, e.g., embed shaders, but not geometry and textures.

lib/loadGltfUris.js
```diff
@@ -0,0 +1,52 @@
...
+
+function loadURI(gltf, path, extras, uri, callback) {
+    var uriPath = path + '\\' + uri;
+    // fs.readFile('./specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTestError.glsl', function (err, data) {
+
+    fs.readFile(uriPath, function (err, data) {
+        if (err) {
+            // throw new Error("Unexpected error!");
+            throw err;
+            callback();
```
pjcozzi: Check out `process.nextTick` (it is in the front of the Node.js book I suggested) and use it whenever calling a callback.

lib/loadGltfUris.js
```diff
@@ -0,0 +1,52 @@
...
+        for (var id in shaders) {
+            if (shaders.hasOwnProperty(id)) {
+                var shader = shaders[id];
+                // process.stdout.write('SHADER: ' + shader + '\n');
+                // var extras = shader.extras;
+                // if (!defined(extras)) {
+                //     shader.extras = {};
+                // }
+                // extras = shader.extras;
+                shader.extras = shader.extras || {};
```
leerichard42: Is this line okay in terms of style, or should I use the commented version instead?

leerichard42:
Updated and jasmine tests pass locally for me, but CI build fails - the error seems to be in the test file read, so could it be the fragmentShader file path?

lib/loadGltfUris.js
```diff
@@ -0,0 +1,61 @@
...
+
+module.exports = loadGltfUris;
+
+function loadGltfUris(path, gltf, callback) {
+    var shaders = gltf.shaders;
+    if (defined(shaders)) {
+        for (var id in shaders) {
+            if (shaders.hasOwnProperty(id)) {
+                var shader = shaders[id];
+                shader.extras = shader.extras || {};
```
pjcozzi: Use Cesium's defaultValue, e.g.,
```javascript
// At the top of the file:
var defaultValue = Cesium.defaultValue;

// Then
shader.extras = defaultValue(shader.extras, {});

File: lib/loadGltfUris.js [hunk 11]
```diff
@@ -0,0 +1,61 @@
...
+                loadURI(gltf, path, shader.extras, uri, callback);
+            }
+        }
+    }
+
+    return gltf;
+}
+
+function loadURI(gltf, path, extras, uri, callback) {
+    if (!/^data\:/i.test(uri)) {
```
pjcozzi: Create a separate function in a separate file (and separate test) for `isDataUri` since we will use this in a few places.

lib/loadGltfUris.js
```diff
@@ -0,0 +1,61 @@
...
+        var uriPath = path + '\\' + uri;
+        fs.readFile(uriPath, function (err, data) {
+            if (err) {
+                if (callback) {
```
pjcozzi: Why call `callback` here when there is an error?  Why not throw?

lib/loadGltfUris.js
```diff
@@ -0,0 +1,61 @@
...
+                    process.nextTick(function() {
+                        callback();
+                    });
+                }
+            }
+        });
+    }
+    else {
+        var buffer = dataUriToBuffer(uri);
```
pjcozzi: Just
```javascript
extras.source = dataUriToBuffer(uri);
```
is better style; it is concise.

specs/lib/loadGltfUrisSpec.js
```diff
@@ -0,0 +1,73 @@
...
+        var gltf = {
+            "shaders": {
+                "CesiumTexturedBoxTest0FS": {
+                    "type": 35632,
+                    "uri": "CesiumTexturedBoxTest0FS.glsl"
+                }
+            }
+        };
+        
+        fs.readFile(fragmentShaderPath, function (err, fragmentShaderData) {
```
pjcozzi: This can be done in a `beforeAll` and assign it to a variable scoped for the lifetime of the test suite, e.g., 
```javascript
describe('loadGltfUris', function() {
    var fragmentShaderData;

    beforeAll(/* ... */);

    it(/* ... */);
    it(/* ... */);
    // ...
});

File: lib/loadGltfUris.js [hunk 15]
```diff
@@ -0,0 +1,61 @@
+'use strict';
+var fs = require('fs');
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+var dataUriToBuffer = require('data-uri-to-buffer');
+
+module.exports = loadGltfUris;
+
+function loadGltfUris(path, gltf, callback) {
```
pjcozzi: I would reorder the arguments to:
```
gltf, basePath, callback
```

lib/loadGltfUris.js
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+    }
+
+    return gltf;
+}
+
+function loadURI(gltf, path, extras, uri, callback) {
+    if (!/^data\:/i.test(uri)) {
+        var uriPath = path + '\\' + uri;
```
pjcozzi: Maybe this is why the test is failing.  Don't join paths by hand.  Use [path.join](https://nodejs.org/api/path.html).

lib/loadGltfUris.js
leerichard42: This was the only way I was able to get the jasmine test to recognize the error without the test stopping at the throw statement - I attempted to separate out the function that throws the exception so I could call it directly from the testing file as we discussed online, but I might have misinterpreted what you meant.

leerichard42:
Updated! The jasmine-node package only included up to version 1.3.1 of jasmine, so I installed the latest jasmine package in order to use beforeAll. I had a little confusion with the installation and CI build -  sorry for all the commits!

leerichard42:
Actually, can you leave this unmerged for now? I'm making some changes because there are some issues with the asynchronicity that I'm trying to fix.

leerichard42:
Updated, installed the async package which fixed some problems and made the callbacks much cleaner.

bin/gltf-pipeline.js
```diff
@@ -39,29 +41,31 @@ fs.readFile(gltfPath, function (err, data) {
     }
 
     var gltf = JSON.parse(data);
-    var stats = new OptimizationStatistics();
-
-    addDefaults(gltf, stats);
+    gltf = loadGltfUris(gltf, path.dirname(gltfPath), function() {
+        var stats = new OptimizationStatistics();
```
pjcozzi: Are you using 8 spaces for tabs?  The rest of the code uses 4.

lib/loadGltfUris.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    return gltf;
+}
+
+function loadURI(gltf, basePath, extras, uri, asyncCallback) {
+    //Load the uri into the extras object based on the uri type
+    if (isDataUri(uri)) {
+        extras.source = dataUriToBuffer(uri);
+        asyncCallback();
```
pjcozzi: Whenever you call a callback use `process.nextTick`.  Did you check out the Node book on this topic?

lib/loadGltfUris.js
```diff
@@ -0,0 +1,59 @@
...
+    var shaders = gltf.shaders;
+
+    //Iterate through each shader and load its uri
+    if (defined(shaders)) {
+        var shaderIds = Object.keys(shaders);
+        async.each(shaderIds, function(shaderId, asyncCallback) {
+            var shader = shaders[shaderId];
+            var uri = shader.uri;
+            shader.extras = defaultValue(shader.extras, {});
+            loadURI(gltf, basePath, shader.extras, uri, function() {
```
pjcozzi: Why create the anonymous function here?  Why not just pass `asyncCallback` (and also just name it `callback`)?

lib/loadGltfUris.js
```diff
@@ -0,0 +1,59 @@
...
+
+module.exports = loadGltfUris;
+
+function loadGltfUris(gltf, basePath, callback) {
+    var shaders = gltf.shaders;
+
+    //Iterate through each shader and load its uri
+    if (defined(shaders)) {
+        var shaderIds = Object.keys(shaders);
+        async.each(shaderIds, function(shaderId, asyncCallback) {
```
pjcozzi: Why exactly is this code using `async.each`?  Why not go through each property in `shaders` as we did for the removal optimizations?  I wouldn't bother adding parallelism yet as this might not even be the right granularity.

pjcozzi: Ah, I see-ish.  So we only want to call `callback` that is passed into this function once after all the shaders are loaded.  Nice!

lib/loadGltfUris.js
pjcozzi: This comment stands, but not the rename.

lib/loadGltfUris.js
pjcozzi: I know this is in a helper function, but we should still do it.

lib/isDataUri.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = isDataUri;
+
+/**
+ * @private
+ */
+ function isDataUri(uri) {
```
pjcozzi: What happens if `uri` is undefined?  We could throw an exception, but I think it would also be OK to just make sure this returns `false`.

specs/lib/loadGltfUrisSpec.js
```diff
@@ -0,0 +1,75 @@
...
+        fs.readFile(fragmentShaderPath, function (err, data) {
+            if (err) {
+                throw err;
+            }
+            fragmentShaderData = data;
+            done();
+        });
+    });
+    
+    it('loads an external shader', function(done) {
```
pjcozzi: To be extra careful testing the async code, I think it would be useful to add another test with two `shaders` (could be the same GLSL source): one that points to a .glsl file and another with a data uri.

pjcozzi:
@leerichard42 nice job here.  I know the async stuff can be confusing (this PR confused me a bit), but it is a great hump to get over!

bin/gltf-pipeline.js
pjcozzi: Ah, I see, it is in the callback.

leerichard42:
Updated!

.gitignore
```diff
@@ -4,6 +4,8 @@ npm-debug.log
 Thumbs.db
 /coverage
 output.gltf
+CesiumTexturedBoxTest0FS.glsl
```
pjcozzi: I would rather ignore a single `output` directory that we commonly use for testing instead of the individual `output.gltf` and its resources.  Can you update this in #32 if you haven't already?

bin/gltf-pipeline.js
```diff
@@ -39,29 +41,35 @@ fs.readFile(gltfPath, function (err, data) {
     }
 
     var gltf = JSON.parse(data);
-    var stats = new OptimizationStatistics();
+    gltf = loadGltfUris(gltf, path.dirname(gltfPath), function(loadError) {
```
pjcozzi: I believe the common Node.js practice is to name this `err`, not `loadError`.  Any recent to deviate here?

leerichard42: I wasn't sure if there would be a problem since I have multiple `err` objects passed back in callbacks, but I will change it back!

.gitignore
leerichard42: Okay, will do.

specs/lib/isDataUriSpec.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+describe('isDataUri', function() {
+    var isDataUri = require('../../lib/isDataUri');
```
pjcozzi: I tightened up the scope here.  Minor, but a good practice.  a3d64b64d499f6a65f49d8983f8d7885fcf99fd8

bin/gltf-pipeline.js
pjcozzi: I'll change this now, no need to hold you up.

pjcozzi: Sorry, didn't see that recent comment.  Yes, I saw that there is `err` in an outer scope and in another inner scope.  In this particular case, I would not worry about it; the current scope overrides, of course.

pjcozzi:
Nice work as always!

PR #34, opened 2016-02-17
Added load image stage

leerichard42:


specs/lib/loadImageUrisSpec.js
```diff
@@ -0,0 +1,88 @@
...
...DurVm/ouJZueb5cRlfPj+t6qt3/8KstyEdxLunI6hHFv+oLu8BGTN7i8kZKi0+xvnu6VfaBpqRIOKDXpvy+OKf2Q2FMjtd2ZUHAOAF9lDr70+07kKyiHk/Y5+1pv4HWsosf3evBOXd4lJv467TGzmeRfqIEQvLH1g28ykMKGIDh4KUBwAYdLXtbPr3EWcPUom0aFT6B+o3FNnmK2iIoyjlAQAYLnD04z+c+LgByUUqCuwVD9S9qKdTlTW4VprybnFzrHnnqX93+8aRbqKBwIl75jxZO/0BBc7bK1R5AAA/49574T9auo4hAYkj11bypZoNqdpcZVZPucq7xfWhpr0fve70jCAlRQ5NqlfMXTcn9wtKXqJUuvIAAAwXbGrfefTynzgeLbWFpzxvwfKKZwyqNIXXMwGUd4sJ/8D+C69f6zuHtHUn7JZpK2u+mWUqTYjaJozybnFjuOlg82Y07fIpNGrjPbOfrMy+K4EOWyWY8gAAvMBfHTpx+OLbDlc/0pya1i2d9XDVtHspQpNYNU885X2iP+7qwLHGy1vGXANTU3MqSrO4fE1N/kqa0CZi/RNVebf119J35HjrzmFn99TRnEGbuqjsbypz7lYRusR9i8RW3i0Ege+ZuHL2xp6W7pPJfX+93ZK3uOyhEvsCaVPZIeVFizvouND1/qlre/wBVzIJjibVswuWVuUtTzfOSJpTBEmlvFswXLBj5OzlnmMfdzcl+s6XPFtZdeHywvR5qsQM5qaW8m4T5Hwdw+cudR291nsusQ585NpKKvOWFmfU6+m0ZG2dZFbebQKM+8bIufaB89f7z/uU2hFTpKowc05xZk2+bW7scoch5ck0FgH8qLen23GlffCj6wMXGCYgb30InJxmK51mK5tuq8w0lcTtfD9SnpxwPDvs7nB4eoacnb2O632jbQwbjLm5MdxqyspOLcxIKcxOKbEZ8uN/0BUpT3Hu0BUYHnZ1jbi6Xb6RCd/IhGd4zD0YYsX7RZMuNc2YnWKwm3XpFp3dZswzazLimUECKS9hrYNhQc7n8Y/6Q64A6xUACLE+VmAYJsByIZZjAAA0pcExHMcImtSoKI2WMqppg5rSq0jdlPVnkUAiE0zmCAWBxjUpumygQ8aQGByZAIGUh0DKQyCQ8hBIeQgEUh4CKQ+BEMH/DAA+qJu2Ww/NPgAAAABJRU5ErkJggg==';
```
pjcozzi: In order to make this `.js` file a lot smaller, can you submit a copy of `Cesium_Logo_Flat.png` that is perhaps 8x8 and then embed that here?

specs/lib/loadImageUrisSpec.js
```diff
@@ -0,0 +1,88 @@
...
+            "images": {
+                "Image0001": {
+                    "uri": "Cesium_Logo_Flat.png"
+                }
+            }
+        };
+
+        gltf = loadGltfUris(gltf, filePath, function() {
+            expect(gltf.images.Image0001.extras.source).toBeDefined();
+            expect(gltf.images.Image0001.extras.source.toString()).toEqual(imageData.toString());
```
pjcozzi: Instead of comparing these as strings, can we compare the raw bytes?  It is more direct.  See if there is a `memcmp` type function for buffers.

lib/loadGltfUris.js
```diff
@@ -33,31 +25,59 @@ function loadGltfUris(gltf, basePath, callback) {
...
+    if (defined(objects)) {
+        var ids = Object.keys(objects);
+        async.each(ids, function(id, asyncCallback) {
+            var object = objects[id];
+            var uri = object.uri;
+            object.extras = defaultValue(object.extras, {});
+
+            //Load the uri into the extras object based on the uri type
+            if (isDataUri(uri)) {
+                object.extras.source = dataUriToBuffer(uri);
```
pjcozzi: I think it is good to just treat images as buffers for many pipelines we will build, but, later, we'll also want to decompress them (like JavaScript's `Image` object) so that, for example, we can build texture atlases.

pjcozzi:
Just to be extra sure, have you been dragging and dropping models we write out (both with and without embedded resources) into http://cesiumjs.org/convertmodel.html?

pjcozzi:
That's all my comments, nice work @leerichard42!

leerichard42:
For the model converter, it seems that it only takes in COLLADA files? An error occurs when I try to drag in any gltf files.

lib/loadGltfUris.js
leerichard42: Okay, got it!

leerichard42:
Updated!

pjcozzi:
> For the model converter, it seems that it only takes in COLLADA files? An error occurs when I try to drag in any gltf files.

Ah, sorry, make sure the model has embedded resources.  It doesn't work when the images, etc. are separate files.

I just tested it, and it worked for me.

Let me know when you verify the models output here and we'll merge this.

pjcozzi:
Recent code changes look good!

leerichard42:
Okay, I tested it by manually replacing the embedded buffer and it works!

leerichard42:
Also added the last buffer loading stage.

pjcozzi:
Thanks for testing!  Looks good, just that one minor comment!

leerichard42:
Updated!

pjcozzi:
:+1:

PR #35, opened 2016-02-20
Added write images stage

leerichard42:


lib/writeGltf.js
```diff
@@ -38,40 +28,93 @@ function writeGltf(gltf, outputPath, isEmbedded, createDirectory, callback) {
...
+            if (defined(object.extras) && defined(object.extras,source)) {
+                var source = object.extras.source;
+                delete object.extras.source;
+
+                //Write the source object as a data or file uri depending on the isEmbedded flag
+                if (isEmbedded) {
+                    var sourceDataUri = new dataUri();
+                    switch (name) {
+                        case 'images':
+                            sourceDataUri.format('.png', source);
```
pjcozzi: Valid glTF images are `png`, `jpg`, `gif`, and `bmp` (since these are the ones supported by browsers).  Perhaps we should save off the format when loading so that we can write out the same image format when writing.

Later, we'll have options to convert images (e.g., convert all images to `jpg`), resize images, etc.

lib/writeGltf.js
```diff
@@ -38,40 +28,93 @@ function writeGltf(gltf, outputPath, isEmbedded, createDirectory, callback) {
...
+                            fileName += '.glsl';
+                            break;
+                        default:
+                            throw new Error('Trying to write out invalid file type.');
+                    }
+
+                    object.uri = fileName;
+                    var outputPath = path.join(filePath, fileName);
+                    fs.writeFile(outputPath, source, function (err) {
+                        if (err) {
```
pjcozzi: This `if...else` block can just be `asyncCallback(err)`.  If `err` is `undefined`, it is basically the same as `asyncCallback()`.

pjcozzi:
Nice job, @leerichard42!  Just those comments.

leerichard42:
Updated!

pjcozzi:
:+1:

I merged in master and fixed conflicts in the tests.

PR #37, opened 2016-02-28
Added parse binary gltf stage

leerichard42:


leerichard42:
I had some trouble with the testing for this stage because the embedded image/shader files were different from those in the normal glTF folder. Namely, the external image file was a different size and had a different buffer from the embedded image, and the embedded shader used 0d0a for line breaks instead of just 0a like the external shader file. 
As a workaround, I wrote out the embedded resources and tested against those, but I'm not too sure if that's a valid test.

pjcozzi:
> As a workaround, I wrote out the embedded resources and tested against those, but I'm not too sure if that's a valid test.

I'll look to confirm that this is reasonable, but it is what I would have done. :smile:

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,60 @@
...
+var defaultValue = Cesium.defaultValue;
+var StringDecoder = require('string_decoder').StringDecoder;
+module.exports = parseBinaryGltf;
+
+function parseBinaryGltf(data) {
+    var sizeOfUint32 = Uint32Array.BYTES_PER_ELEMENT;
+    var byteOffset = 0;
+
+    if (data.slice(byteOffset, sizeOfUint32).toString() !== 'glTF') {
+        throw new Error('File is not valid binary glTF');
```
pjcozzi: Throw a new `DeveloperError` (`Cesium.DeveloperError`).  We should do that throughout if we are not already.

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    byteOffset += 3*sizeOfUint32;
```
pjcozzi: Instead of skipping over the version field, let's check it, and throw if it isn't version `1`.

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,60 @@
...
+    var dataView = new DataView(data.buffer);
+    var sceneLength = dataView.getUint32(byteOffset, true);
+
+    byteOffset += 2*sizeOfUint32;
```
pjcozzi: Also, check the scene format field and throw if it isn't `0` (indicating JSON).

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,60 @@
...
+var defined = Cesium.defined;
+var defaultValue = Cesium.defaultValue;
+var StringDecoder = require('string_decoder').StringDecoder;
+module.exports = parseBinaryGltf;
+
+function parseBinaryGltf(data) {
+    var sizeOfUint32 = Uint32Array.BYTES_PER_ELEMENT;
+    var byteOffset = 0;
+
+    if (data.slice(byteOffset, sizeOfUint32).toString() !== 'glTF') {
```
pjcozzi: Do you know how fast `slice` is?  Does it make a copy?  For performance, we would rather just "peek" into a subset of the buffer (using the right type).  In client-side JavaScript this is done with Typed Arrays.

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,60 @@
...
+    byteOffset += 2*sizeOfUint32;
+    var decoder = new StringDecoder();
+    var scene = decoder.write(data.slice(byteOffset, byteOffset + sceneLength));
+    var gltf = JSON.parse(scene);
+
+    byteOffset += sceneLength;
+    var body = data.slice(byteOffset);
+    var buffers = gltf.buffers;
+    if (defined(buffers)) {
+        if (defined(buffers.KHR_binary_glTF || defined(buffers.binary_glTF))) {
```
pjcozzi: Cesium has to support a bug in our exporter that generated buffers named `"KHR_binary_glTF"`, but we fixed that bug so we don't need that check here.

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,60 @@
...
+    var dataView = new DataView(data.buffer);
+    var sceneLength = dataView.getUint32(byteOffset, true);
+
+    byteOffset += 2*sizeOfUint32;
+    var decoder = new StringDecoder();
+    var scene = decoder.write(data.slice(byteOffset, byteOffset + sceneLength));
+    var gltf = JSON.parse(scene);
+
+    byteOffset += sceneLength;
+    var body = data.slice(byteOffset);
```
pjcozzi: With the above said, this `slice` is probably what we want here and the `slice` above is for a small enough buffer that we might not care.

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,60 @@
...
+    var buffers = gltf.buffers;
+    if (defined(buffers)) {
+        if (defined(buffers.KHR_binary_glTF || defined(buffers.binary_glTF))) {
+            var buffer = defaultValue(buffers.KHR_binary_glTF, buffers.binary_glTF);
+            buffer.extras = defaultValue(buffer.extras, {});
+            buffer.extras.source = body;
+        }
+    }
+
+    if (defined(gltf.extensionsUsed) && gltf.extensionsUsed.indexOf("KHR_binary_glTF") > -1) {
```
pjcozzi: If the file is valid binary glTF, this should always be true.

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,60 @@
...
+function loadSourceFromBody(gltf, body, name) {
+    var objects = gltf[name];
+    if (defined(objects)) {
+        for (var objectId in objects) {
+            var object = objects[objectId];
+            var objectExtensions = object.extensions;
+            if (defined(objectExtensions) && defined(objectExtensions.KHR_binary_glTF)) {
+                var objectView = objectExtensions.KHR_binary_glTF.bufferView;
+                var bufferView = gltf.bufferViews[objectView];
+                var source = body.slice(bufferView.byteOffset, bufferView.byteOffset + bufferView.byteLength);
```
pjcozzi: Ah, I see what you were asking in your email now about storing the buffer.

When we write back out the glTF asset, we don't want to duplicate any data in buffers.  We also want to be able to convert non-binary glTF to binary glTF, and perform all optimizations in a generic manner regardless of if the source is binary glTF or not.

Given all this, perhaps the best approach is to not actually store the full `"binary_glTF"` buffer in our in-memory data structure, and instead extract a new buffer for each buffer view (includes those in the extension for images and shaders pointing to buffer views).

Then, we can add a "combine buffers" optimization stage that combines all buffers into one buffer (taking into account if two buffer views point to the same exact part of a buffer, or if a buffer view points to a subset of a buffer pointed to by another buffer view, e.g., we want to avoid duplicating data).

What do you think of this approach?

specs/lib/parseBinaryGltfSpec.js
```diff
@@ -0,0 +1,68 @@
...
+            }
+            done();
+        });
+    });
+
+    it('loads a glTF scene', function() {
+        var gltf = parseBinaryGltf(testData.binary);
+
+        expect(gltf).toBeDefined();
+        expect(gltf.buffers.KHR_binary_glTF).toBeDefined();
```
pjcozzi: Look for `binary_glTF`.  We just fixed the sample models today: https://github.com/KhronosGroup/glTF/pull/542

specs/lib/parseBinaryGltfSpec.js
```diff
@@ -0,0 +1,68 @@
...
+        expect(gltf.shaders.CesiumTexturedBoxTest0VS).toBeDefined();
+        delete gltf.buffers.KHR_binary_glTF.extras;
+        delete gltf.images.Image0001.extras;
+        delete gltf.shaders.CesiumTexturedBoxTest0FS.extras;
+        delete gltf.shaders.CesiumTexturedBoxTest0VS.extras;
+
+        expect(gltf).toEqual(JSON.parse(testData.gltf));
+    });
+
+    it('loads an embedded image', function() {
```
pjcozzi: In a separate PR and once we have our binary glTF writer, let's add an "integration test" in the sense that we'll take our test glTF asset with all the properties that can be removed, convert it to binary glTF, then the remove optimizations, and then verify that it generated an optimized asset.

pjcozzi:
A few tests are failing on Travis: https://travis-ci.org/AnalyticalGraphicsInc/gltf-pipeline/builds/112356787

Any ideas?  Is it the node version?

pjcozzi:
That's all my comments.  Great start here!

lib/parseBinaryGltf.js
leerichard42: I think `slice` looks directly into the memory of the buffer subset - https://nodejs.org/api/buffer.html#buffer_buf_slice_start_end. The documentation says that the Node Buffer is an implementation of the Uint8Array Typed Array.

lib/parseBinaryGltf.js
leerichard42: I think this sound good - I'm thinking maybe I could first load the sources for the images and shaders and delete their bufferviews, and then I could create a buffer for each remaining bufferview, now excluding the image/shader bufferviews. Do you think deleting these bufferviews would be valid, or could that lead to problems?

lib/parseBinaryGltf.js
pjcozzi: Cool. :+1:

lib/parseBinaryGltf.js
pjcozzi: I _think_ it will be OK as long as no other accessors are pointing to the same buffer viewer.  The bookkeeping should be straightforward enough.

leerichard42:
Updated! Do you think the tests for managing the overlapping buffers are too verbose? I just wanted to be thorough and check the cases in which one buffer is entirely within another, as well as the case where two buffers are partially overlapping. I've also only tested with two overlapping buffers, maybe I should test three? Also, when would these kinds of overlapping buffer views be used in practice?

pjcozzi:
Thanks @leerichard42.  I am at GDC and will aim to review this on Thursday.  I'll let you know then if we might need more tests.

As for when this happens in practice, I suspect progressive LOD is the most prominent use case.

lib/loadGltfUris.js
```diff
@@ -94,7 +98,12 @@ function loadURI(gltf, basePath, name, callback) {
     }
 }
 
+//Return the extension of the data uri
  function getImageDataUriFormat(uri) {
-    //Return the extension of the data uri
-    return '.' + uri.match('\/([^)]+)\;')[1];
+    var extension = uri.match('\/([^)]+)\;');
+    if (extension === null) {
```
pjcozzi: Is assuming `.png` the right thing to do here?  Does this case ever happen for valid data URIs?  Would it be better to throw an exception?

leerichard42: This was mainly to address images in binary glTF, where we have the image data but the data URI is just "data:,", so I just used a placeholder since no format is specified (I think? I couldn't find the image format being represented in the binary image object.)

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,170 @@
...
+    var gltf = JSON.parse(scene);
+
+    //Extract binary body
+    byteOffset += sceneLength;
+    var body = data.slice(byteOffset);
+    
+    //Find bufferViews used by accessors
+    var usedBufferViews = getUsedBufferViews(gltf);
+
+    //Add image and shader sources, and delete their bufferView if not by an accessor
```
pjcozzi: "not **referenced** by accessor"

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,170 @@
...
+            for (var bufferViewId in bufferViews) {
+                if (bufferViews.hasOwnProperty(bufferViewId)) {
+                    var bufferView = bufferViews[bufferViewId];
+                    bufferView.extras = defaultValue(bufferView.extras, {});
+                    bufferView.extras.id = bufferViewId;
+                }
+            }
+
+            //Create bufferView array and sort by increasing byteOffset
+            var bufferViewArray = objectValues(bufferViews);
```
pjcozzi: Perhaps `sortedBufferViews` is a better name.

lib/parseBinaryGltf.js
```diff
@@ -0,0 +1,170 @@
...
+            bufferViewArray.sort(function(a, b) {
+                return a.byteOffset - b.byteOffset;
+            });
+            //Get bufferViews referencing binary_glTF
+            bufferViewArray = bufferViewArray.filter(function(bufferView) {
```
pjcozzi: Why not filter first, then we would potentially sort fewer objects?

specs/lib/parseBinaryGltfSpec.js
```diff
@@ -0,0 +1,164 @@
+'use strict';
+var fs = require('fs');
+var async = require('async');
+var crypto = require('crypto');
```
pjcozzi: Is this used?

specs/lib/parseBinaryGltfSpec.js
```diff
@@ -0,0 +1,164 @@
...
+        var gltf = parseBinaryGltf(testData.binary);
+
+        expect(gltf).toBeDefined();
+        expect(gltf.buffers.bufferView_29_buffer).toBeDefined();
+        expect(gltf.buffers.bufferView_30_buffer).toBeDefined();
+        expect(gltf.images.Image0001).toBeDefined();
+        expect(gltf.shaders.CesiumTexturedBoxTest0FS).toBeDefined();
+        expect(gltf.shaders.CesiumTexturedBoxTest0VS).toBeDefined();
+
+        delete gltf.bufferViews.bufferView_29.extras;
```
pjcozzi: We can remove this deletes if we add a stage like I mentioned in #38.

pjcozzi:
@leerichard42 overall, this looks great!  Just those minor comments.  Remember that when we code the binary glTF writer, we'll need a stage to merge buffers as we previously discussed.

lib/loadGltfUris.js
pjcozzi: Ah, I see.

So we're assuming .png for binary glTF...for example if someone used gltf-pipeline to convert from binary glTF to glTF with separate resources, this would fail if the images were actually jpg?

I _think_ we're going to have to inspect the header of the binary blob to determine the image format (it can only be `jpg`, `png`, `bmp`, or `gif`).

Go for it now if this interests you; otherwise, submit an issue with a link to this discussion and we can consider it in a bit.

leerichard42:
Updated! I added the extras._pipeline stage, and checked the image format using the first two bytes of their binary blobs (used http://www.garykessler.net/library/file_sigs.html as a reference).

lib/loadGltfUris.js
```diff
@@ -41,23 +42,26 @@ function loadURI(gltf, basePath, name, callback) {
             var object = objects[id];
             var uri = object.uri;
             object.extras = defaultValue(object.extras, {});
+            object.extras._pipeline = defaultValue(object.extras._pipeline, {});
```
pjcozzi: No need to change this now, but just a heads up that this just adds `extras._pipeline` to buffers, images, and shaders.  In the future, I suspect we will want a stage early in the pipeline to add it to all objects so later stages know it will be there, and then a stage at the end of the pipeline to remove it (and `extras` when possible) from all objects.

pjcozzi: `removePipelineExtras` will work just fine when we make this change.

lib/removePipelineExtras.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+
+module.exports = removePipelineExtras;
+
+function removePipelineExtras(gltf) {
```
pjcozzi: It can be in a separate PR, but glTF objects can be nested more than two levels deep, e.g., render states, and glTF allows adding `extras` to any glTF object property (including the root property).  For example, see: https://github.com/KhronosGroup/glTF/tree/master/specification#properties

The right way to handle this is to recursively walk into objects (not numbers/strings) and check for `extras`.  Base the code off of https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/clone.js

Also, write unit tests.

Again, a separate PR is fine so we can merge this one.

pjcozzi:
Great progress!  We are _close_ to a converter for all the glTF variants - just need to combine buffer views and write binary glTF!

lib/removePipelineExtras.js
leerichard42: I see - when you say check for extras, would I be able to do this if the nested glTF object didn't have an extras object, since they're optional? Would I need to separately keep track of the actual glTF properties hierarchy?

And just checking, is the removePipelineExtrasSpec file I added sufficient for unit testing (in this case)?

leerichard42: Actually, I didn't realize that the recursive walk was for removing objects - I see what you mean in that case. When I'm adding the extras objects though, I would need some kind of separate reference for which objects are glTF objects, right?

pjcozzi: > When I'm adding the extras objects though, I would need some kind of separate reference for which objects are glTF objects, right?

I don't think so.  What exactly do you mean? You would recursively walk objects and add `extras`.  Recursion would stop if you reach a non-object property or an `extras` object.

> And just checking, is the removePipelineExtrasSpec file I added sufficient for unit testing (in this case)?

For the unit tests, I would create a glTF asset with all possible object properties (one of each), and have object properties

1. with no `extras` - verify that it still doesn't have `extras`

2. with an empty `extras` - verify that it is not removed.

3. with just `extras._pipeline` - verify that it is all removed.

4. with `extras` and an extras sub-property and `extras._pipeline` - verify that only `_pipeline` is removed.

Note that 2. is an interesting case that we didn't consider yet where an input asset has an `extras : {}` and _we should not remove it_ in case the creator of the asset wanted it to be there for use later.  So perhaps the way to handle this is to add a flag to the `_pipeline` object that says `deleteExtras`.  It will be true if `extras` was not originally defined.

leerichard42: Okay, sounds good! I'll add those tests.

I'm thinking of something like the following case for adding extras:
```
"animations": {
    "animation_0": {
        "channels": [
            {
                "sampler": "animation_0_scale_sampler",
                "target": {
                    "id": "Bone",
                    "path": "scale"
                }
            }
        ],
        "parameters": {
            "TIME": "animAccessor_0"
        }
    }
}
```
In this case, we want to add extras to `animation_0`, the single channel in `channels`, and that channel's target. However, we wouldn't add extras to the `parameters` object in `animation_0`. So I guess I'm just confused about how I would differentiate between an object like `parameters` versus a glTF object like `target`.

pjcozzi: Ah, you are right.  So we would still use an approach like `clone`, but we need to check each object to see if it is one of the objects that the glTF spec says can have an `extras` as you suggested above..

PR #39, opened 2016-03-29
Added writeBinaryGltf stage and updated removePipelineExtras

leerichard42:
Will create a separate pull request for addPipelineExtras after this.


leerichard42:
For the removePipelineExtras spec, is it valid if I run removePipelineExtras one in the beforeAll function, and then run my tests on the resulting glTF object? Or is it better if I run the function in each test?

package.json
```diff
@@ -36,8 +36,11 @@
     "async": "^1.5.2",
     "buffer-equal": "^1.0.0",
     "cesium": "^1.17.0",
+    "clone": "^1.0.2",
```
pjcozzi: Could we just use `Cesium.clone` instead? - https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/clone.js

pjcozzi:
As we discussed in a previous PR, when we load binary glTF, we basically have to create a buffer for each buffer view.  In order to avoid writing out inefficient glTF assets, we planned to add a stage to basically merge buffers together.  It looks like this is done in `writeBinaryGltf.js`.  How hard is it to decouple this from writing binary glTF so it is just a general purpose pipeline stage like the others, e.g., removing unused objects?

pjcozzi:
@leerichard42 did you verify a few different output binary glTF assets by visualizing them with: http://cesiumjs.org/convertmodel.html

Please test ones with animations and skins.

For test models, see https://github.com/KhronosGroup/glTF/tree/master/sampleModels

A simple manual test could just read a binary glTF asset and then output it, verify it still renders OK, and that we didn't do anything to make the asset inefficient.

pjcozzi:
That's all my comments.  Really fantastic progress here!

package.json
leerichard42: Yes, I will replace the instances of `clone`.

leerichard42:
Separating the merge buffers stage should be simple, so I will add that. I've been testing with the box model, but I will test with others. I think there might be a problem with the RiggedSimple binary sample model, because the special binary buffer has id `KHR_binary_glTF` rather than just `binary_glTF`.

package.json
leerichard42: Actually, it looks like `Cesium.clone` when I try to deep clone the gltf object with loaded sources, because the Node Buffer class doesn't have a default constructor. Is there a workaround for this?

leerichard42:
Updated - I didn't replace the `clone` module yet due to the constructor issue, but I've separated the merge buffers stage.

pjcozzi:
> Separating the merge buffers stage should be simple, so I will add that. I've been testing with the box model, but I will test with others. I think there might be a problem with the RiggedSimple binary sample model, because the special binary buffer has id KHR_binary_glTF rather than just binary_glTF.

Sounds like bad sample data.  Drag and drop the [original COLLADA](https://github.com/KhronosGroup/glTF/tree/master/sampleModels/RiggedSimple) file [here](http://cesiumjs.org/convertmodel.html) and test with that.  Could you also make a pull request into the glTF repo to fix the sample model and any others if you find other issues?

pjcozzi:
> Actually, it looks like Cesium.clone when I try to deep clone the gltf object with loaded sources, because the Node Buffer class doesn't have a default constructor. Is there a workaround for this?

@mramato what do you think?  For more context, see https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/39#discussion_r57883631.

pjcozzi:
@leerichard42 this looks good, just make the RiggedSimple updates please, and we'll see about `clone`.

You are welcome to start on our next stage: #40.

mramato:
> @mramato what do you think? For more context, see #39 (comment).

I would just go with the node module.  It's a popular module and will handle Node specific cases that Cesium's simple clone function does not.  However, it looks like it's only used in tests, so it should be a dev dependency and not a regular dependency.

pjcozzi:
Sounds good, thanks @mramato.

@leerichard42 let me know when this is ready to merge.

leerichard42:
Updated the dependency! I tried to convert the COLLADA files, and I think the converter is actually where the `KHR_binary_gltf` id is added - when I convert boxTextured, whose .glb file previously worked, it adds the `KHR_binary_gltf` id to the resulting .glb scene.

pjcozzi:
@tfili I thought we fixed our converter so it names the buffer in a binary glTF file `"binary_glTF"`.  @leerichard42 says it is still naming it `"KHR_binary_gltf"`.

tfili:
It was fixed in PR https://github.com/AnalyticalGraphicsInc/agi-asset-pipeline/issues/242.

It looks like the online converter didn't get the update. Working on addressing this now.

pjcozzi:
@tfili did we deploy the updated converter?

tfili:
Yep. I tested and got the correct output.

pjcozzi:
@leerichard42 can you please test the animated/skinned models now that the converter is fixed?

leerichard42:
Yes, writing glTF and binary glTF work now with .glb as input! I will make a pull request into the glTF repo to fix the files now.

leerichard42:
Update: It looks like RiggedSimple was the only file with that problem, and everything else was fine.

pjcozzi:
Sounds great, thanks @leerichard42!

PR #41, opened 2016-04-07
Added addPipelineExtras stage

leerichard42:


lib/addPipelineExtras.js
```diff
@@ -0,0 +1,115 @@
...
+
+    return gltf;
+}
+
+function addPipelineExtra(object, reference) {
+    if (object === null || typeof object !== 'object') {
+            return object;
+    }
+
+    if (defined(reference)) {
```
pjcozzi: Since this is a private function that is only called from the function above, this check is not needed since it will always be defined.

lib/addPipelineExtras.js
```diff
@@ -0,0 +1,115 @@
...
+        "addPipelineExtra": true
+    };
+
+    addPipelineExtra(gltf, reference);
+
+    return gltf;
+}
+
+function addPipelineExtra(object, reference) {
+    if (object === null || typeof object !== 'object') {
```
pjcozzi: It would be better separation of concerns to move this check to the function above (e.g., that function is the interface and this is the implementation).  Also, why not use `defined` here?

lib/addPipelineExtras.js
pjcozzi: Ah, if here and above are needed to stop recursion then leave them, of course.

lib/addPipelineExtras.js
```diff
@@ -0,0 +1,115 @@
+'use strict';
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+var defaultValue = Cesium.defaultValue;
+
+module.exports = addPipelineExtras;
+
+//Add the pipeline extras object to each glTF object.
+function addPipelineExtras(gltf) {
+    var reference = {
```
pjcozzi: Does the `reference` object need the `addPipelineExtra` properties?  My original - but perhaps too simplistic - thought was just having the glTF properties name would be enough, for example:
```javascript
var reference = {
    "accessors" : true,
    "animations" : {  // implicitly true since all glTF object properties have an extras sub-property
        "channels" : [ // same comment
            {
                "target" : true
            }
        ],
        "samplers" : true
    }
    // ...
}
```
Does this make the algorithm harder (or impossible) to implement?

lib/addPipelineExtras.js
leerichard42: I used this line to match the check in `Cesium.clone` - I think I could alternatively use `defined` and also check that the type of the object is `object`. Would one of these be preferable over the other?

lib/addPipelineExtras.js
leerichard42: After looking over it again, I think the main reason I had the `addPipelineExtra` property and the additional reference objects like `accessorObject`, `animationObject`, etc. in the reference was to differentiate between glTF objects which directly had extras, like:

```
"asset": {
        "generator": "collada2gltf@ceec062e3d5793f2f249f53cbd843aee382ad40b",
        "premultipliedAlpha": true,
        "profile": {
            "api": "WebGL",
            "version": "1.0.2"
        },
        "version": 1
    },
```
where `extras` is added to the `asset` object,
vs. 
```
"accessors": {
        "accessor_21": {
            "bufferView": "bufferView_29",
            "byteOffset": 0,
            "componentType": 5123,
            "count": 36,
            "type": "SCALAR"
        }
    }
```
where `extras` is added to the `accessor_21` object.
So, I think I would still need some kind of marker variable which determines which kind of glTF object it is.

lib/addPipelineExtras.js
leerichard42: I condensed this line with the next if statement in this file and the `removePipelineExtras` module.

pjcozzi:
All sounds good @leerichard42!

Issue #33, opened 2016-02-13
Read/write shader test failures when ran locally


pjcozzi:
Details: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/32#issuecomment-182601524

Potential workaround: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/32#issuecomment-183161478

PR #44, opened 2016-04-18
Added case for dos file endings in tests

lasalvavida:
In reference to #33. The Base64 shader uris in these tests needed to be updated and were the cause of the failing tests.

Edit: This works locally, I'm looking into why travis is failing; I'm pretty sure its a windows->unix line-ending problem


lasalvavida:
So, the issue was that the uri produced is different if your git checks out the files with dos line endings. I've modified the tests to allow either.

pjcozzi:
It's a bit awkward to have to keep both of those strings, and this is not ideal maintenance for adding future tests.

Can we write a utility function that wraps `bufferEqual(dataUriToBuffer(..` and encapsulates this?  Also include a comment about why we are doing this since it will be non-obvious.

If you do some Node research, there may be a way to normalize this when reading the file.

lasalvavida:
@pjcozzi Is there a strong argument for keeping the hard-coded uris at all? Something like this would probably be a lot cleaner:

```javascript
var fragmentShaderUri = 'data:text/plain;base64,' + new Buffer(fragmentShaderData).toString('base64');
```

pjcozzi:
Sounds good, give it a try!

lasalvavida:
@pjcozzi This is ready for another look

pjcozzi:
Very nice, thanks @lasalvavida!

Issue #43, opened 2016-04-12
Combine primitives/nodes


pjcozzi:
@leerichard42 let's do this next.  It is a great performance optimization since it reduces the number of draw calls required at runtime and some exporters can generate really sub-optimal models. For example, I've seen the SketchUp export COLLADA where each triangle needed a separate draw call!  After running an optimization like this, the model only needed one draw call.

* [x] Stage to combine `primitives` in a node (if they have the same attribute formats, material, and mode; this will also require combining accessors/bufferviews/buffers)
* [x] Stage to combine `meshes` in a node
* [x] Stage to combine nodes (if they are not targeted for animation and have the same material), KhronosGroup/glTF#20

@tfili any additional advice here?

tfili:
The only thing that was a bit of a pain was when meshes were used in multiple nodes with different transforms. When you combine those the mesh data needs to be duplicated because you will have to premultiply the transform before you combine them.

pjcozzi:
@leerichard42 any update here?

leerichard42:
Just checking, do you mean "combine primitives in a mesh"? Or can different meshes share common primitives?

pjcozzi:
I mean if a mesh has an array of three primitives, for example, that those primitives would be combined into one primitive and the mesh would refer to that.

leerichard42:
How should I deal with cases in which primitives share the same material/mode/attributes, but differ in properties in their accessors/bufferViews/buffers?
For example, if two similar primitives pointed to accessors with different types, or bufferViews with different targets, etc.
Right now, I have a check which does not merge the primitives if this happens, but I'm just wondering if certain cases should be handled in a different way?

pjcozzi:
Can you provide a concrete example?  For example, if two primitives have the same attributes then the accessors will have the same format and we should be able to merge them.

leerichard42:
I see, that answers my question - I wasn't sure if we were assuming that accessors for the same attribute would necessarily be in the same format.

However, I have a similar question about min/max values in accessors, such as in the following example from the boxAnimated sample model:
```
"accessors": {
    "accessor_31": {
        "bufferView": "bufferView_65",
        "byteOffset": 0,
        "byteStride": 12,
        "componentType": 5126,
        "count": 224,
        "max": [
            0.5,
            0.5,
            0.5
        ],
        "min": [
            -0.5,
            -0.5,
            -0.5
        ],
        "type": "VEC3"
    },
[...]

PR #46, opened 2016-04-25
Added combinePrimitives stage

leerichard42:
Finally got this one working! 
A few questions:
Can we assume that indices will be UInt16? Or should I check for the type when I calculate index offsets?

Should I carry over any name/extension/extra data from the old accessors/bufferViews/buffers? I didn't, because I figured there wouldn't be a way to choose between multiple objects with different properties.

And finally, is testing the object names like I did in the spec sufficient? I was mainly testing by using the converter to confirm that it was working.


pjcozzi:
> Can we assume that indices will be UInt16? Or should I check for the type when I calculate index offsets?

I suppose they could also be uint8 even though I've never seen it in practice.

There will also be an extension to allow uint32, see https://github.com/KhronosGroup/glTF/issues/570.

pjcozzi:
> Should I carry over any name/extension/extra data from the old accessors/bufferViews/buffers?

Ah, interesting point.

Can we output a warning if there is an extension/extras that it will be lost.  The name should not be an issue since there are basically internal objects compared to, for example, animations and nodes.

pjcozzi:
> And finally, is testing the object names like I did in the spec sufficient?

Will comment inline if I notice anything.

pjcozzi:
For #43.

pjcozzi:
@tfili can you also do a quick review here?

lib/combinePrimitives.js
```diff
@@ -0,0 +1,270 @@
...
+var defined = Cesium.defined;
+var defaultValue = Cesium.defaultValue;
+var DeveloperError = Cesium.DeveloperError;
+var loadGltfUris = require('./loadGltfUris');
+module.exports = combinePrimitives;
+
+//Combines primitives in a mesh which have the same material, mode, and attributes.
+function combinePrimitives(gltf) {
+    if (defined(gltf.accessors) && defined(gltf.bufferViews) && defined(gltf.buffers)) {
+        loadGltfUris(gltf);
```
pjcozzi: Should this just happen once?  For example, we will build pipelines that have several optimizations and we'll just one to load urls once.

For now, users can just be required to call it, and just include it in the main `.js` file for the command-line tool.

lib/combinePrimitives.js
```diff
@@ -0,0 +1,270 @@
...
+        combinedPrimitives.push(combinePrimitiveGroup(gltf, meshId, primitiveGroups[j]));
+    }
+
+    return combinedPrimitives;
+}
+
+
+//Checks if two primitives can be combined based on their attributes, materials, and modes.
+function readyToCombine(a, b) {
+    if (a.material !== b.material || a.mode !== b.mode || defined(a.indices) !== defined(b.indices)) {
```
pjcozzi: Do the indices need to match?  Should they be combined as well?

pjcozzi: Or does that likely happen in an earlier stage of the pipeline?

pjcozzi: If so, include a code comment here.

lib/combinePrimitives.js
```diff
@@ -0,0 +1,270 @@
...
+        return false;
+    }
+
+    var aKeys = Object.keys(a.attributes);
+    var bKeys = Object.keys(b.attributes);
+    if (aKeys.length !== bKeys.length) {
+        return false;
+    }
+    else {
```
pjcozzi: It's cleaner to just drop the `else` here since it is not needed.

lib/combinePrimitives.js
```diff
@@ -0,0 +1,270 @@
...
+        var bufferViewSource = bufferSource.slice(viewByteOffset, viewByteOffset + viewByteLength);
+
+        //Calculate the length of the accessor source using the count and byteStride or componentType.
+        var accessorByteLength = accessor.count;
+        if (accessor.byteStride > 0) {
+            accessorByteLength *= accessor.byteStride;
+        }
+        else {
+            //Int16
+            if (accessor.componentType === 5122 || accessor.componentType === 5123) {
```
pjcozzi: It is technically a private API in Cesium, but I suggest using https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Renderer/WebGLConstants.js here.

pjcozzi: And throughout to replace hardcoded WebGL constants.

specs/lib/combinePrimitivesSpec.js
```diff
@@ -0,0 +1,82 @@
...
+    it('does not affect single primitives', function() {
+        var boxCopy = clone(box);
+        addPipelineExtras(boxCopy);    
+        combinePrimitives(boxCopy);
+        writeGltf(boxCopy, './boxOutput.gltf', true, true, function() {
+            expect(boxCopy).toEqual(box);
+        });
+    });
+
+    it('combines two primitives referencing the same bufferView', function() {
```
pjcozzi: Can we also test the case when primitives can't be combined.  There is some missing coverage: https://coveralls.io/jobs/13867440/source_files/310007864

In general consider:
* When all primitives in a mesh can be combined
* When no primitives in a mesh can be combined
* When some primitives in a mesh can be combined
   * When, for example, a mesh with 5 primitives, can combine 2 primitives into 1, another 2 primitives into 1, and not combine the last primitive so 5 primitives become 3.

pjcozzi:
Nice start here @leerichard42!

lib/combinePrimitives.js
leerichard42: I think I assumed that the indices property would need to exist in similar primitives, so I treated them like the attributes and combined their accessors as well. However, should I just combine primitives even if they don't both have the indices property? I wasn't sure what a case without indices would be.

pjcozzi: Ah, I didn't catch the `defined` call.  You are right.  Both should have indices.  We could consider adding indices if one primitive has indices and the other does not, but this would also need heuristics to see when it is worth it.  No need to consider this until we know if it is an issue in practice.

leerichard42:
Updated! Note - this stage ends up creating additional unused accessors/bufferViews/buffers after it is run. Should there be another removeUnused call in the command-line after all the pipeline stages are run? One to remove unused objects in the original file, and one to remove objects which become unused later on?

pjcozzi:
> Should there be another removeUnused call in the command-line after all the pipeline stages are run?

Yes, add a comment to say why it is being ran a second time (we can't just only run it at the end because then this stage wouldn't catch everything, right?).

leerichard42:
Updated! Working on the other combine stages now.

lib/combinePrimitives.js
```diff
@@ -0,0 +1,291 @@
...
+    var referenceAccessor = defined(attributeType) ? accessors[referencePrimitive.attributes[attributeType]] : accessors[referencePrimitive.indices];
+    var byteStride = defaultValue(referenceAccessor.byteStride, 0);
+    var componentType = referenceAccessor.componentType;
+    var type = referenceAccessor.type;
+    var count = 0;
+    var target, max, min;
+    var maxIndex = -1;
+
+    var attributeKey = defaultValue(attributeType, 'INDEX');
+    var newBuffer = {
```
pjcozzi: I believe we have another stage that combines buffers, and compacts buffers to remove their unused parts.  Remind me, where is that?  Perhaps add a comment here to indicate that stage should be ran later in the pipeline after this since looking at this in isolation doesn't make that clear.

Feel free to add this comment directly in master or another PR since I am going to merge this.

pjcozzi:
Fantastic, thanks again @leerichard42!

PR #47, opened 2016-05-10
Added combineMeshes stage

leerichard42:
This is a branch off the combinePrimitives stage, and I made a small fix in the jasmine describe header for the combinePrimitives test in addition to the combineMeshes stage.

### DIFF
--- a/lib/combineMeshes.js
+++ b/lib/combineMeshes.js
@@ -3,7 +3,8 @@ var Cesium = require('cesium');
 var defined = Cesium.defined;
 module.exports = combineMeshes;
 
-//Combines meshes in a node.
+//Combines meshes in a node. 
+//Should be run before the combinePrimitives stage to merge all primitives for a node into one mesh.
 function combineMeshes(gltf) {
     var nodes = gltf.nodes;
     var meshes = gltf.meshes;
--- a/lib/combineMeshes.js
+++ b/lib/combineMeshes.js
@@ -1,0 +1,57 @@
+'use strict';
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+module.exports = combineMeshes;
+
+//Combines meshes in a node.
+function combineMeshes(gltf) {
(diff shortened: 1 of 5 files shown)


pjcozzi:
Do you want to call `combineMeshes` from index.js for the example pipeline?

pjcozzi:
Part of #43.

lib/combineMeshes.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+module.exports = combineMeshes;
+
+//Combines meshes in a node.
+function combineMeshes(gltf) {
```
pjcozzi: This needs to run before the `combinePrimitives` stage, right?  This creates just one mesh with a big array with all the primitives, then it is expected that `combinePrimitives` will combine primitives.

I think it is worth adding a comment here.

pjcozzi:
Great start, @leerichard42!

Can you also add test case for the case @tfili mentioned in https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/43#issuecomment-209470582?

This would be a bit more of an integration test where the test would have both the combine mesh and primitives stages.

lib/combineMeshes.js
leerichard42: Yes, I will add this comment.

leerichard42:
I thought that the transforms test case was for the combine nodes stage? Because wouldn't combining the nodes, rather than the meshes, affect the data, because the nodes with different transforms would be merged into one node with a single transform?

pjcozzi:
> I thought that the transforms test case was for the combine nodes stage? Because wouldn't combining the nodes, rather than the meshes, affect the data, because the nodes with different transforms would be merged into one node with a single transform?

Ah, yes, you are right.  I was confusing the glTF and COLLADA terminology.

pjcozzi:
Great work as always!

Issue #1, opened 2015-11-15
Roadmap


pjcozzi:
- [x] Stage to quantize vertices to use fewer bits using [WEB3D_quantized_attributes](https://github.com/KhronosGroup/glTF/blob/master/extensions/Vendor/WEB3D_quantized_attributes/README.md)
- [x] Stage to prebake AO
- [ ] Command-line and JSON input, e.g., like pdal (see [here](http://www.pdal.io/apps.html) and [here](http://www.pdal.io/pipeline.html))
- [x] Option to output separate resources (e.g., images) when writing out binary glTF
- [ ] Stage to minify JSON and GLSL, details: https://github.com/KhronosGroup/glTF/issues/35 and https://github.com/KhronosGroup/glTF/issues/36
- [x] Stage for pre and post vertex cache optimization.  Use Cesium's [implementation](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/Tipsify.js).
- [x] Stage for oct encoding, see [AttributeCompression.octEncode](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/AttributeCompression.js#L43)
- [x] Stage to remove duplicate (property-wise) techniques/materials.
- [x] Stage to remove/merge duplicate vertices, https://github.com/KhronosGroup/glTF/issues/177
- [x] Stage to convert glTF 1.0 to 1.1, https://github.com/KhronosGroup/glTF/issues/605
- [ ] Stages from FBX2glTF.  See [Slide 5](https://www.khronos.org/assets/uploads/developers/library/2016-siggraph/glTF-MobileVR-Oculus-BOF-Update-SIGGRAPH_Jul16.pdf) from SIGGRAPH 2016
- [ ] Stage to compress textures, see http://gpuopen.com/compressonator-is-going-open-source/
- [ ] Stage to create texture atlases
- [x] Stage to replace [gltf-statistics](https://github.com/AnalyticalGraphicsInc/gltf-statistics)?
- [ ] Stage to interleave vertex attributes, https://github.com/KhronosGroup/glTF/issues/21
- [ ] Investigate GLSL optimizer, https://github.com/KhronosGroup/glTF/issues/34
- [ ] Stage to limit number of joints influencing a vertex, https://github.com/KhronosGroup/glTF/issues/203
- [x] Stage to generate normals if they are missing.
[...]

PR #50, opened 2016-05-13
Added removeUnusedVertices stage

leerichard42:
I think it's a good call to separate this into two stages, so this stage removes unused buffer data based on bufferViews, and the other stage will modify and split up bufferViews based on indices. Do you think I should keep both of these stages in the same file as private functions and just have two test files, or should I make a new lib file?

### DIFF
--- a/lib/removeUnusedVertices.js
+++ b/lib/removeUnusedVertices.js
@@ -28,8 +28,8 @@ function removeUnusedBufferData(gltf) {
     sortedBufferViews.sort(function(a, b) {
         return a.byteOffset - b.byteOffset;
     });
-
-    for (var i = 0; i < sortedBufferViews.length; i++) {
+    var bufferViewsLength = sortedBufferViews.length;
+    for (var i = 0; i < bufferViewsLength; i++) {
         var bufferView = sortedBufferViews[i];
         var viewedBuffer = buffers[bufferView.buffer];
         var source = viewedBuffer.extras._pipeline.source;
@@ -85,15 +85,7 @@ function removeUnusedIndexData(gltf) {
     var bufferViews = gltf.bufferViews;
     var meshes = gltf.meshes;
 
-    var primitives = [];
-    for (var meshId in meshes) {
-        if (meshes.hasOwnProperty(meshId)) {
-            if (defined(meshes[meshId].primitives)) {
-                primitives.push.apply(primitives, meshes[meshId].primitives);
-            }
-        }
-    }
-    calculateIndexIntervals(accessors, bufferViews, buffers, primitives);
+    calculateIndexIntervals(accessors, buffers, bufferViews, meshes);
 
     compressBuffers(accessors, bufferViews, buffers);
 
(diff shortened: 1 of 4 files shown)


pjcozzi:
@tfili can you also take a quick look at this?  For https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/1#issuecomment-215082269

pjcozzi:
> Do you think I should keep both of these stages in the same file as private functions and just have two test files?

Yes, I think when the end user builds a glTF pipeline they will just want to "remove unused vertices", even if it is implemented as two separate stages.

As for the tests, it would even be fine to put all the tests in one file.  Up to you.

specs/lib/removeUnusedVerticesSpec.js
```diff
@@ -0,0 +1,254 @@
...
+            "bufferView_1": {
+                "buffer": "buffer_0",
+                "byteOffset": 240,
+                "byteLength": 600
+            }
+        };
+
+        removeUnusedVertices(testGltf);
+
+        delete testGltf.buffers.buffer_0.extras._pipeline.offset;
```
pjcozzi: These two lines could go into an `afterEach` function.  Or replace the file-scoped `gltf` variable with a function that returns the glTF object that is called in each function so we don't have to worry about managing state between each test.

pjcozzi:
Good start here!  Nice tests.

pjcozzi:
@tfili perhaps wait until this PR includes removing vertices not referenced by indices before reviewing.

specs/lib/removeUnusedVerticesSpec.js
leerichard42: Actually, turns out I don't need these lines as I was using them in a test which I deleted.

leerichard42:
Updated - the index stage was tough! 

Here's a image of the box with a removed face:
image

pjcozzi:
> Updated - the index stage was tough!

Cool.  Hope it was fun!

> Here's an image of the box with a removed face:

You meant to remove the face, right?  :)

leerichard42:
Yep, I did tests where I removed the top, side, and bottom face indices! They're in the spec as well.

lib/removeUnusedVertices.js
```diff
@@ -0,0 +1,364 @@
...
+}
+
+//Remove sections of buffers which are not accessed by indices.
+function removeUnusedIndexData(gltf) {
+    var accessors = gltf.accessors;
+    var buffers = gltf.buffers;
+    var bufferViews = gltf.bufferViews;
+    var meshes = gltf.meshes;
+
+    var primitives = [];
```
pjcozzi: Is this array really needed?  Couldn't the top-level `for` loop in `calculateIndexIntervals` be replaced with a loop like the one below?

lib/removeUnusedVertices.js
```diff
@@ -0,0 +1,364 @@
...
+            indexArray.sort(function(a, b) {
+                return a - b;
+            });
+
+            //Calculate the index intervals based on the array of sorted indices.
+            var indexIntervals = [];
+            var intervalStart, prevIndex;
+            var newAccessorCount = 0;
+
+            for (var j = 0; j < indexArray.length; j++) {
```
pjcozzi: Throughout this file, we tend to write loops like this with (unless the length potential changes):
```javascript
var length = indexArray.length;
for (var j = 0; j < length; j++) {
```
to avoid the potential lookup each time.

pjcozzi:
@tfili this is ready. Can you also take a look at it?

pjcozzi:
> Yep, I did tests where I removed the top, side, and bottom face indices! They're in the spec as well.

The tests look good!

pjcozzi:
@tfili this is waiting on you.

tfili:
@pjcozzi Looks good to me.

PR #51, opened 2016-05-20
Added combineNodes stage

leerichard42:
The specs for this stage can be found in the output folder after running and viewed in the converter.

**TODO**

* [ ] Verify sample models, https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/51#issuecomment-220724989


bin/gltf-pipeline.js
```diff
@@ -78,6 +79,7 @@ fs.readFile(gltfPath, function (err, data) {
 
         combineMeshes(gltf);
         combinePrimitives(gltf);
+        combineNodes(gltf);
```
pjcozzi: Are we sure that we don't want to call this before `combineMeshes` and `combinePrimitives`?  Or does `combineNodes` also explicitly combine its meshes/primitives?

lib/combineNodes.js
```diff
@@ -0,0 +1,312 @@
+'use strict';
+var arrayEqual = require('array-equal');
+var Cesium = require('cesium');
+var clone = Cesium.clone;
+var defined = Cesium.defined;
+var defaultValue = Cesium.defaultValue;
+var IDENTITY = [1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1];
```
pjcozzi: I think it will be cleaner to use Cesium's [Matrix4](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/Matrix4.js).  Just construct it as needed from the array and then use all its helper functions.

bin/gltf-pipeline.js
leerichard42: I was thinking of calling `combineNodes` first, but right now it works based off the fact that each node has a combined mesh. Is it okay if I call `combineMeshes` at the end again to recombine the new meshes?

pjcozzi: Yes, that is no problem, just make sure to comment why things happen in the order that they do.

leerichard42:
Updated!

pjcozzi:
@leerichard42 can you merge in master?  I suspect there is a trivial conflict in gltf-pipeline.js since I recently merged #50.

pjcozzi:
@leerichard42 do you have to run each [sample model](https://github.com/KhronosGroup/glTF/tree/master/sampleModels) through this and look at the before/after results to see if everything makes sense to you?

[gltf-statistics](https://github.com/AnalyticalGraphicsInc/gltf-statistics) could be of use for this, but it only supports glTF 0.8 (it might just work with 1.0 too, I have not tested).

pjcozzi:
@tfili can you also look at this one to make sure we considered all the times that nodes/sub-trees can and can't be combined?

lib/combineNodes.js
```diff
@@ -0,0 +1,310 @@
...
+var Matrix4 = Cesium.Matrix4;
+var Cartesian3 = Cesium.Cartesian3;
+var combineMeshes = require('./combineMeshes');
+var combinePrimitives = require('./combinePrimitives');
+module.exports = combineNodes;
+
+//Combines nodes which are not targeted for animation and share the same material. 
+//Should be run after the combineMeshes and combinePrimitives stages, giving nodes one mesh each.
+function combineNodes(gltf, preserveCameras) {
+    var animations = gltf.animations;
```
pjcozzi: So V8 knows more about the type and has the potential to optimize better, perhaps add `preserveCameras = defaultValue(preserveCameras, false);` instead of relying on `undefined` to implicitly convert to `false`.

lib/combineNodes.js
```diff
@@ -0,0 +1,310 @@
...
+            var animation = animations[animationId];
+            if (defined(animation.channels)) {
+                var channelLength = animation.channels.length;
+                for (var i = 0; i < channelLength; i++) {
+                    nodes[animation.channels[i].target.id].extras._pipeline.isMergeable = false;
+                }
+            }
+        }
+    }
+    for (var nodeId in nodes) {
```
pjcozzi: Why not put this loop first so we don't need the `if (!defined(node.extras._pipeline.isMergeable)) {` check?

lib/combineNodes.js
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        //Merge the children which are not animated and have the same properties into the current subtree root.
+        var childIndex = 0;
+        while (childIndex < children.length) {
+            var childId = children[childIndex];
+            var child = nodes[childId];
+            if (canCombine(root, child, preserveCameras)) {
+                mergeChildWithParent(gltf, rootId, childId, childIndex, preserveCameras);
+                children.splice(childIndex, 1);
```
pjcozzi: This can end of being O(n^2) since each splice O(n).  Perhaps just copy surviving children into a new array and assign it to `root.children`.

pjcozzi:
> @leerichard42 do you have to run each sample model through this and look at the before/after results to see if everything makes sense to you?
>
> gltf-statistics could be of use for this, but it only supports glTF 0.8 (it might just work with 1.0 too, I have not tested).

Also, please make sure that each optimized model renders correctly espeically ones for complex node hiearchies that may have been partly combined for optimization, http://cesiumjs.org/convertmodel.html

lib/combineNodes.js
```diff
@@ -0,0 +1,310 @@
...
+                        }
+                    }
+                }
+                meshes[newMeshId] = newMesh;
+            }
+        }
+        if (defined(parent.meshes)) {
+            parent.meshes.push(newMeshId);
+        }
+        else {
```
pjcozzi: Minor, but put `else` on the line above.

lib/combineNodes.js
```diff
@@ -0,0 +1,310 @@
...
+
+    return resultBuffer;
+}
+
+//Returns true if two nodes do not have the same value for a property.
+function propertyNotShared(nodeA, nodeB, property) {
+    return defined(nodeA[property]) !== defined(nodeB[property]) || (defined(nodeA[property]) && (nodeA[property] !== nodeB[property]));
+}
+
+//Creates a new object id based on the attribute type and existing objects.
```
pjcozzi: Should this be in its own file?  Could anywhere else in gltf-pipeline use this?

pjcozzi:
Great test coverage for combineNodes.js!

lib/combineNodes.js
```diff
@@ -0,0 +1,310 @@
...
+                var mesh = gltf.meshes[meshId];
+                var newMesh = clone(mesh, true);
+                var primitives = newMesh.primitives;
+                if (defined(primitives)) {
+                    var primitivesLength = primitives.length;
+                    for (var i = 0; i < primitivesLength; i++) {
+                        var attributes = primitives[i].attributes;
+                        if (defined(attributes)) {
+                            transformAttribute(gltf, meshId, transform, 'POSITION', attributes, gltf.accessors, gltf.buffers, gltf.bufferViews);
+                            transformAttribute(gltf, meshId, transform, 'NORMAL', attributes, gltf.accessors, gltf.buffers, gltf.bufferViews);
```
pjcozzi: Are normals transformed with the inverse transpose matrix?

pjcozzi:
@leerichard42 I am almost done reviewing this; I will try to finish tomorrow morning.

lib/combineNodes.js
leerichard42: No, I will fix that! In that case, I will give transformAttribute three cases - one for positions, normals, and vectors.

lib/combineNodes.js
```diff
@@ -0,0 +1,290 @@
...
+        delete parent.camera;
+    }
+    delete parent.name;
+
+    delete nodes[childId];
+
+    return parent;
+}
+
+//Transforms the attribute data and creates a new accessor, buffer view, and buffer to access the data.
```
pjcozzi: This relies on buffers, etc. getting combined later in the pipeline.  Is everything in the right order in gltf-pipeline.js?  I just want to confirm that there aren't cases were we leave the model less efficient than before it went through the pipeline.

lib/combineNodes.js
```diff
@@ -0,0 +1,290 @@
...
+}
+
+//Returns a new buffer with the elements premultiplied by the transform matrix.
+function premultiplyTransformWithBuffer(matrix, buffer, attributeId, min, max) {
+    var bufferLength = buffer.length;
+    var resultBuffer = new Buffer(bufferLength);
+    var bufferIndex = 0;
+
+    while (bufferIndex < bufferLength) {
+        var currentVector = new Cartesian3(buffer.readFloatLE(bufferIndex), buffer.readFloatLE(bufferIndex + 4), buffer.readFloatLE(bufferIndex + 8));
```
pjcozzi: Use scratch variables here and in any tight loops.  This improves performance by reducing GC pressure.  See https://github.com/AnalyticalGraphicsInc/cesium/tree/master/Documentation/Contributors/CodingGuide#result-parameters-and-scratch-variables

lib/combineNodes.js
```diff
@@ -0,0 +1,290 @@
...
+            Matrix4.multiplyByPointAsVector(inverseTranspose, currentVector, newVector);
+        } else {
+            Matrix4.multiplyByPointAsVector(matrix, currentVector, newVector);
+        }
+
+        resultBuffer.writeFloatLE(newVector.x, bufferIndex);
+        resultBuffer.writeFloatLE(newVector.y, bufferIndex + 4);
+        resultBuffer.writeFloatLE(newVector.z, bufferIndex + 8);
+
+        //Update the min and max values if defined.
```
pjcozzi: This can be a lot less code by making `min`/`max` a `Cartesian3` (convert back to an array later) and using [Cartesian3.minimumByComponent](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/Cartesian3.js#L249).

pjcozzi:
Great start, @leerichard42.  I'll do another review once you've tested all the sample models.  Looking forward to getting this merged; this is one of the most important optimizations in the pipeline!

leerichard42:
Updated with the comments, but I think this stage has the same issues as removeUnusedVertices in that it assumes that the data is un-interleaved and takes entire chunks of data at a time, so this stage should also depend on the un-interleaving stage.

pjcozzi:
Sounds like we should get moving on the un-interleave stage.  @JoshuaStorm do you want to do this with some guidance from @lilleyse?

lasalvavida:
@leerichard42 with the addition of [uninterleaveAndPackBuffers](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/uninterleaveAndPackBuffers.js), I would recommend taking the same approach used in `removedUnusedVertices` and `quantizeAttributes`.

That is, just remove references to the data when it becomes obsolete, don't worry about trying to manipulate the buffer, then just call `uninterleaveAndPackBuffers` when you're done.

lasalvavida:
Closing in favor of #151

PR #53, opened 2016-06-01
Added gltf-pipeline command line interface

JoshuaStorm:
Added gltf-pipeline command line interface.
Refactored pipeline to separate file writing and optimizations. 
Added gltfPipelineSpec tests.

Pair programming with @likangning93


pjcozzi:
JSHint is failing: https://travis-ci.org/AnalyticalGraphicsInc/gltf-pipeline/builds/134604710

pjcozzi:
@lilleyse do you want to do a first review before I look at this?

@leerichard42 are you also interested in looking at this?

lilleyse:
> @lilleyse do you want to do a first review before I look at this?

Sure.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,106 @@
...
+            throw err;
+        }
+
+        var gltf;
+        if (fileExtension === '.glb') {
+            gltf = parseBinaryGltf(data);
+        }
+        else if (fileExtension === '.gltf') {
+            gltf = JSON.parse(data);
+            addPipelineExtras(gltf);
```
likangning93: We moved `addPipelineExtras` here from further down to resolve a bug where the gltfPipeline couldn't load glbs with the `KHR_binary_glTF extension`. It looks like glbs with that extension have an empty data URI, which isn't a problem for the pipeline if the data that should be there is in the assets' `extra` attribute. `parseBinaryGltf` was handling this, but then the later call to `addPipelineExtras` seemed to wipe out the `extra` attributes, causing an error later down in `loadGltfUris`.

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,53 @@
...
-var convertDagToTree = require('../').convertDagToTree;
-var combineMeshes = require('../').combineMeshes;
-var combinePrimitives = require('../').combinePrimitives;
-var removeUnusedVertices = require('../').removeUnusedVertices;
-var OptimizationStatistics = require('../').OptimizationStatistics;
 var Cesium = require('cesium');
+var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
+var writeGltf = require('../').writeGltf;
+var writeBinaryGltf = require('../').writeBinaryGltf;
```
lilleyse: Change these two includes to be like the new ones: `require('../lib/writeGltf');` (unless there was a good reason to keep it the original way)

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,53 @@
...
+var gltfPipeline = require('../lib/gltfPipeline');
 
 if (!defined(argv._[0]) || defined(argv.h) || defined(argv.help)) {
-	var help =
+    var help =
         'Usage: node ' + path.basename(__filename) + ' [path-to.gltf or path-to.bgltf] [OPTIONS]\n' +
-        '  -o=PATH  Write optimized glTF to the specified file.\n';
```
lilleyse: Put each of the options of their own line. For organization purposes, -o should go after -i.

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,53 @@
...
+        '  -i, input=PATH Read unoptimized glTF from the specified file.\n  -b, write binary glTF file.\n' +
+        '  -s, output non-embedded files.\n  -o, output=PATH write optimized glTF to the specified file.\n';
```
lilleyse: Change the -s description, maybe like this from ModelConverter:

'writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.'

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,53 @@
...
+var gltfPath = defaultValue(argv._[0], argv.i);
+var fileExtension = path.extname(gltfPath);
+var fileName = path.basename(gltfPath, fileExtension);
+var filePath = path.dirname(gltfPath);
 
-    gltf = loadGltfUris(gltf, filePath, function(err) {
-        if (err) {
-            throw err;
-        }
+var outputPath = argv.o;
```
lilleyse: Also worth mirroring -i with something like:

`var outputPath = defaultValue(argv._[1], argv.o);`

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,53 @@
...
 
-        var outputPath = argv.o;
-        if (!defined(outputPath)) {
-            // Default output.  For example, path/asset.gltf becomes path/asset-optimized.gltf
-            outputPath = path.join(filePath, fileName + '-optimized' + fileExtension);
+gltfPipeline(options, function(gltf) {
+    if (exportBinary) {
+        var outputExtension = path.extname(outputPath);
+        if (outputExtension !== ".glb") {
+            outputPath = path.basename(outputPath, outputExtension) + ".glb";
```
lilleyse: Usually we use single quotes around strings.

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,53 @@
...
         }
```
lilleyse: You can move this check to `writeBinaryGltf`.

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,53 @@
...
+        writeBinaryGltf(gltf, outputPath, true);
```
lilleyse: Note: we'll also want to pass `isSeparate` in here too soon.

index.js
```diff
@@ -26,5 +26,6 @@ module.exports = {
     combineMeshes : require('./lib/combineMeshes'),
     combinePrimitives : require('./lib/combinePrimitives'),
     removeUnusedVertices : require('./lib/removeUnusedVertices'),
-    OptimizationStatistics : require('./lib/OptimizationStatistics')
+    OptimizationStatistics : require('./lib/OptimizationStatistics'),
+    gltfPipeline : require('./lib/gltfPipeline')
```
lilleyse: I'm starting to wonder if `optimizeGltf` is a better name here now that it doesn't input and output files.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,106 @@
...
+var removeUnusedBuffers = require('../').removeUnusedBuffers;
+var removeUnusedBufferViews = require('../').removeUnusedBufferViews;
+var removeUnusedMaterials = require('../').removeUnusedMaterials;
+var removeUnusedSkins = require('../').removeUnusedSkins;
+var removeUnusedCameras = require('../').removeUnusedCameras;
+var removeUnusedTextures = require('../').removeUnusedTextures;
+var removeUnusedMeshes = require('../').removeUnusedMeshes;
+var removeUnusedNodes = require('../').removeUnusedNodes;
+var removeUnusedAccessors = require('../').removeUnusedAccessors;
+var removeUnused = require('./removeUnused');
```
lilleyse: Same comment here making the includes use the same style.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,106 @@
...
+var combineMeshes = require('./combineMeshes');
+var combinePrimitives = require('./combinePrimitives');
+var removeUnusedVertices = require('./removeUnusedVertices');
+var OptimizationStatistics = require('./OptimizationStatistics');
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+var defaultValue = Cesium.defaultValue;
+var DeveloperError = Cesium.DeveloperError;
+
+
```
lilleyse: Remove extra line.

lib/writeBinaryGltf.js
```diff
@@ -106,6 +106,8 @@ function writeBinaryGltf(gltf, outputPath, createDirectory, callback) {
     var scene = new Buffer(sceneString);
     var glb = Buffer.concat([header, scene, body], glbLength);
 
+    console.log(outputPath);
```
lilleyse: Remove this.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,30 @@
...
+var gltfPipeline = require('../../lib/gltfPipeline');
+var writeBinaryGltf = require('../../lib/writeBinaryGltf');
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+
+var gltfPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf';
+var glbPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.glb';
+
+describe('gltfPipeline', function() {
+    it('checks that gltf is defined when passing in a gltf file', function(done) {
```
lilleyse: Make the test name a little more descriptive, like `optimizes a gltf file`.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,30 @@
...
+        gltfPipeline({
+            inputPath : gltfPath,
+            separate : false
+        }, function(gltf) {
+            expect(defined(gltf)).toEqual(true);
```
lilleyse: Use `.toBeDefined()` instead. Also is there any way to make this more thorough? Maybe at least check that the gltfs don't equal each other.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,30 @@
...
+            done();
+        });
+    });
+
+    it('checks that gltf is defined when passing in a glb file', function(done) {
```
lilleyse: And here: "optimized a binary glTF file"

lib/gltfPipeline.js
```diff
@@ -0,0 +1,106 @@
...
+module.exports = gltfPipeline;
+
+function gltfPipeline(options, callback) {
```
lilleyse: Since this will be the starting point for projects that use gltf-pipeline as a library, we should also accept gltf objects directly. Definitely the obj2gltf will use it this way instead of writing a file first, then calling this. Maybe there should be another file that reads gltf files? @pjcozzi what do you think?

JoshuaStorm:
Updated, addressed @lilleyse's comment, aside from last (awaiting @pjcozzi's input).

Currently working to add more gltfPipelineSpec tests.

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,53 @@
 #!/usr/bin/env node
 'use strict';
 var fs = require('fs');
+var argv = require('yargs').argv;
```
lilleyse: Add yargs to package.json, and remove minimist.

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,53 @@
...
-var removeUnusedVertices = require('../').removeUnusedVertices;
-var OptimizationStatistics = require('../').OptimizationStatistics;
 var Cesium = require('cesium');
+var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
+var writeGltf = require('../').writeGltf;
+var writeBinaryGltf = require('../').writeBinaryGltf;
+var gltfPipeline = require('../lib/gltfPipeline');
 
 if (!defined(argv._[0]) || defined(argv.h) || defined(argv.help)) {
```
lilleyse: Now that we support -i for input files, `argv._[0]` may be undefined but shouldn't go the the help screen. As an alternative you can check for `process.argv.length < 3`.

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,51 @@
...
-var removeUnusedVertices = require('../').removeUnusedVertices;
-var OptimizationStatistics = require('../').OptimizationStatistics;
 var Cesium = require('cesium');
+var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
+var writeGltf = require('./writeGltf');
+var writeBinaryGltf = require('./writeBinaryGltf');
+var gltfPipeline = require('../lib/gltfPipeline');
 
 if (!defined(argv._[0]) || defined(argv.h) || defined(argv.help)) {
```
lilleyse: Adding this comment again because the previous comment is hidden:

Now that we support -i for input files, argv._[0] may be undefined but shouldn't go the the help screen. As an alternative you can check for process.argv.length < 3.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,106 @@
...
+        gltf = loadGltfUris(gltf, filePath, function (err) {
+            if (err) {
+                throw err;
+            }
+
+            combineMeshes(gltf);
+            combinePrimitives(gltf);
+
+            removeUnusedVertices(gltf);
+
```
leerichard42: I was actually thinking about moving removeUnusedVertices to come before the combine meshes/primitives/nodes stage, so that the actual scene data would be optimized before condensing the gltf structure.

index.js
pjcozzi: Keep in mind that it will have stages to do more than optimization, e.g., add AO.

lib/gltfPipeline.js
pjcozzi: Yes, exactly.  At the top-level there should be the command-line app, which just calls a function like this, which then calls a function that takes glTF in-memory as input.

Same deal for output - one function would just provide glTF in-memory and another would write to disk.

So the complete Node.js server -> obj2gltf -> gltf-pipeline flow will not write to disk.

Note that the challenge here is to define the in-memory data structure, i.e., it's one thing for the JSON to be loaded, but what about the .bin, .glsl, and images?  Perhaps gltf-pipeline should document its simple in-memory structure and provide helpers to read and write it (which it already does).

lib/gltfPipeline.js
```diff
@@ -0,0 +1,106 @@
...
+    if (!defined(gltfPath)) {
+        throw new DeveloperError('Input path is undefined.');
+    }
+    
+    var fileExtension = path.extname(gltfPath);
+    var fileName = path.basename(gltfPath, fileExtension);
+    var filePath = path.dirname(gltfPath);
+
+    if (fileExtension !== '.glb' && fileExtension !== '.gltf') {
+        throw new Error('Invalid glTF file.');
```
lilleyse: To be consistent this should be `DeveloperError`.

JoshuaStorm:
Updated. 
- [ ] Still need to cleanup callbacks in gltf-pipeline and loadGltfUris

pjcozzi:
@lilleyse let me know when this is ready for me to review.

likangning93:
Updated!

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,55 @@
...
 var defined = Cesium.defined;
-
-if (!defined(argv._[0]) || defined(argv.h) || defined(argv.help)) {
-	var help =
+var writeGltf = require('../lib/writeGltf');
+var writeBinaryGltf = require('../lib/writeBinaryGltf');
+var gltfPipeline = require('../lib/gltfPipeline');
+var addPipelineExtras = require('../lib/addPipelineExtras');
+var readGltf = require('../lib/readGltf');
+var DeveloperError = Cesium.DeveloperError;
```
lilleyse: Put this include near the other Cesium ones.

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,55 @@
...
+
+if (process.argv.length < 3 || defined(argv.h) || defined(argv.help)) {
+    var help =
         'Usage: node ' + path.basename(__filename) + ' [path-to.gltf or path-to.bgltf] [OPTIONS]\n' +
-        '  -o=PATH  Write optimized glTF to the specified file.\n';
+        '  -i, input=PATH Read unoptimized glTF from the specified file.\n ' +
+        '  -b, write binary glTF file.\n' +
+        '  -s, writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.\n  ' +
+        '  -o, output=PATH write optimized glTF to the specified file.\n';
```
lilleyse: Put the -o line after the -i.

bin/gltf-pipeline.js
```diff
@@ -1,114 +1,55 @@
...
-    process.stdout.write('BufferViews removed: ' + stats.numberRemoved.bufferViews + '\n');
-    process.stdout.write('Techniques removed: ' + stats.numberRemoved.techniques + '\n');
-    process.stdout.write('Textures removed: ' + stats.numberRemoved.textures + '\n');
-    process.stdout.write('Buffers removed: ' + stats.numberRemoved.buffers + '\n');
-    process.stdout.write('Programs removed: ' + stats.numberRemoved.programs + '\n');
-    process.stdout.write('Images removed: ' + stats.numberRemoved.images + '\n');
-    process.stdout.write('Samplers removed: ' + stats.numberRemoved.samplers + '\n');
-    process.stdout.write('Shaders removed: ' + stats.numberRemoved.shaders + '\n');
-}
+});
```
lilleyse: Not a big deal, but add a newline here. And wherever else you see it.

lib/loadGltfUris.js
```diff
@@ -15,21 +15,11 @@ function loadGltfUris(gltf, basePath, callback) {
     async.each(['buffers', 'images', 'shaders'], function(name, asyncCallback) {
             loadURI(gltf, basePath, name, asyncCallback);
         }, function(err) {
```
lilleyse: There's some extra indentation in this area.

lib/loadGltfUris.js
```diff
@@ -15,21 +15,11 @@ function loadGltfUris(gltf, basePath, callback) {
...
-                if (callback) {
-                    callback(err);
-                }
-                else{
-                    throw err;
-                }
-            }
-            else if (callback) {
-                callback();
+            if (defined(callback)) {
```
lilleyse: The callback should always be defined, right?

lib/readGltf.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+var path = require('path');
+var fs = require('fs');
+var parseBinaryGltf = require('./parseBinaryGltf');
+var addPipelineExtras = require('./addPipelineExtras');
+var loadGltfUris = require('../lib/loadGltfUris');
```
lilleyse: Simplify to `var loadGltfUris = require('./loadGltfUris');`

lib/readGltf.js
```diff
@@ -0,0 +1,47 @@
...
+var defined = Cesium.defined;
+var DeveloperError = Cesium.DeveloperError;
+
+module.exports = readGltf;
+
+function readGltf(gltfPath, callback) {
+
+    var fileExtension = path.extname(gltfPath);
+    var filePath = path.dirname(gltfPath);
+
```
lilleyse: Remove extra line.

lib/readGltf.js
```diff
@@ -0,0 +1,47 @@
...
+            addPipelineExtras(gltf);
+        }
+        loadGltfUris(gltf, filePath, function(err, gltf) {
+            if (err) {
+                throw err;
+            }
+            callback(gltf);
+        });
+    });
+}
```
lilleyse: Newline.

lib/writeBinaryGltf.js
```diff
@@ -15,13 +15,19 @@ var removePipelineExtras = require('./removePipelineExtras');
...
+        outputPath = path.basename(outputPath, outputExtension) + '.glb';
+    }
+    // Create the output directory if specified
     if (createDirectory) {
         outputPath = path.join(path.dirname(outputPath), 'output', path.basename(outputPath));
         mkdirp.sync(path.dirname(outputPath));
     }
 
-    //Create the special binary buffer from the existing buffers
+
```
lilleyse: Remove extra line.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,33 @@
...
+var readGltf = require('../../lib/readGltf');
+var writeBinaryGltf = require('../../lib/writeBinaryGltf');
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+
+var gltfPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf';
+var glbPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.glb';
+
+describe('gltfPipeline', function() {
+    it('optimizes a gltf file, checks that a new gltf is defined when passing in a gltf file', function(done) {
```
lilleyse: For this title, just `optimizes a gltf file` should be enough. Same comment below.

specs/lib/readGltfSpec.js
```diff
@@ -0,0 +1,54 @@
...
+                readGltf(undefined, function() {
+                    done();
+                });
+            }
+            catch (err) {
+                expect(err).toBeDefined();
+                expect(err.message).toEqual('Input path is undefined.');
+                throw err;
+            }
+        }).toThrow();
```
lilleyse: These two tests can be simplified, for example this one could just be:

```
expect(function() {
    readGltf(undefined);
}).toThrow();
```

lib/loadGltfUris.js
likangning93: `combinePrimitivesSpec` has a couple specs that run `loadGltfUris` without a callback. I'll see if I can change these.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,33 @@
...
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+
+var gltfPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf';
+var glbPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.glb';
+
+describe('gltfPipeline', function() {
+    it('optimizes a gltf file', function(done) {
+        readGltf(gltfPath, function(gltf) {
+            var stringCopy = JSON.stringify(gltf);
```
lilleyse: Now that I've looked more carefully, we should just use the `clone` library like some other tests do. Look at `writeShaderSpec` for example.

lilleyse:
@pjcozzi You can review this whenever you're ready.

lilleyse:
Actually there are still a few spec changes in progress, I'll update when they're ready.

likangning93:
I pushed some changes to the combinePrimitivesSpec. Some documentation on what happened:

We found out by changing `expect` to `expect.not` to see if all the tests were actually running properly, which revealed the `expect`s in `does not affect single primitives` and `does not combine two primitives` were never getting hit. This seemed to be due to an error in calls to `writeGltf`, which were preventing the callbacks containing the `expect`s to run. Aside: @leerichard42 , what was the `writeGltf` here for?
Figuring this out was obfuscated heavily by other tests failing with bizarre outputs, like `expect`s receiving whole gltf objects instead of the small values they were supposed to be comparing with. We suspect what might have happened was that jasmine somehow blocks file output and leaves those unwritten gltfs floating around somewhere that they can interfere with the other tests.

We also found that using `clone` here, including `Cesium`'s recursive clone, would run into un-clonable objects, hence the use of `stringify` in some of the updated tests.

leerichard42:
The writeGltf calls in the spec were intended to return and write out a gltf object with the extras removed and data uris written out, so that I would be able to visually confirm the file in the converter.

JoshuaStorm:
Updated. Refactored gltfPipeline to follow @pjcozzi's guidelines more closely, my still need some work.
Changed index.js to only expose those specific files. 

Changed some require syntax in a few specs that were creating issues with Jasmine.

likangning93:
Small update: I added the `options` and `callback` parameters back into the functions in  `gltfPipeline` as per @lilleyse's request

pjcozzi:
What's the latest on this?  Do you need me to review yet?

lilleyse:
I still need to do a review of the latest changes, I'll tell you when I'm ready.

index.js
```diff
@@ -1,30 +1,9 @@
...
+    readGltf : require('./lib/readGltf'),
     writeBinaryGltf : require('./lib/writeBinaryGltf'),
-    addPipelineExtras : require('./lib/addPipelineExtras'),
-    convertDagToTree : require('./lib/convertDagToTree'),
-    combineMeshes : require('./lib/combineMeshes'),
-    combinePrimitives : require('./lib/combinePrimitives'),
-    removeUnusedVertices : require('./lib/removeUnusedVertices'),
-    OptimizationStatistics : require('./lib/OptimizationStatistics')
+    writeGltf : require('./lib/writeGltf'),
+    writeSource : require('./lib/writeSource')
```
lilleyse: Should this be exposed here?

lib/gltfPipeline.js
```diff
@@ -0,0 +1,108 @@
...
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+var defaultValue = Cesium.defaultValue;
+
+module.exports = {
+    processJSON : processJSON,
+    processFile : processFile,
+    processJSONToDisk : processJSONToDisk,
+    processFileToDisk : processFileToDisk
+};
```
lilleyse: In order to handle the case where the user has a gltf json in memory but not with the extras, `processJSONToDisk` and `processJSON` should take in `gltf` instead of `gltfWithExtras`. They can add the extras to the gltf then call a function named `processJSONWithExtras` which actually does the pipeline steps (like what `processJSON` does currently). The new `processJson` could also remove pipeline extras before sending back the gltf.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,108 @@
...
+
+function processJSON(gltfWithExtras, options, callback) {
```
lilleyse: Maybe one option to start out with here is `printStats`, which is `false` by default. Probably not worth adding to the command-line options.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,108 @@
...
+
+function processFileToDisk(inputPath, outputPath, options, callback) {//isBinary, isEmbedded) {
+    processFile(inputPath, options, function(gltf) {
+        var isBinary = defaultValue(options.isBinary, false);
+        var isEmbedded = defaultValue(options.isEmbedded, false);
+        if (isBinary) {
+            writeBinaryGltf(gltf, outputPath, true, callback);
+        } else {
+            writeGltf(gltf, outputPath, isEmbedded, true, callback);
+        }
```
lilleyse: A lot of the code in `processFileToDisk` and `processJSONToDisk` can be factored out into a common function.

lib/readGltf.js
```diff
@@ -0,0 +1,46 @@
...
+var DeveloperError = Cesium.DeveloperError;
+
+module.exports = readGltf;
+
+function readGltf(gltfPath, callback) {
+
+    var fileExtension = path.extname(gltfPath);
+    var filePath = path.dirname(gltfPath);
+
+    if (!defined(gltfPath)) {
```
lilleyse: This check should go first.

specs/lib/combinePrimitivesSpec.js
```diff
@@ -1,99 +1,84 @@
...
+
             expect(doubleBoxToCombine.accessors['meshTest_INDEX_accessor_0']).toEqual({
                 "bufferView": "meshTest_INDEX_bufferView_0",
                 "byteOffset": 0,
                 "byteStride": 0,
                 "componentType": 5123,
                 "type": "SCALAR",
-                "count": 516
+                "count": 516,
+                "extras": { "_pipeline": { "deleteExtras": true } }
```
lilleyse: Is there a way to get around checking this, it just seems out of place. Maybe a call to `removePipelineExtras` beforehand?

specs/lib/combinePrimitivesSpec.js
```diff
@@ -102,15 +87,17 @@ describe('combinePrimitives', function() {
                 "type": "VEC3",
                 "count": 320,
                 "max": [0.5, 0.5, 0.5],
-                "min": [-0.5, -0.5, -0.5]
+                "min": [-0.5, -0.5, -0.5],
+                "extras": { "_pipeline": { "deleteExtras": true } }
```
lilleyse: Same here.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+var clone = require('clone');
```
lilleyse: This reminds me that there are still other areas that use `Cesium.clone`. Those should probably switch over to use this library as well.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,48 @@
...
+        var options = {};
+        readGltf(gltfPath, function(gltf) {
+            gltfCopy = clone(gltf);
+            processJSON(gltf, options, function(gltf) {
+                expect(gltf).toBeDefined();
+                expect(gltf).not.toBe(gltfCopy);
+                done();
+            });
+        });
+        processFile(gltfPath, options, function(gltf) {
```
lilleyse: Make this second check a different test.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,48 @@
...
+        var options = {};
+        readGltf(glbPath, function(gltf) {
+            gltfCopy = clone(gltf);
+            processJSON(gltf, options, function(gltf) {
+                expect(gltf).toBeDefined();
+                expect(gltf).not.toBe(gltfCopy);
+                done();
+            });
+        });
+        processFile(glbPath, options, function(gltf) {
```
lilleyse: Also make this a separate test.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,48 @@
...
+            expect(gltf).toBeDefined();
+            expect(gltf).not.toBe(gltfCopy);
+            done();
+        });
+    });
+});
```
lilleyse: This file tests `processFile`, `processJson`, but we should also test all the other functions in `gltfPipeline`. Also in order to test the write function (json and binary), we can spy on `fs.writeFile` and check that the outputPath argument is correct. (Check out this for an example: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Specs/Scene/ShadowMapSpec.js#L948-L967, but don't use `callThrough` because we don't want to save the file).

specs/lib/readGltfSpec.js
```diff
@@ -0,0 +1,36 @@
...
+            expect(gltf).toBeDefined();
+            done();
+        });
+    });
+
+    it('throws error when an input path is undefined', function(done) {
+        expect(function() {
+            readGltf(undefined);
+        }).toThrow();
+        done();
```
lilleyse: I think it's ok to remove `done()` if the test isn't asynchronous. Also remove the `done` argument.

specs/lib/readGltfSpec.js
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    it('throws error when file extension of input file is invalid', function(done) {
+        expect(function() {
+            readGltf(invalidPath);
+        }).toThrow();
+        done();
```
lilleyse: Same comment here.

specs/lib/readGltfSpec.js
JoshuaStorm: I thought that were the case, but in these two specific tests they timeout if the `done()` is not there. I'm not exactly sure why.

lilleyse: If you remove the `done` argument it should be fine.

lilleyse:
If anyone has feedback on the names `processJSON`, `processFile`, `processJSONToDisk`, and `processFileToDisk` that would be helpful. I think they are pretty good, but could possibly be better.

lib/gltfPipeline.js
JoshuaStorm: With `ProcessJSON` should we just assume that the gltf URIs have already been loaded into the gltf file? If not, we would need an input path as an argument, sort of defeating one purpose of using a json from memory.

lilleyse: Ah.. I forgot about that. Maybe options could optionally include a `resourcePath`?

JoshuaStorm: Do you mean to optionally include a `resourcePath`, but otherwise assume that the URIs are already loaded?

lilleyse: Yup that sounds fine to me.

JoshuaStorm:
Updated!

index.js
```diff
@@ -1,30 +1,9 @@
...
-    addPipelineExtras : require('./lib/addPipelineExtras'),
-    convertDagToTree : require('./lib/convertDagToTree'),
-    combineMeshes : require('./lib/combineMeshes'),
-    combinePrimitives : require('./lib/combinePrimitives'),
-    removeUnusedVertices : require('./lib/removeUnusedVertices'),
-    OptimizationStatistics : require('./lib/OptimizationStatistics')
-};
\ No newline at end of file
+    writeGltf : require('./lib/writeGltf'),
+    writeSource : require('./lib/writeSource')
```
lilleyse: I think my comment from this got lost, but do we need `writeSource` in here?

lib/gltfPipeline.js
```diff
@@ -0,0 +1,130 @@
...
+var defined = Cesium.defined;
+
+module.exports = {
+    processJSON : processJSON,
+    processJSONWithExtras : processJSONWithExtras,
+    processFile : processFile,
+    processJSONToDisk : processJSONToDisk,
+    processFileToDisk : processFileToDisk,
+    writeFile : writeFile
+};
```
lilleyse: If possible we should try not to expose `processJSONWithExtras` and `writeFile` since they are really just internal to this file. Some tests may have to adjust for this.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,130 @@
...
+
+function processJSON(gltf, options, callback) {
+    if (defined(options.resourcePath)) {
+        loadGltfUris(gltf, options.resourcePath, function(err, gltf) {
+            if (err) {
+                throw err;
+            }
+            processJSONWithExtras(gltf, options, function(gltf) {
+                callback(gltf);
+            });
```
lilleyse: You can simplify this to `processJSONWithExtras(gltf, options, callback)`

lib/gltfPipeline.js
```diff
@@ -0,0 +1,130 @@
...
+        });
+    } else {
+        // If a resourcePath is not passed in, assume JSON is already embedded
+        processJSONWithExtras(gltf, options, function(gltf) {
```
lilleyse: And here too.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,130 @@
...
+            callback(gltf);
+        });
+    }
+}
+
+
```
lilleyse: Remove extra line.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,130 @@
...
+    var isBinary = defaultValue(options.isBinary, false);
+    var isEmbedded = defaultValue(options.isEmbedded, false);
+    if (isBinary) {
+        writeBinaryGltf(gltf, outputPath, true, callback);
+    } else {
+        writeGltf(gltf, outputPath, isEmbedded, true, callback);
+    }
+}
+
+function processJSONToDisk(gltf, outputPath, options, callback) {// isBinary, isEmbedded) {
```
lilleyse: It's fine to remove the commented code here.

lib/gltfPipeline.js
```diff
@@ -0,0 +1,130 @@
...
+    processJSON(gltf, options, function(gltf) {
+        writeFile(gltf, outputPath, options, callback);
+    });
+}
+
+function processFileToDisk(inputPath, outputPath, options, callback) {//isBinary, isEmbedded) {
```
lilleyse: And here.

lib/readGltf.js
```diff
@@ -0,0 +1,45 @@
...
+
+module.exports = readGltf;
+
+function readGltf(gltfPath, callback) {
+    var fileExtension = path.extname(gltfPath);
+    var filePath = path.dirname(gltfPath);
+
+    if (!defined(gltfPath)) {
+        throw new DeveloperError('Input path is undefined.');
+    }
```
lilleyse: Put this check above since `fileExtension` and `filePath` expect it to be defined.

lib/writeGltf.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-var fs = require('fs');
+var fs = require('fs');
```
lilleyse: Extra space here.

specs/lib/combinePrimitivesSpec.js
lilleyse: Handle this comment as well, and the one below.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,135 @@
...
+var removePipelineExtras = require('../../lib/removePipelineExtras');
+var addPipelineExtras = require('../../lib/addPipelineExtras');
+var writeBinaryGltf = require('../../lib/writeBinaryGltf');
+
+var gltfPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf';
+var glbPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.glb';
+var outputPath = './output/';
+
+describe('gltfPipeline', function() {
+    it('optimizes a gltf JSON given an embedded JSON', function() {
```
lilleyse: A clearer description might be `optimizes a gltf JSON with embedded resources`.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,135 @@
...
+        readGltf(gltfPath, function(gltf) {
+            gltfCopy = clone(gltf);
+            processJSON(gltf, options, function(gltf) {
+                expect(gltf).toBeDefined();
+                expect(gltf).not.toBe(gltfCopy);
+            });
+        });
+    });
+    
+    it('optimizes a gltf JSON given a resourcePath', function() {
```
lilleyse: And here `optimizes a gltf JSON with external resources`.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,135 @@
...
+        var gltf;
+        var gltfCopy;
+        var options = { 'resourcePath' : path.dirname(gltfPath) };
+        fs.readFile(gltfPath, function(err, data) {
```
lilleyse: Why not use `readGltf` here?

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,135 @@
...
+        var gltfCopy;
+        var options = {};
+        readGltf(glbPath, function(gltf) {
+            gltfCopy = clone(gltf);
+            processJSONWithExtras(gltf, options, function(gltf) {
+                expect(gltf).toBeDefined();
+                expect(gltf).not.toBe(gltfCopy);
+            });
+        });
+    });
```
lilleyse: You could probably drop these two tests (related to my other comment).

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,135 @@
...
+
+    it('optimizes a glTF file', function() {
+        var gltfCopy;
+        var options = {};
+        readGltf(gltfPath, function(gltf) {
+            gltfCopy = gltf;
```
lilleyse: Not actually doing a clone here.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,135 @@
...
+                expect(gltf).not.toBe(gltfCopy);
+            });
+        });
+    });
+
+    it('optimizes a glb file', function() {
+        var gltfCopy;
+        var options = {};
+        readGltf(glbPath, function(gltf) {
+            gltfCopy = gltf;
```
lilleyse: And here.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,135 @@
...
+            processFile(glbPath, options, function (gltf) {
+                expect(gltf).toBeDefined();
+                expect(gltf).not.toBe(gltfCopy);
+            });
+        });
+    });
```
lilleyse: All of these tests should have `done` callbacks.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,135 @@
...
+    it('will write a file', function(done) {
+        var spy = spyOn(fs, 'writeFile').and.callFake(function(file, data, callback) {
+            callback();
+        });
+        var options = {};
+        processFileToDisk(gltfPath, outputPath, options, function() {
+            expect(spy).toHaveBeenCalled();
+            done();
+        });
+    });
```
lilleyse: These two tests can be merged into one. Don't forget to check that the expected output path.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,135 @@
...
+        });
+        var options = {};
+        readGltf(gltfPath, function(gltf) {
+            processJSONToDisk(gltf, outputPath, options, function() {
+                expect(spy).toHaveBeenCalled();
+                done();
+            });
+        });
+    });
+});
```
lilleyse: Add newline.

specs/lib/readGltfSpec.js
```diff
@@ -0,0 +1,32 @@
...
+        readGltf(gltfPath, function(gltf) {
+            expect(gltf).toBeDefined();
+        });
+    });
+
+    it('parses a .glb input path, checks that a gltf JSON is defined', function() {
+        readGltf(glbPath, function(gltf) {
+            expect(gltf).toBeDefined();
+        });
+    });
```
lilleyse: You still need the `done` callback for these two tests.

specs/lib/gltfPipelineSpec.js
JoshuaStorm: I had an issue with trying to check for an expected output path. Because multiple files are written, it is difficult to test exactly what the output path would be (or at least I am not sure how I would go about it). 

For example: 
I tried `expect(spy).toHaveBeenCalledWith(outputPath, jasmine.any(String), jasmine.any(Function));` but the result ends up being

 `[ outputPath/output.gltf, <a stringified JSON blob>, <a function>], [ outputPath/output0VS.glsl, <a stringified JSON>, <a function>], ... `

Overall, I am just unsure how I can write a Jasmine expect for a spy that seems to see multiple outputs with different expected files. Maybe something with regular expressions?

lilleyse: Oh ok, it may help if you set the `isEmbedded` options to true. Or `isEmbedded` should default to true instead of false..

JoshuaStorm: That is a way easier solution, thanks!

specs/lib/gltfPipelineSpec.js
JoshuaStorm: readGltf includes the `loadGltfUris' step, and this test is checking that `processJSON` properly handling a JSON that has yet to have that step applied

specs/lib/gltfPipelineSpec.js
JoshuaStorm: I am having issues with merging these two tests into one where it is throwing: `Error: EISDIR: illegal operation on a directory, open 'output/output'
    at Error (native)`

I initially thought this was because it was trying to create a new directory in a path that already exists, but even after adding an option to not create a new directory I am still getting the same error.

Any thoughts? Should I just leave it as separate tests?

JoshuaStorm:
Updated.

I believe I addressed all the comments, aside from the one about combining the two file writing tests in gltfPipelineSpec (see comment above).

specs/lib/gltfPipelineSpec.js
lilleyse: I'm tried calling just the second one and it passes for me.

specs/lib/gltfPipelineSpec.js
lilleyse: Ah right.

specs/lib/gltfPipelineSpec.js
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    it('optimizes a glb file', function(done) {
+        var gltfCopy;
+        var options = {};
+        readGltf(glbPath, function(gltf) {
+            gltfCopy = clone(gltf);
+            processFile(glbPath, options, function (gltf) {
+                expect(gltf).toBeDefined();
+                expect(gltf).not.toBe(gltfCopy);
```
lilleyse: I missed these before, but it should be `.not.toEqual` for all these tests. (`.toBe` checks that they are the same object, `.toEqual` checks values, like `===` vs `==`).

JoshuaStorm: That explains why they weren't failing when I was forgetting to clone them. Thanks for the explanation

specs/lib/gltfPipelineSpec.js
```diff
@@ -112,24 +92,23 @@ describe('gltfPipeline', function() {
         var spy = spyOn(fs, 'writeFile').and.callFake(function(file, data, callback) {
             callback();
         });
-        var options = {};
+        var options = { 'createDirectory' : false };
         processFileToDisk(gltfPath, outputPath, options, function() {
-            expect(spy).toHaveBeenCalled();
+            expect(spy).toHaveBeenCalledWith('./output/', jasmine.any(String), jasmine.any(Function));
```
lilleyse: Cleaner with something like `expect(spy.calls.first().args[0]).toEqual('./output/');`

lilleyse:
Just those few comments, this is getting close.

JoshuaStorm:
Updated

specs/lib/gltfPipelineSpec.js
lilleyse: The new changes look good, just check this last thing and it will be ready.

JoshuaStorm:
Addressed those last two comments about the `gltfPipelineSpec` and `combinePrimitivesSpec`. Am I missing anything?

lilleyse:
I think that's everything, I'm going to merge and if we missed anything we can make another PR.

pjcozzi:
This is a good start for the command-line interface; is the ability to configure custom pipelines via command-line, JSON input, and a function taking a JSON object coming as a separate PR?

JoshuaStorm:
Unless I am misunderstanding, we currently have functions to take a JSON as input in the `gltfPipeline.js` file --`processJSON` and `processJSONToDisk`.

Also, could you elaborate a bit more on what you mean by "ability to configure custom pipelines via command-line"

Thanks!

pjcozzi:
From #1:

> * [ ] Command-line and JSON input, e.g., like pdal (see [here](http://www.pdal.io/apps.html) and [here](http://www.pdal.io/pipeline.html))

lasalvavida:
@JoshuaStorm I could be wrong, but I think the idea is that the pipeline should be able to output to stdout, read from stdin, and specify specific operations to perform. Then you could chain commands together using pipes. Something like:
`./bin/gltf-pipeline.js --stdout myModel.gltf --stage addDefaults | ./bin/gltf-pipeline.js --stdin --stage quantizeAttributes`

PR #55, opened 2016-06-03
Added KHR_materials_common materials/techniques/shaders to addDefaults

likangning93:
@JoshuaStorm and I wrote this to have `defaultMaterials` try to match the existing gltf material values to one of the `KHR_materials_common` materials, apply the appropriate extension, and then pass the gltf to `modelMaterialsCommon` at the end of `addDefaults`.

Also added some tests to `addDefaultsSpec`.

See https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/1


pjcozzi:
@lilleyse can you please do an initial review here?

likangning93:
Updated. When testing assimp exported gltf (gltf without techniques) we ran into a problem where cesium's `modelMaterialsCommon` would produce shaders with empty string `uri`s. We got around it with a customized `modelMaterialsCommon` in the lib folder, similar to what's in OBJ2GLTF.

Unfortunately assimp exported gltfs still don't seem to work, but the test case we've been using is also weird. It has buffer views for indices, positions, normals, and uvs but they all point at the same buffer and all have offset 0. Very strange. See attached.

[assimp.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/298229/assimp.zip)

pjcozzi:
For assimp, have you tried just the box COLLADA model, https://github.com/KhronosGroup/glTF/tree/master/sampleModels/Box/collada

likangning93:
I gave it a try, but it seems to have the same problem with the bufferViews. It actually runs into the same error in the pipeline and when I try to view it in the online COLLADA to glTF convertor.

pjcozzi:
What about using a simple OBJ file with assimp, perhaps from https://github.com/SaschaWillems/Vulkan/tree/master/models

Just trying to understand if it is the assimp importer or assimp glTF exporter.

likangning93:
I tried it with a basic cube and with hammardillo, but it seems to be the same problem, `bufferViews` that all start at 0.

likangning93:
I'm pretty sure it's in the assimp gltf exporter. I'm going to try and take a look at their code.

pjcozzi:
If you figure it out in a reasonable amount of time, open a pull request; otherwise, some an issue to the assimp repo and include sample data and simple instructions to reproduce it.

lib/WebGLConstants.js
```diff
@@ -0,0 +1,300 @@
+"use strict";
```
lilleyse: Talked offline, but you use Cesium's `defined`, `defaultValue`, and `WebGLConstants` instead for `modelMaterialsCommon`.

lib/addDefaults.js
```diff
@@ -190,6 +191,59 @@ function lightDefaults(gltf) {
     }
 }
 
+function checkIfFloatVec4(value) {
+    if (defined(value) && value.length === 4) {
+        if (typeof(value[0]) === 'number' &&
+            typeof(value[1]) === 'number' &&
+            typeof(value[2]) === 'number' &&
+            typeof(value[3]) === 'number') return value;
```
lilleyse: Style thing - put the return on the next line.

lib/addDefaults.js
```diff
@@ -190,6 +191,59 @@ function lightDefaults(gltf) {
...
+    }
+    else return [0, 0, 0, 1];
+}
+
+function guessExtensionTechnique(material) {
+    // Replaces the material technique, extension, and values with a close equivalent from KHR_materials_common.
+    // If no equivalent can be found, defaults to blinn for specular materials and lambert otherwise
+    var technique = 'LAMBERT';
+    var useBlinn = false;
```
lilleyse: This isn't being used.

lib/addDefaults.js
```diff
@@ -190,6 +191,59 @@ function lightDefaults(gltf) {
...
+    var values = defined(material.values) ? material.values : {};
+
+    var ambient = checkIfFloatVec4(values.ambient);
+    var diffuse = typeof(values.diffuse) === 'string' ? values.diffuse : checkIfFloatVec4(values.diffuse);
+    var doubleSided = typeof(values.doubleSided) === 'boolean' ? values.doubleSided : false;
```
lilleyse: Would a gltf material actually have a `doubleSided` property? Also this could just be made a bit cleaner as `defaultValue(values.doubleSided, false)`

lib/addDefaults.js
```diff
@@ -190,6 +191,59 @@ function lightDefaults(gltf) {
...
+    var emission = typeof(values.emission) === 'string' ? values.emission : checkIfFloatVec4(values.emission);
+    var specular = typeof(values.specular) === 'string' ? values.specular : checkIfFloatVec4(values.specular);
+    var shininess = typeof(values.shininess) === 'number' ? values.shininess : 0.0;
```
lilleyse: And `defaultValue(values.shininess, 0.0)` here.

lib/addDefaults.js
```diff
@@ -190,6 +191,59 @@ function lightDefaults(gltf) {
...
+    if (defined(specular) || defined(shininess)) {
```
lilleyse: Won't both of these always be defined because of what's above? Probably check `values.specular` and `values.shininess` here.

lib/addDefaults.js
```diff
@@ -190,6 +191,59 @@ function lightDefaults(gltf) {
...
+        technique = 'BLINN';
+    }
+    var transparency = typeof(values.transparency) === 'number' ? values.transparency : 1.0;
+    var transparent = typeof(values.transparent) === 'boolean' ? values.transparent : false;
```
lilleyse: Same code simplify comments like above for these two. Also will these be part of the `Material` also? You could also check both of these with the diffuse.a channel.

lib/addDefaults.js
likangning93: Changed to always false in this case, as discussed offline.

lib/addDefaults.js
likangning93: We're keeping the `typeof` checks to make sure something like `shininess` as a texture map in a more customized gltf doesn't make its way into the `KHR_materials_common`.

likangning93:
Ok. Updated!

lilleyse:
Nice, looks good!

lilleyse:
@pjcozzi this is ready whenever you want to look

lib/addDefaults.js
```diff
@@ -190,6 +191,59 @@ function lightDefaults(gltf) {
...
 
+function checkIfFloatVec4(value) {
+    if (defined(value) && value.length === 4) {
+        if (typeof(value[0]) === 'number' &&
+            typeof(value[1]) === 'number' &&
+            typeof(value[2]) === 'number' &&
+            typeof(value[3]) === 'number') {
+            return value;
+        }
+    }
```
pjcozzi: To follow our style guide, just write
```javascript
return [0, 0, 0, 1];
```
And drop the `else`; it is not needed.

lib/modelMaterialsCommon.js
```diff
@@ -0,0 +1,754 @@
+"use strict";
+var Cesium = require('cesium');
+var WebGLConstants = Cesium.WebGLConstants;
+var defined = Cesium.defined;
+var defaultValue = Cesium.defaultValue;
+
+module.exports = modelMaterialsCommon;
+
+function webGLConstantToGlslType(webGLValue) {
```
pjcozzi: Could be useful as a standalone function in a separate file...not sure if you want to move it now.

lib/modelMaterialsCommon.js
```diff
@@ -0,0 +1,754 @@
...
+                    case 'ambient':
+                        var ambient = light.ambient;
+                        result[lightBaseName + 'Color'] = {
+                            type: WebGLConstants.FLOAT_VEC3,
+                            value: ambient.color
+                        };
+                        break;
+                    case 'directional':
+                        var directional = light.directional;
+                        result[lightBaseName + 'Color'] =
```
pjcozzi: Throughout this file, to follow our style lines like this should look like:
```javascript
result[lightBaseName + 'Color'] = {
```
with the `{` on the same line as the `=`.

lib/modelMaterialsCommon.js
```diff
@@ -0,0 +1,754 @@
...
+                }
+                ++lightCount;
+            }
+        }
+    }
+
+    return result;
+}
+
+function getNextId(dictionary, baseName, startingCount) {
```
pjcozzi: @leerichard42 are there other parts of gltf-pipeline where we could replace custom code with a call to this function?

lib/modelMaterialsCommon.js
likangning93: Cesium has its own `modelMaterialsCommon` with some implementation differences, so I'm guessing the change would be better happening over there.

lib/modelMaterialsCommon.js
```diff
@@ -0,0 +1,754 @@
...
+        if (techniqueParameters.hasOwnProperty(paramName)) {
+            var param = techniqueParameters[paramName];
+            techniqueUniforms['u_' + paramName] = paramName;
+            var arraySize = defined(param.count) ? '['+param.count+']' : '';
+            if (((param.type !== WebGLConstants.FLOAT_MAT3) && (param.type !== WebGLConstants.FLOAT_MAT4)) ||
+                param.useInFragment) {
+                fragmentShader += 'uniform ' + webGLConstantToGlslType(param.type) + ' u_' + paramName + arraySize + ';\n';
+                delete param.useInFragment;
+            }
+            else {
```
pjcozzi: `else` should be on the same line as the previous `}`.

pjcozzi: Same comment throughout this file.

pjcozzi:
Looks like test coverage could be improve a good bit: https://coveralls.io/github/AnalyticalGraphicsInc/gltf-pipeline?branch=materialsStageKHR

lib/modelMaterialsCommon.js
pjcozzi: Changes to both would be fine for now; longer term, Cesium will depend on some files in gltf-pipeline.

Also, for Cesium, I would not move this to a separate file; just make the trivial style changes.

likangning93:
Updated! Apparently I misunderstood `extensions` versus `extensionsUsed,` so fixing that and adding some more to the tests I had helped increase the coverage.

pjcozzi:
@lilleyse please merge when ready.

lilleyse:
Looks ready to me.

lilleyse:
Ah forgot, thanks!

PR #56, opened 2016-06-03
Web3d quantized attributes

lasalvavida:
Adds support for quantized attributes.

Related to [#308](https://github.com/AnalyticalGraphicsInc/agi-asset-pipeline/pull/308)


lasalvavida:
@pjcozzi

pjcozzi:
I'll do a first review here since I have also been looking at https://github.com/AnalyticalGraphicsInc/cesium/pull/3891, but @lilleyse can you please do the final review and merge?

pjcozzi:
@JoshuaStorm @likangning93 can you make a new branch and include this in your interface?

lib/getBinaryGltf.js
```diff
@@ -0,0 +1,102 @@
...
+var defined = Cesium.defined;
+var defaultValue = Cesium.defaultValue;
+var sizeOf = require('image-size');
+var mime = require('mime');
+var mergeBuffers = require('./mergeBuffers');
+var removePipelineExtras = require('./removePipelineExtras');
+
+module.exports = getBinaryGltf;
+
+function getBinaryGltf(gltf, callback) {
```
pjcozzi: Are there unit tests for this?

lasalvavida: There are unit tests for writeBinaryGltf which uses this directly.

lib/getBinaryGltf.js
```diff
@@ -0,0 +1,102 @@
...
+    gltf.buffers = defaultValue(gltf.buffers, {});
+    mergeBuffers(gltf, 'binary_glTF');
+
+    var bufferViews = gltf.bufferViews;
+    var buffers = gltf.buffers;
+    var currentOffset = buffers.binary_glTF.byteLength;
+    var body = buffers.binary_glTF.extras._pipeline.source;
+    var currentBinaryView = 0;
+    //Update object with KHR_binary_glTF properties and add to body and bufferViews
+    function updateBinaryObject(name) {
```
pjcozzi: I think it is better to put this function outside of the outer `getBinaryGltf` function and just pass in any data needs.  I expect it will get optimized better.

lib/getBinaryGltf.js
```diff
@@ -0,0 +1,102 @@
...
+        var objects = gltf[name];
+        if (defined(objects)) {
+            for (var objectId in objects) {
+                if (objects.hasOwnProperty(objectId)) {
+                    var object = objects[objectId];
+
+                    //Update object with binary format
+                    object.uri = "data:,";
```
pjcozzi: Use `'`.

lib/getBinaryGltf.js
```diff
@@ -0,0 +1,102 @@
...
+                    var bufferViewKeys = Object.keys(bufferViews);
+                    while (bufferViewKeys.indexOf('binary_bufferView' + currentBinaryView) != -1) {
+                        currentBinaryView++;
+                    }
+
+                    var objectSource = object.extras._pipeline.source;
+                    var bufferViewId = 'binary_bufferView' + currentBinaryView;
+                    KHR_binary_glTF.bufferView = bufferViewId; //Create bufferview
+                    bufferViews[bufferViewId] = {
+                        "buffer": "binary_glTF",
```
pjcozzi: Drop `"` for properties.  Use `'` for the string values.

I don't think gltf-pipeline is completely consistent here, but this is the standard Cesium.  There's some exceptions, for example, if the entire object is JSON (we do this for CZML examples in Sandcastle).

lib/getBinaryGltf.js
```diff
@@ -0,0 +1,102 @@
...
+    header.writeUInt32LE(glbLength, 8);
+    header.writeUInt32LE(sceneLength, 12);
+    header.writeUInt32LE(0, 16);
+
+    //Create scene buffer and overall buffer
+    var scene = new Buffer(sceneString);
+    var glb = Buffer.concat([header, scene, body], glbLength);
+    
+    if (callback) {
+        callback(header, scene, body);
```
pjcozzi: @leerichard42 shouldn't we use `process.nextTick()` throughout for callbacks?

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
+'use strict';
```
pjcozzi: Did you check code coverage for this file?  See https://coveralls.io/github/AnalyticalGraphicsInc/gltf-pipeline

lib/getBinaryGltf.js
pjcozzi: Sounds good, can you double check the coverage and update if needed?

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+ *                     "attributes": {
+ *                         "NORMAL": "accessor_0",
+ *                         "POSITION": "accessor_1"
+ *                     }
+ *                 }
+ *             ]
+ *         }
+ *     }
+ * };
+ * var options = {
```
pjcozzi: What is the full documentation for `options`?

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+    var meshId;
+    var primitives;
+    var primitive;
+    var i, j;
+    var index;
+
+    // Retrieve the accessors that should be quantized
+    var quantizeAll = true;
+    if (defined(options)) {
+        attributes = options.attributes;
```
pjcozzi: Also move this into its own function.  I think it will make this function flow really well.

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+            // Quantize the data
+            var data = accessorData.data;
+            var encoded = new Uint16Array(data.length);
+            var index;
+            for (j = 0; j < data.length; j++) {
+                index = j % min.length;
+                encoded[j] = (data[j] - min[index]) * precision / (max[index] - min[index]);
+            }
+
+            //TODO: Repack the buffers once, don't do it for each accessor
```
pjcozzi: What's the plan with this `TODO`?  We generally don't merge code with `TODO`s.

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+
+            // Quantize the data
+            var data = accessorData.data;
+            var encoded = new Uint16Array(data.length);
+            var index;
+            for (j = 0; j < data.length; j++) {
+                index = j % min.length;
+                encoded[j] = (data[j] - min[index]) * precision / (max[index] - min[index]);
+            }
+
```
pjcozzi: Extra whitespace.

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+                "WEB3D_quantized_attributes": {}
+            };
+            attributes = accessor.extensions.WEB3D_quantized_attributes;
+            var accessorData = getAccessorData(gltf, accessorId);
+            var min = accessorData.min;
+            var max = accessorData.max;
+
+            var decodeMatrix = [];
+            var precision = Math.pow(2, 16) - 1;
+            for (j = 0; j < min.length + 1; j++) {
```
pjcozzi: Is there a more straightforward - even if more verbose - way to create `decodeMatrix`?

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+                    else if (j == k) {
+                        decodeMatrix.push((max[j] - min[j]) / precision);
+                    }
+                    else {
+                        decodeMatrix.push(0);
+                    }
+                }
+            }
+            // decodedMin and decodedMax are required for POSITION attributes
+            if (attribute === "POSITION") {
```
pjcozzi: `'`

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+                attributes.decodedMax = max;
+            }
+            attributes.decodeMatrix = decodeMatrix;
+
+
+            // Quantize the data
+            var data = accessorData.data;
+            var encoded = new Uint16Array(data.length);
+            var index;
+            for (j = 0; j < data.length; j++) {
```
pjcozzi: Throughout and in general, assign `data.length` to a local before the loop unless the length can change during the loop.

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+                }
+            }
+        }
+    }
+    // Add WEB3D_quantized_attributes extension to extensionsUsed
+    if (!defined(extensionsUsed)) {
+        extensionsUsed = [];
+        gltf.extensionsUsed = extensionsUsed;
+    }
+    extensionsUsed.push('WEB3D_quantized_attributes');
```
pjcozzi: Perhaps only add this if it doesn't already exist (make a helper function like `addExtensionsUsed(gltf, 'WEB3D_quantized_attributes')` that does all the required checks.

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+    data.max = [];
+
+    for (var i = 0; i < typedArray.length; i++) {
+        var element = typedArray[i];
+        var index = i % numComponents;
+        if (index >= data.min.length) {
+            data.min.push(element);
+            data.max.push(element);
+        }
+        else {
```
pjcozzi: Here and throughout if needed, put `else` on the same line as `}`.

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+            }
+            if (element > data.max[index]) {
+                data.max[index] = element;
+            }
+        }
+    }
+    return data;
+}
+
+function getByteLengthForComponentType(type) {
```
pjcozzi: Move this to its own file incase it is useful elsewhere.

pjcozzi: Actually I could go either way.  Up to you.

lib/quantizeAttributes.js
pjcozzi: In general, does this stage workaround if some accessors are already quantized?

pjcozzi:
Just those comments from me.

This might also need more unit tests.

lasalvavida:
> This might also need more unit tests.

Yes, it definitely needs more robust testing

lib/writeBinaryGltf.js
```diff
@@ -2,15 +2,7 @@
...
-var Cesium = require('cesium');
-var defined = Cesium.defined;
-var defaultValue = Cesium.defaultValue;
-var sizeOf = require('image-size');
-var mime = require('mime');
-var objectValues = require('object-values');
-var writeSource = require('./writeSource');
-var mergeBuffers = require('./mergeBuffers');
-var removePipelineExtras = require('./removePipelineExtras');
+var getBinaryGltf = requilre('./getBinaryGltf');
```
mlimper: typo?

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+        case 5126:
+            return 4;
+    }
+}
+
+function arrayForComponentType(type, buffer, offset, length) {
+    var arrayBuffer;
+    switch (type) {
+        case 5120:
+            return new Int8Array(arrayBuffer);
```
mlimper: why is arrayBuffer undefined and still being used? is this intended?

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,348 @@
...
+            var encoded = new Uint16Array(data.length);
+            var index;
+            for (j = 0; j < data.length; j++) {
+                index = j % min.length;
+                encoded[j] = (data[j] - min[index]) * precision / (max[index] - min[index]);
+            }
+
+            //TODO: Repack the buffers once, don't do it for each accessor
+            //Adjust buffer to be float aligned
+            var shift = (encoded.length * 2) % 4;
```
mlimper: does this mean that the length of the quantized position array, for example, will be padded to the next 4-byte-boundary if necessary (using two additional zero bytes)?

lib/quantizeAttributes.js
lasalvavida: `options` allows you to specify which primitives you want quantized. I'll make the documentation more robust. It's really just for generating test models though, in practice I think you would probably want to quantize everything possible, which is the default behavior if options is undefined.

lib/quantizeAttributes.js
lasalvavida: There's nothing explicitly wrong with doing it the way it is, but it would probably be more efficient to reduce the amount of buffer modification. I'll try to get this fixed now.

lib/quantizeAttributes.js
lasalvavida: Agrred, I'll see if I can find a way to make this more readable

lib/quantizeAttributes.js
lasalvavida: It does not, but it definitely should.

lib/writeBinaryGltf.js
lasalvavida: Whoops, yes that's a typo

lib/quantizeAttributes.js
lasalvavida: Nope, it can be removed, good catch

lib/quantizeAttributes.js
lasalvavida: WebGL throws exceptions if data doesn't start on a multiple of its boundary size. We could look for unused bytes in the buffer and try to repack it to remove those holes, but I think that might be better as a separate stage.

pjcozzi:
@mlimper thanks for the prompt review here!  We really value your support.

@lasalvavida sounds like you are making quick work of this.  Let me know when this is ready for another review.

lib/quantizeAttributes.js
pjcozzi: #50 should do this.

lasalvavida: It should do this, but I'm getting garbled models as in #57. I'll be pushing what I have very shortly, and then I'm going to take a look at fixing removeUnusedVertices.

pjcozzi: Sounds good, thanks for looking at `removeUnusedVertices`.

lasalvavida:
@pjcozzi I think I hit everything in the main function bodies. I'm working on tests and fixing removeUnusedVertices right now, but feel free to take a look at what's here

lib/getAccessorByteStride.js
```diff
@@ -0,0 +1,21 @@
...
+ * If the byteStride is 0, it is calculated based on type and componentType
+ *
+ * @param {Object} [accessor] The accessor.
+ * @returns {Number} The byte stride of the accessor.
+ */
+function getAccessorByteStride(accessor) {
+    if (accessor.byteStride > 0) {
+        return accessor.byteStride;
+    }
+    else {
```
pjcozzi: Drop the `else`.

See "Do not use an unnecessary else block at the end of a function" here: https://github.com/AnalyticalGraphicsInc/cesium/tree/master/Documentation/Contributors/CodingGuide#functions

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,281 @@
...
+
+/**
+ * Quantizes attributes in the provided glTF using the WEB3D_quantized_attributes extension.
+ * If options is undefined, all quantizable attributes will be quantized with the default options.
+ * If options.attributes is undefined, all quantizable attributes will be quantized with the provided options.
+ * If options.precision is undefined, the decodeMatrix be written in its full form.
+ *
+ * @param {Object} [gltf] A javascript object holding a glTF hierarchy.
+ * @param {Object} [options=undefined] Defines more specific quantization behavior.
+ * @param {Object} [options.attributes=undefined] Defines specific attributes to be quantized. This should be as arranged as meshId -> primitiveIndex -> [semantic_0, semantic_1, ... semantic_N]
```
pjcozzi: Do you think it needs to be this precise?  Imagine what a UI for this would look like.

Perhaps just start by providing an array of semantics, e.g., "I want to quantize all POSITIONS."

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,281 @@
...
+ *     }
+ * });
+ */
+function quantizeAttributes(gltf, options) {
+    var accessors = gltf.accessors;
+    var bufferViews = gltf.bufferViews;
+    var buffers = gltf.buffers;
+    var i, j;
+    var accessorIds = {};
+    var precision = undefined;
```
pjcozzi: Just:
```javascript
var precision;
```
`undefined` is implicit and I believe generates a JSHint or WebStorm warning.

lib/quantizeAttributes.js
pjcozzi: This would probably simplify the implementation a lot too.

lasalvavida: I needed it to be that precise for generating my test models. Particularly the mismatch models wouldn't be possible without it. That being said, I do think it would be a good idea to have another option for specifying semantics.

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,281 @@
...
+    if (size === 3) {
+        return createDecodeMatrix3(min, max, range);
+    } else if (size === 4) {
+        return createDecodeMatrix4(min, max, range);
+    } else if (size === 5) {
+        return createDecodeMatrix5(min, max, range);
+    }
+}
+
+function createDecodeMatrix3(min, max, range) {
```
pjcozzi: Here and below, the intention is for the whole numbers to be floating-point, right?  Append `.0`.

See https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Documentation/Contributors/CodingGuide/README.md#basic-code-construction

lib/quantizeAttributes.js
pjcozzi: Can you point me to the specific example of where this much flexibility is needed?

lasalvavida: See [CesiumMilkTruck-mismatch](https://github.com/lasalvavida/cesium/blob/bf81d91d104fa3d0d1d74ab6a5451070c1f3c75f/Specs/Data/Models/WEB3DQuantizedAttributes/CesiumMilkTruck-Mismatch-Quantized.gltf)

Just the POSITIONS and NORMALS on one primitive in a mesh are quantized, and just the TEXCOORD on a primitive in a different mesh is quantized. 

In practice, I agree, you generally just want to quantize everything or a specific semantic, but I needed those options to help generate potentially breaking test cases.

pjcozzi: I don't think it is worth it; virtually no user will want to control quantization in such a fine-grained way and the simpler implementation is better for maintenance.  I think it is a better trade off to keep this simple even if it means some breaking cases can't be tested.

lasalvavida:
@pjcozzi Just tests left to do; everything seems to work so far. One question regarding removeUnusedVertices: packBuffers has effectively the same behavior except that it will re-order and un-interleave accessors to optimize buffer size. Should we remove removeUnusedVertices completely?

lasalvavida:
@pjcozzi This is ready for another look.

Also, I don't know why Travis is failing, the tests all pass locally.

Coverage is:
Statements   : 94.06% ( 1949/2072 )
Branches     : 72.38% ( 663/916 )
Functions    : 97.58% ( 121/124 )
Lines        : 94.06% ( 1949/2072 )

vs. master:
Statements   : 93.81% ( 1714/1827 )
Branches     : 73.09% ( 611/836 )
Functions    : 99.05% ( 104/105 )
Lines        : 93.81% ( 1714/1827 )

pjcozzi:
> packBuffers has effectively the same behavior except that it will re-order and un-interleave accessors to optimize buffer size. Should we remove removeUnusedVertices completely?

I'm not completely following; we still need a stage that looks at indices and removes unused vertices.  I don't believe `packBuffers` does that, right?

pjcozzi:
There are CI test failures: https://travis-ci.org/AnalyticalGraphicsInc/gltf-pipeline/builds/135985901

pjcozzi:
> Also, I don't know why Travis is failing, the tests all pass locally.

Ah, just saw this.  It could be a different OS or Node.js version.

lib/getBinaryGltf.js
pjcozzi: @lasalvavida can you check for `process.nextTick()` elsewhere in gltf-pipeline?  I'm pretty sure we should be using it here for consistent Node.js callback async vs. sync behavior.

lasalvavida:
> I'm not completely following; we still need a stage that looks at indices and removes unused vertices. I don't believe packBuffers does that, right?

Yes, sorry I misunderstood. I'll work packBuffers into removeUnusedVertices. That should also fix #57

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * @returns glTF with quantized attributes
+ *
+ * @example
+ * var gltf = {
+ *     accessors: {
+ *         accessor_0: {...}
+ *         accessor_1: {...}
+ *     },
+ *     meshes: {
```
pjcozzi: Is this example old?

lib/packBuffers.js
```diff
@@ -0,0 +1,132 @@
...
+
+var byteLengthForComponentType = require('./byteLengthForComponentType');
+var getAccessorByteStride = require('./getAccessorByteStride');
+var numberOfComponentsForType = require('./numberOfComponentsForType');
+
+module.exports = packBuffers;
+
+/**
+ * Repacks the accessed buffer data into contiguous chunks.
+ * Also has the effect of un-interleaving interleaved accessors.
```
pjcozzi: Is it reasonable/useful to un-interleave as a separate stage since several other stages will also need to depend on it?

lib/quantizeAttributes.js
lasalvavida: It looks okay to me. Do you see something wrong with it?

lib/packBuffers.js
```diff
@@ -0,0 +1,132 @@
...
+
+function packGltfBuffer(gltf, bufferId, packBufferViews) {
+    var buffers = gltf.buffers;
+    var buffer = buffers[bufferId];
+    var source = buffer.extras._pipeline.source;
+    var packBuffer = new Uint8Array(source.length);
+    var accessors = getAccessorsByTargetAndByteLength(gltf, bufferId);
+    var offset = 0;
+
+    var targets = [34962, 0, 34963];
```
pjcozzi: Use `Cesium.WebGLConstants`.  See https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Renderer/WebGLConstants.js

lib/packBuffers.js
```diff
@@ -0,0 +1,132 @@
...
+function getAccessorsByTargetAndByteLength(gltf, bufferId) {
+    var accessors = gltf.accessors;
+    var bufferViews = gltf.bufferViews;
+    var accessorsByTargetAndByteLength = {};
+    for (var accessorId in accessors) {
+        var accessor = accessors[accessorId];
+        var bufferView = bufferViews[accessor.bufferView];
+        if (bufferView.buffer === bufferId) {
+            var target = bufferView.target;
+            if (!defined(target)) {
```
pjcozzi: Use `defaultValue`; it will be more concise.

lib/quantizeAttributes.js
```diff
@@ -0,0 +1,245 @@
...
+        attributeSemantic.indexOf("COLOR") >= 0;
+}
+
+function accessorIsQuantizable(accessor) {
+    // This accessor is already quantized
+    if (accessorIsQuantized(accessor)) {
+        return false;
+    }
+    // Only 32-bit float to 16-bit int quantization is supported
+    return accessor.componentType == 5126;
```
pjcozzi: Same `WebGLConstants` comment here and throughout if needed.

lib/packBuffers.js
lasalvavida: I don't think so; the un-interleaving is just a by-product of packBuffers copying the data into the smallest space possible. 

I could write a separate function to do un-interleaving without removing unused data, but it wouldn't make this one simpler and I can't think of a case where you would want to intentionally keep unused data.

lib/quantizeAttributes.js
pjcozzi: Did you see https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/56#discussion_r66105451

Perhaps the right way to keep the interface simple and pragmatic and also allow testing the corner cases is to allow the test to pass in an undocumented callback function that takes the right parameters to decide if an attribute should be quantized or not.  That keeps this implementation simple and moves the corner-case complexity (and overall less complexity) to the tests where it is the focus.

lib/packBuffers.js
pjcozzi: OK, perhaps name this `uninterleaveAndPackBuffers` (generally, putting "and" in a name is not ideal, but I think it is valuable here) since other stages will want to depend on it for the uninterleaving part.

lib/quantizeAttributes.js
lasalvavida: I followed the suggestions in that comment and changed the example; that was part of this round of changes. Instead of having the user specify specifically which attributes in which meshes, it is now just options.semantics: [SEMANTIC1, SEMANTIC2 ...].

A callback is an interesting idea, but I'm not sure it's necessary. I needed to generate corner cases for the reader, in the writer it doesn't really matter, so it's probably better just to keep it simple.

lasalvavida: I'll remove the example gltf hierarchy at the top though. It isn't really necessary anymore since the examples aren't targeting specific id's anymore

pjcozzi: Ah, right!  This is the glTF, not `options`! 😆

pjcozzi: OK

lasalvavida:
@pjcozzi I think I hit everything here. Fixing removeUnusedVertices will be part of a separate pull request; it's outside the scope of these changes.

bin/gltf-pipeline.js
```diff
@@ -18,7 +18,8 @@ if (process.argv.length < 3 || defined(argv.h) || defined(argv.help)) {
         '  -i, input=PATH Read unoptimized glTF from the specified file.\n ' +
         '  -o, output=PATH write optimized glTF to the specified file.\n' +
         '  -b, write binary glTF file.\n' +
-        '  -s, writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.\n  ';
+        '  -s, writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.\n' + 
+        '  -q, quantize the attributes of this model.\n';
```
pjcozzi: Also update README.md please.

lib/quantizeAttributes.js
pjcozzi: How hard is it to get test coverage for this file to 100%? https://coveralls.io/builds/6535609/source?filename=lib%2FquantizeAttributes.js

pjcozzi:
Just those trivial comments.  Looking forward to merging this!

lasalvavida:
Updated.

pjcozzi:
Awesome!

PR #59, opened 2016-06-07
Added handling for attribute type and index type in `removeUnusedVertices`

### LINKED ISSUE #58: `removeUnusedVertices` relies on byte strides exclusively
I found out that if you run the pipeline on a model that has accessors with `byteStride` set to `0` but `type`  set to `VEC3`, `compressBuffers` in `removeUnusedVertices` strips out too much information and causes an `indexOutOfBounds` error soon after. 

From my reading of the gltf spec ([this part](https://github.com/KhronosGroup/glTF/blob/master/specification/README.md#accessor)) it seems like the pipeline should be able to observe from `"byteOffset": 0` and `"type": "VEC3"` that the data is tightly packed vec3s.

likangning93:
Fixes #58.

Added handling so when `removeUnusedVertices` sees that an attribute's type is `VEC3` and that the accessor's byte stride is 0 it assumes the attributes are tightly packed. Also added additional handling for `unsigned short` indices. I used switch statements here to make room for float indices if we ever want to support that, but I think the current stable version of Node doesn't allow reading floats from buffers yet. I'll make a note in the code.

Also added specs for tightly packed attributes and short indices.


lib/removeUnusedVertices.js
```diff
@@ -132,7 +146,15 @@ function calculateIndexIntervals(accessors, buffers, bufferViews, meshes) {
...
-                            var index = WebGLConstants.UNSIGNED_SHORT ? accessorSource.readUInt16LE(accessorIndex) : accessorSource.readUInt8LE(accessorIndex);
+                            var index = 0;
+                            switch(indexStride) {
+                                case 2:
+                                    index = accessorSource.readUInt16LE(accessorIndex);
+                                    break;
+                                case 1:
+                                    index = accessorSource[accessorIndex];
+                                    break;
+                            }
```
likangning93: I think according to the gltf spec it's also possible that the index type can be `FLOAT`, but it seems like Node 4.4.5 LTS doesn't support `Buffer.readFloatLE`.

lilleyse: Yeah don't worry about floats, if anything we may want to handle uint32, even though it's an extension.

lib/removeUnusedVertices.js
```diff
@@ -122,7 +122,21 @@ function calculateIndexIntervals(accessors, buffers, bufferViews, meshes) {
                         var indexView = bufferViews[indexAccessor.bufferView];
                         var indexBuffer = buffers[indexView.buffer];
                         var indexSource = indexBuffer.extras._pipeline.source.slice(indexView.byteOffset, indexView.byteOffset + defaultValue(indexView.byteLength, 0));
-                        var indexStride = indexAccessor.componentType === WebGLConstants.UNSIGNED_SHORT ? 2 : 1;
+                        var indexStride = 2; // default to SHORT
+                        switch(indexAccessor.componentType) {
```
lasalvavida: As discussed offline see [byteLengthForComponentType](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/web3d-quantized-attributes/lib/byteLengthForComponentType.js)

lib/removeUnusedVertices.js
```diff
@@ -179,10 +200,56 @@ function calculateIndexIntervals(accessors, buffers, bufferViews, meshes) {
...
+                                                componentWidth = 2;
+                                                break;
+                                            case WebGLConstants.UNSIGNED_SHORT:
+                                                componentWidth = 2;
+                                                break;
+                                            case WebGLConstants.FLOAT:
+                                                componentWidth = 4;
+                                                break;
+                                        }
+                                        switch (attributeAccessor.type) {
```
lasalvavida: See [numberOfComponentsForType](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/web3d-quantized-attributes/lib/byteLengthForComponentType.js)

lasalvavida:
You absolutely don't have to since they aren't in master yet, but if you want to pull those utility files I pointed out from the web3d quantized attributes [branch](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/tree/web3d-quantized-attributes) it should help your readability a bit here.

likangning93:
Thanks Rob! Updated!

lib/removeUnusedVertices.js
```diff
@@ -179,10 +188,15 @@ function calculateIndexIntervals(accessors, buffers, bufferViews, meshes) {
...
                                         attributeStride = attributeAccessor.byteStride;
                                     }
+                                    else if (defined(attributeAccessor.type)){
+                                        // If an attribute type is defined, a byteStride of 0 is equivalent to a byteStride
+                                        // of BYTE_COUNT(type). The attributeStride is the width of the attribute type.
+                                        attributeStride = getAccessorByteStride(attributeAccessor);
+                                    }
                                     else {
                                         //Int16
                                         if (attributeAccessor.componentType === WebGLConstants.SHORT || attributeAccessor.componentType === WebGLConstants.UNSIGNED_SHORT) {
```
lilleyse: Could this whole section:

```javascript
var attributeStride = 1; // Default to width of BYTE and UNSIGNED BYTE
if (attributeAccessor.byteStride > 0) {
    attributeStride = attributeAccessor.byteStride;
}
else if (defined(attributeAccessor.type)){
    // If an attribute type is defined, a byteStride of 0 is equivalent to a byteStride
    // of BYTE_COUNT(type). The attributeStride is the width of the attribute type.
    attributeStride = getAccessorByteStride(attributeAccessor);
}
else {
    //Int16
    if (attributeAccessor.componentType === WebGLConstants.SHORT || attributeAccessor.componentType === WebGLConstants.UNSIGNED_SHORT) {
        attributeStride = 2;
    }
    //Float32
    if (attributeAccessor.componentType === WebGLConstants.FLOAT) {
        attributeStride = 4;
    }
}
```

be replaced by

```javascript
var attributeStride = getAccessorByteStride(attributeAccessor);
```

The accessor's `type` is required right? If this isn't a safe assumption, then we could modify `getAccesorByteStride` or `numberOfComponentsForType` to account for an undefined type instead of checking here.

lib/removeUnusedVertices.js
```diff
@@ -353,14 +366,24 @@ function updateIndices(bufferViews, buffers) {
...
-                    var index = (indexStride === 2) ? indexSource.readUInt16LE(sourceIndex) : indexSource.readUInt8LE(sourceIndex);
+                    var index = 0;
+                    switch(indexStride) {
+                        case 2:
+                            index = indexSource.readUInt16LE(sourceIndex);
+                            break;
+                        case 1:
+                            index = indexSource[sourceIndex];
+                            break;
+                    }
```
lilleyse: We should probably have a helper function somewhere for reading/writing indexes like this, especially so it's easy to add uint32 support later. I noticed similar logic is in `combinePrimitives` as well. @lasalvavida have you done anything like this in your branch?

specs/lib/removeUnusedVerticesSpec.js
```diff
@@ -419,6 +419,200 @@ describe('removeUnusedVertices', function() {
...
+                    "extras": {
+                        "_pipeline": {}
+                    }
+                }
+            },
+            "bufferViews": {
+                "position_view": {
+                    "buffer": "buffer_0",
+                    "byteOffset": 6 * 2,
+                    "byteLength": 4 * 3 * 2,
```
lilleyse: At first isn't this one `7 * 3 * 2`?

specs/lib/removeUnusedVerticesSpec.js
```diff
@@ -419,6 +419,200 @@ describe('removeUnusedVertices', function() {
...
+                    "target": 34963,
+                    "extras": {
+                        "_pipeline": {}
+                    }
+                }
+            },
+            "buffers": {
+                "buffer_0": {
+                    "uri": "data:",
+                    "byteLength": indices.concat(positions).length * 2,
```
lilleyse: `(indices.length + position.length) * 2` is a bit clearer

specs/lib/removeUnusedVerticesSpec.js
likangning93: Right. Sorry, that was careless of me. I also realized I got the byteLength for a lot of things wrong... I forgot that the positions use 32 bit floats instead of 16.

likangning93:
Updated!

lib/removeUnusedVertices.js
likangning93: Should we shoot to do that in this pull request?

specs/lib/removeUnusedVerticesSpec.js
```diff
@@ -435,9 +435,9 @@ describe('removeUnusedVertices', function() {
             2,2,2,
             2,2,2
         ];
-        var positionsBuffer = new Buffer(positions.length * 2);
+        var positionsBuffer = new Buffer(positions.length * 4);
         for (var i = 0; i < positions.length; i++) {
-            positionsBuffer.writeUInt16LE(positions[i], i * 2);
+            positionsBuffer.writeUInt16LE(positions[i], i * 4);
```
lilleyse: `writeUInt16LE` should change to `writeFloatLE` to be consistent with the new change.

lib/removeUnusedVertices.js
lilleyse: Eh, let's save it for another. It's not really critical at the moment.

likangning93:
Updated!

lilleyse:
Looks good!

lib/removeUnusedVertices.js
likangning93: I stand corrected on what I said about Node 4.4.5 and `Buffer.readFloatLE`, it's in the 4.4.5 documentation. My mistake.

PR #61, opened 2016-06-08
Added npmignore

JoshuaStorm:
I forgot to make a PR for adding the npmignore during the code sprint. Here it is! Let me know if I am missing anything.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,9 @@
+/.npmignore
+/.travis.yml
+/build
+/coverage
+/doc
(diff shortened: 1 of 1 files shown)


lilleyse:
Also include `.jshintrc` and `.idea` to this list. Also you don't need to put slashes before every item.

JoshuaStorm:
Updated. I used the slashes just to mirror the style of the Cesium ignore file.

lilleyse:
Ah, alright. Looks good.

PR #62, opened 2016-06-08
Debugging support for tests

lilleyse:
Added a gulp task for running / debugging unit tests. In the gulp pane in Webstorm right-click `test` and select `Debug 'test'`. This should respond to any breakpoints set.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,6 +2,8 @@
 
 var jshint = require('gulp-jshint');
 var gulp = require('gulp');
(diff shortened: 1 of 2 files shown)


lilleyse:
@lasalvavida Could you quickly take a look at this?

lasalvavida:
Tried it and it works, should be good.

pjcozzi:
Please update README.md with any instructions.  As build/debug/test/etc instructions grow, we'll move them from README.md to their own guide like Cesium.

PR #64, opened 2016-06-08
Make removeUnusedVertices use uninterleaveAndPackBuffers

### LINKED ISSUE #57: The removeUnusedVertices stage of the pipeline mangled a model
The model in question is the chinese dragon model from [here](https://github.com/SaschaWillems/Vulkan/blob/master/models/chinesedragon.obj).

I ran it through obj2gltf to produce this [valid gltf](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/298298/chinesedragon.zip).

dragon-normal

Remove unusedVertices outputs [this](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/298302/chinesedragon-mangled.zip).

chinese-dragon-mangled

lasalvavida:
Resolves #57.

I think this is also more readable.
Maybe consider renaming to removeUnusedAttributeValues or something similar since I think that may be a more accurate description.


lasalvavida:
Found the cause of the CI failures: Node needs to be a higher version to use Buffer.read{...}LE() functions. This should fix the web3d branch as well.

lasalvavida:
I still need to make sure this can handle the case where different primitives use the same accessors with different indices.

lasalvavida:
@lilleyse Could you look at this when you get a chance? Just integrated the tests that @likangning93 wrote from master and they pass with this version of removeUnusedVertices.

@leerichard42 I'd also be very interested in your feedback here.

lilleyse:
Sure I'll take a look soon.

lasalvavida:
As far as the coverage situation goes the new removeUnusedVertices does have 100% coverage. The drop in percentage looks like it is from not all data types in numberOfComponentsForType being used and from removeUnusedVertices not having as many tests, so the hits/line is lower.

If this is something we think needs addressing, I'm open to suggestions.

lilleyse:
Nothing much to say here, it looks great and the tests are very comprehensive.

As for the test coverage and naming, I think they're both fine.

leerichard42:
Agreed, looks awesome! Thanks for fixing this up!

PR #67, opened 2016-06-09
Added separateImage flag 

JoshuaStorm:
Added separateImage flag to embed glTF geometry and shaders, but leaving textures separate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Include `build/gltf-pipeline.js` it with a `script` tag.  For a simple example, 
 |`-b`, `--binary`|Write binary glTF file.|No, default `false`|
 |`-s`, `--separate`|Writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.|No, default `false`|
 |`-h`, `--help`|Display help|No|
-|`-t`, &nbsp;&nbsp;&nbsp;&nbsp;`--separateTexture`|Write out separate textures, but embed geometry/animation data files, and shader files.|No, default `false`|
+|`-t`, &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`--separateTexture`|Write out separate textures, but embed geometry/animation data files, and shader files.|No, default `false`|
 
 
 ## Build Instructions
--- a/README.md
+++ b/README.md
@@ -42,7 +42,8 @@ Include `build/gltf-pipeline.js` it with a `script` tag.  For a simple example, 
 |`-b`, `--binary`|Write binary glTF file.|No, default `false`|
 |`-s`, `--separate`|Writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.|No, default `false`|
 |`-h`, `--help`|Display help|No|
-|`-t`, `--separateTexture`|Write out separate textures, but embed geometry/animation data files, and shader files.|No, default `false`|
+|`-t`,
+`--separateTexture`|Write out separate textures, but embed geometry/animation data files, and shader files.|No, default `false`|
 
 
 ## Build Instructions
--- a/README.md
+++ b/README.md
@@ -37,12 +37,12 @@ Include `build/gltf-pipeline.js` it with a `script` tag.  For a simple example, 
 
 |Flag|Description|Required|
 |----|-----------|--------|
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -42,6 +42,8 @@ Include `build/gltf-pipeline.js` it with a `script` tag.  For a simple example,
 |`-b`|Write binary glTF file.|No, default `false`|
 |`-s`|Writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.|No, default `false`|
 |`-h`|Display help|No|
+|`--separateImage`|Write out separate textures, but embed geometry/animation data files, and shader files.|No, default `false`|
```
lasalvavida: For consistency, maybe we should have a single letter flag and a '--name' style flag for each option. @lilleyse, what do you think?

JoshuaStorm: Probably a good call. @lilleyse and I briefly discussed this: do you have any suggestions for a single letter flag for `separateImage`?

lasalvavida: How about `-t` `--separateTextures`?

lilleyse: Yeah we have both for the asset pipeline, seems like a good idea here too.

README.md
```diff
@@ -37,11 +37,13 @@ Include `build/gltf-pipeline.js` it with a `script` tag.  For a simple example,
...
-|`-o`|Directory or filename for the exported glTF file.|No|
-|`-b`|Write binary glTF file.|No, default `false`|
-|`-s`|Writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.|No, default `false`|
-|`-h`|Display help|No|
+|`-i, --input`|Path to the input glTF file.| :white_check_mark: Yes|
+|`-o, --output`|Directory or filename for the exported glTF file.|No|
+|`-b, --binary`|Write binary glTF file.|No, default `false`|
+|`-s, --separate`|Writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.|No, default `false`|
+|`-h, --help`|Display help|No|
+|`-h, --separateImage`|Write out separate textures, but embed geometry/animation data files, and shader files.|No, default `false`|
```
lasalvavida: This should be `-t` I think

lasalvavida: And the `--separateImage` should be `--separateTexture`

JoshuaStorm: Copy paste mistakes strike again

lasalvavida:
Other than that, this looks good to me

bin/gltf-pipeline.js
```diff
@@ -47,7 +49,8 @@ if (!defined(outputPath)) {
 
 var options = {
     isBinary : isBinary,
-    isEmbedded : !isSeparate
+    isEmbedded : !isSeparate,
+    isImageEmbedded : !isTextureSeparate
```
lilleyse: The naming is a bit inconsistent with `texture` and `image`. In the context of gltf, I think we should always stick with `image` (though the command line input could stay as `-t` I guess only because `-i` is taken. Also I started to make this change on my branch, but I renamed these options to just `binary`, `embed`, etc. Seems more directive... So this could be `embedImages`.

lilleyse:
Could you also add a spec for this? Write out the textured box model and see how many files `fs` writes, using the spy method from before.

lib/writeSource.js
```diff
@@ -11,7 +11,7 @@ module.exports = writeSource;
 /**
  * @private
  */
-function writeSource(gltf, basePath, name, isEmbedded, callback) {
+function writeSource(gltf, basePath, name, isEmbedded, isImageEmbedded, callback) {
```
lilleyse: Can you also change the names here to `embed` and `embedImage`.

Also throughout can you rename to `embedImages`?

bin/gltf-pipeline.js
```diff
@@ -15,22 +15,24 @@ var readGltf = require('../lib/readGltf');
...
-var gltfPath = defaultValue(argv._[0], argv.i);
+var gltfPath = defaultValue(argv._[0], argv.i || argv.input);
 var fileExtension = path.extname(gltfPath);
 var fileName = path.basename(gltfPath, fileExtension);
 var filePath = path.dirname(gltfPath);
 
-var outputPath = defaultValue(argv._[1], argv.o);
-var isSeparate = defaultValue(argv.s, false);
-var isBinary = defaultValue(argv.b, false);
+var outputPath = defaultValue(argv._[1], argv.o || argv.output);
```
lilleyse: To follow the coding guide, just do `defaultValue(argv._[1], defaultValue(argv.o, argv.output))` for these instead of `||`.

lilleyse:
Merge in master to fix the conflict.

PR #65, opened 2016-06-09
Required min max

lasalvavida:
Fills in accessor min and max if they aren't present as those will be required properties in glTF 1.0.1

PR #69, opened 2016-06-09
added some helper methods for working with nodes and node hierarchies

likangning93:
Mostly in preparation for Ambient Occlusion, but this should also be generally helpful. It might also help condense  https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/51, @leerichard42 can you take a look too?


likangning93:
Some updates: I also added a function that I'm going to be using to generate triangle soup for raytracing ambient occlusion. I only just noticed that this does something pretty similar to  https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/65, just more general purpose.

lib/readAccessor.js
```diff
@@ -0,0 +1,112 @@
...
+    var componentCount = numberOfComponentsForType(accessor.type);
+    var componentBytes = byteLengthForComponentType(accessor.componentType);
+    var attributeStride = componentCount * componentBytes;
+    var accessorType = accessor.type;
+    var componentType = accessor.componentType;
+    var attributes = [];
+    var currentBufferIndex = accessor.byteOffset + bufferView.byteOffset;
+
+    var readFunction;
+    switch (componentType) {
```
lasalvavida: See [readBufferComponentType](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/readBufferComponentType.js). Should simplify this.

lib/nodeHelpers.js
```diff
@@ -0,0 +1,101 @@
...
+    for (var index in rootNodeIDs) {
+        var rootNode = nodes[rootNodeIDs[index]];
+        depthFirstTraversal(rootNode, nodes, addNodeFunction);
+    }
+    return allSceneNodes;
+}
+
+// Perform a stack based iterative depth first traversal on the specified node.
+// Apply functionNodeMatrix to all of its children.
+// functionNodeMatrix functions take this node and the parent as arguments
```
lilleyse: Update comment with the new name.

lib/nodeHelpers.js
```diff
@@ -0,0 +1,101 @@
...
+    var nodeStack = [rootNode];
+    while (nodeStack.length > 0) {
+        var currentNode = nodeStack.pop();
+        // apply the function to all children and push them onto the stack
+        if (!defined(currentNode.children)) {
+            continue;
+        }
+        var childrenIDs = currentNode.children;
+        for (var index in childrenIDs) {
+            var childID = childrenIDs[index];
```
lilleyse: Check `hasOwnProperty` here. Heads up that Webstorm highlights the variable whenever it is missing. There are a few other places in this file.

specs/lib/readAccessorSpec.js
```diff
@@ -0,0 +1,137 @@
...
+    function fitsBounds(min, max, attributes) {
+        // check if the data in values is bounded by min and max precisely
+        var minInValues;
+        var maxInValues;
+        var attributeToArray;
+        var numberValues = 0;
+
+        switch(attributes.type) {
+            case 'number':
+                attributeToArray = function(value) {
```
lasalvavida: I think you can make this less verbose. See the Cesium [pack](https://cesiumjs.org/Cesium/Build/Documentation/Cartesian2.html?classFilter=cartes#.pack) functions.

lib/readAccessor.js
```diff
@@ -0,0 +1,112 @@
...
+        for (var j = 0; j < componentCount; j++) {
+            values.push(readFunction(internalBufferIndex));
+            internalBufferIndex += componentBytes;
+        }
+        attributes.push(generateAttributeFunction(values));
+        currentBufferIndex += attributeStride;
+    }
+    return {
+        "type" : type,
+        "data" : attributes
```
lilleyse: Except when creating test json examples, we usually don't put quotes around the key names.

specs/lib/readAccessorSpec.js
lasalvavida: Actually, I think just try to use [findAccessorMinMax](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/findAccessorMinMax.js) here to avoid the duplicate code. Then do your comparison in the test

lasalvavida:
Since `nodeHelpers` exports an object and not a single function, it should be `NodeHelpers`.

lib/nodeHelpers.js
likangning93: I wasn't sure that also applied to arrays. I guess are javascript arrays just dictionaries with indices instead of names?

lasalvavida: If it is an array, don't traverse it using `in`. Store the length and then iterate from 0 to the length of the array. We typically only use `in` to iterate over Objects.

specs/lib/readAccessorSpec.js
likangning93: I wanted to test `readAccessor` output for both containment and min/max fit. Renamed to reflect that. I'm not sure how to do the same thing using `findAccessorMinMax` with significantly less code.

likangning93:
Okay. Updated! Although git doesn't seem to care if I rename nodeHelpers.js?

lasalvavida:
Git bash isn't case sensitive, so you may have to rename it to something else then to the desired end name with the case changed.

likangning93:
Thanks Rob! Updated.

lib/readAccessor.js
```diff
@@ -0,0 +1,83 @@
...
+        for (var j = 0; j < componentCount; j++) {
+            values.push(readBufferComponentType(source, componentType, internalBufferIndex));
+            internalBufferIndex += componentBytes;
+        }
+        attributes.push(generateAttributeFunction(values));
+        currentBufferIndex += attributeStride;
+    }
+    return {
+        'type' : type,
+        'data' : attributes
```
lilleyse: Sorry, I should have been clearer, but you don't need any quotes.

specs/lib/readAccessorSpec.js
```diff
@@ -0,0 +1,125 @@
...
+
+        var allAccessors = testBoxGltf.accessors;
+        for (var accessorID in allAccessors) {
+            if (allAccessors.hasOwnProperty(accessorID)) {
+                var accessor = allAccessors[accessorID];
+                accessorIDtoData[accessorID] = readAccessor(accessor, testBoxGltf);
+                if (defined(accessor.min) && defined(accessor.max)) {
+                    accessorIDtoMinMax[accessorID] = {
+                        "min" : accessor.min,
+                        "max" : accessor.max
```
lilleyse: Change these too.

likangning93:
Got it. Updated!

specs/lib/readAccessorSpec.js
```diff
@@ -0,0 +1,125 @@
...
+            else {
+                boxGltf = JSON.parse(data);
+                addPipelineExtras(boxGltf);
+                loadGltfUris(boxGltf, path.dirname(gltfPath), function(err, gltf) {
+                    if (err) {
+                        throw err;
+                    }
+                    done();
+                });
+            }
```
lilleyse: Sorry, one last comment. Could you use `readGltf` here instead?

likangning93:
Updated.

lilleyse:
Nice the new functions will be helpful.

lib/NodeHelpers.js
```diff
@@ -0,0 +1,106 @@
...
+    getAllNodesInScene: getAllNodesInScene,
+    depthFirstTraversal: depthFirstTraversal
+};
+
+// computes flattened Matrix4 transforms for every node in the scene
+// store these matrices for each node in extras._pipeline.flatTransform
+function computeFlatTransformScene(scene, nodes) {
+    var rootNodeIDs = scene.nodes;
+    var scratchMat4 = new Matrix4();
+    var flattenTransformFunction = function(node, parent) {
```
pjcozzi: Use `function flattenTransformFunction` style syntax.

Also declare this outside of the `computeFlatTransformScene` function and pass in any needed parameters.

It should be better for the optimizer.

lib/NodeHelpers.js
```diff
@@ -0,0 +1,106 @@
...
+        var localTransform = getLocalMatrix4(node, scratchMat4);
+        var parentTransform = Matrix4.IDENTITY;
+        if (defined(parent)) {
+            parentTransform = parent.extras._pipeline.flatTransform;
+        }
+        var flatTransform = new Matrix4();
+        Matrix4.multiply(parentTransform, localTransform, flatTransform);
+        node.extras._pipeline.flatTransform = flatTransform;
+    };
+    for (var index in rootNodeIDs) {
```
pjcozzi: Throughout, instead of calling this `index` call it `id` so we are consistent with glTF terminology.

lib/NodeHelpers.js
```diff
@@ -0,0 +1,106 @@
...
+function computeFlatTransformScene(scene, nodes) {
+    var rootNodeIDs = scene.nodes;
+    var scratchMat4 = new Matrix4();
+    var flattenTransformFunction = function(node, parent) {
+        var localTransform = getLocalMatrix4(node, scratchMat4);
+        var parentTransform = Matrix4.IDENTITY;
+        if (defined(parent)) {
+            parentTransform = parent.extras._pipeline.flatTransform;
+        }
+        var flatTransform = new Matrix4();
```
pjcozzi: Perhaps just
```javascript
node.extras._pipeline.flatTransform = Matrix4.multiply(parentTransform, localTransform, new Matrix4());
```

lib/NodeHelpers.js
pjcozzi: > Use function flattenTransformFunction style syntax.

Same comment throughout.

> Also declare this outside of the computeFlatTransformScene function and pass in any needed parameters.

Same comment throughout, but it looks like you'll have to pass the parameter through `depthFirstTraversal` so I'm not 100% sure it is better, but I suspect it is.

likangning93: What about for `addNodeFunction`? Is there an accepted way to give this access to `allSceneNodes` or another node list?

lib/NodeHelpers.js
```diff
@@ -0,0 +1,106 @@
...
+function getLocalMatrix4(node, result) {
+    if (!defined(result)) {
+        result = new Matrix4();
+    }
+    // if there is a matrix already, return the matrix
+    if (defined(node.matrix)) {
+        return Matrix4.fromColumnMajorArray(node.matrix, result);
+    }
+    // get it from TRS
+    var translation = Cartesian3.fromArray(defaultValue(node.translation, [0, 0, 0]));
```
pjcozzi: Does this stage depend on `addDefaults`?  Are we sure that we need to add them here?

If so, they should be file-level scratch variables so we don't allocate on each call.

lib/NodeHelpers.js
```diff
@@ -0,0 +1,106 @@
...
+
+// Perform a stack based iterative depth first traversal on the specified node.
+// Apply functionChildParent to all of its children.
+// functionChildParent functions take this node and the parent as arguments
+function depthFirstTraversal(rootNode, allNodes, functionChildParent) {
+    // process the root node
+    functionChildParent(rootNode);
+
+    // start the traversal with the root node in the queue.
+    var nodeStack = [rootNode];
```
pjcozzi: Here and throughout this file, there should be file-scoped scratch variables.

See https://github.com/AnalyticalGraphicsInc/cesium/tree/master/Documentation/Contributors/CodingGuide#result-parameters-and-scratch-variables

lib/readAccessor.js
```diff
@@ -0,0 +1,83 @@
...
+var byteLengthForComponentType = require('./byteLengthForComponentType');
+var numberOfComponentsForType = require('./numberOfComponentsForType');
+var readBufferComponentType = require('./readBufferComponentType');
+
+module.exports = readAccessor;
+
+// read the accessor shown and return an array containing its contents
+// VECx types become CartesianX
+// MATx types become MatrixX
+function readAccessor(accessor, gltf) {
```
pjcozzi: Pass in `gltf` first.  In general, that should come first in function parameters, e.g., less specific to more specific, e.g., `(gltf, mesh, primitive)`.

specs/lib/readAccessorSpec.js
```diff
@@ -0,0 +1,115 @@
...
+var path = require('path');
+var clone = require('clone');
+var loadGltfUris = require('../../lib/loadGltfUris');
+var addPipelineExtras = require('../../lib/addPipelineExtras');
+var readAccessor = require('../../lib/readAccessor');
+var readGltf = require('../../lib/readGltf');
+
+var gltfPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf';
+
+describe('bakeAmbientOcclusion', function() {
```
pjcozzi: `bakeAmbientOcclusion` -> is this the best name?

lib/NodeHelpers.js
likangning93: Ah. File level scratch. Okay.

pjcozzi:
Just those mostly style comments that can be addressed in a new PR.

specs/lib/readAccessorSpec.js
lilleyse: Ah.. saw that but forgot to comment.

PR #70, opened 2016-06-09
Fixes for the json interface and name changes

lilleyse:
I made some fixes to `gltfPipeline` since we were missing some required steps.


lib/modelMaterialsCommon.js
```diff
@@ -558,16 +558,18 @@ function generateTechnique(gltf, khrMaterialsCommon, lightParameters) {
     // Add shaders
     shaders[vertexShaderId] = {
         type: WebGLConstants.VERTEX_SHADER,
-        uri: 'data:,' + vertexShader,
         extras: {
-            source: vertexShader
+            _pipeline : {
+                source : vertexShader
+            }
```
lilleyse: I made this change to fix a crash along the way since every `source` expects there to be `extras._pipeline.source`. I will have to handle Cesium compatibility for this https://github.com/AnalyticalGraphicsInc/OBJ2GLTF/issues/4#issuecomment-224075451

pjcozzi: Why not store the source in the uri base64 encoded?

lilleyse: This step happens after `loadGltfUris`, and after that point the pipeline doesn't really look in the uri.

lasalvavida:
@lilleyse Can you merge master? It also looks like there is a test throwing the missing basePath exception you added.

lilleyse:
Merged master and fixed the test, as well fixed some formatting things.

lasalvavida:
@lilleyse these changes look good to me, is it ready to merge?

lilleyse:
Yup it's ready.

Issue #12, opened 2016-01-16
Browserify doesn't run on Windows


leerichard42:
```
Richard@Richard-PC /cygdrive/c/Users/Richard/src/gltf-pipeline
$ npm run build

> gltf-pipeline@0.0.0 build C:\Users\Richard\src\gltf-pipeline
> browserify index.js --standalone gltfPipeline -o build/gltf-pipeline.js

Error: Cannot find module './process' from 'C:\Users\Richard\src\gltf-pipeline\node_modules\cesium\node_modules\requirejs\bin'
    at C:\Users\Richard\AppData\Roaming\npm\node_modules\browserify\node_modules\resolve\lib\async.js:55:21
    at load (C:\Users\Richard\AppData\Roaming\npm\node_modules\browserify\node_modules\resolve\lib\async.js:69:43)
    at onex (C:\Users\Richard\AppData\Roaming\npm\node_modules\browserify\node_modules\resolve\lib\async.js:92:31)
    at C:\Users\Richard\AppData\Roaming\npm\node_modules\browserify\node_modules\resolve\lib\async.js:22:47
    at FSReqWrap.oncomplete (fs.js:82:15)

npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run" "build"
npm ERR! node v4.2.4
npm ERR! npm  v2.14.12
npm ERR! code ELIFECYCLE
npm ERR! gltf-pipeline@0.0.0 build: `browserify index.js --standalone gltfPipeline -o build/gltf-pipeline.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the gltf-pipeline@0.0.0 build script 'browserify index.js --standalone gltfPipeline -o build/gltf-pipeline.js'.
npm ERR! This is most likely a problem with the gltf-pipeline package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     browserify index.js --standalone gltfPipeline -o build/gltf-pipeline.js
npm ERR! You can get their info via:
npm ERR!     npm owner ls gltf-pipeline
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     C:\Users\Richard\src\gltf-pipeline\npm-debug.log
```

[...]

PR #71, opened 2016-06-10
Tweak README.md

pjcozzi:
Made it less redundant.


pjcozzi:
I also downplayed the usage as a client-side JavaScript library since it is incomplete and we do not have the time to focus on it; it would be a welcomed external contribution.

README.md
```diff
@@ -68,7 +61,13 @@ For complete coverage details, open `coverage/lcov-report/index.html`.
 
 The tests and coverage covers the Node.js module; it does not cover the command-line interface (which is tiny) or the built JavaScript library (which is the same code just ran through browserify).
 
-### Building the JavaScript library
+## Experimental JavaScript library
+
+### Usage
+
+Include `build/gltf-pipeline.js` it with a `script` tag.  For a simple example, see [build/index.html](build/index.html).
```
lilleyse: `build/gltf-pipeline.js` is empty now, but even before was pretty out of date. And `index.js` is a lot smaller than it used to be, so `index.html` won't work. Do we still want to include these instructions?

pjcozzi:
Removed instructions and browserify build.  Implicitly fixes #12.

PR #73, opened 2016-06-10
ModelMaterialsCommon shader source is now in extras._pipeline

likangning93:
`modelMaterialsCommon` used to just put the shader source just in `extras.source`, which was preventing the rest of the pipeline from writing gltf that has new shaders (for example, models from the assimp converter) since the rest of the pipeline all works out of `extras._pipeline`.

Also updated a KHR_materials_common spec to check for this.


specs/lib/addDefaultsSpec.js
```diff
@@ -244,6 +245,14 @@ describe('addDefaults', function() {
                 expect(gltfClone.materials[materialID].values).toEqual(expectValues);
             }
         }
+
+        // check that shader source is in extras._pipeline.source
+        var shaders = gltfClone.shaders;
+        expect(Object.keys(shaders).length > 0).toEqual(true);
+        for (var shaderID in shaders) {
+            var shader = shaders[shaderID];
```
lilleyse: When looping over keys, it's safer to check for `hasOwnProperty` like https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/addDefaults.js#L19-L21.

specs/lib/addDefaultsSpec.js
```diff
@@ -244,6 +245,14 @@ describe('addDefaults', function() {
...
+            expect(defined(shader.extras._pipeline.source)).toEqual(true);
```
lilleyse: Use `.toBeDefined()` instead.

lilleyse:
This should be ready to go once https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/70 is in.

likangning93:
Okay. Updated!

lasalvavida:
@likangning93 You have jsHint failures

likangning93:
I think those are outside my files though. Should I still address them here?

lasalvavida:
Your branch has a committed half-merged file in `bin/gltf-pipeline.js` which is causing the failure.

lasalvavida:
If you pull master and push I think it should fix it.

likangning93:
Thanks Rob! Updated.

lilleyse:
@likangning93 can you merge in master quickly?

lasalvavida:
Looks good! Thanks @likangning93

PR #80, opened 2016-06-13
Cache optimization

JoshuaStorm:
Adds initial cache optimization step with Tipsify. 

See #1 -  Stage for pre and post vertex cache optimization. Use Cesium's implementation.

Any thoughts on making a more specific test for tipsifyGltf without completely copy-pasting the function to get access to the indices array?

~~I added Cesium_Ground.gltf as boxTexturedUnoptimized's indices seem to already be in cache optimized order.~~


lib/tipsifyGltf.js
```diff
@@ -0,0 +1,31 @@
...
+            var primitivesLength = primitives.length;
+            for (var i = 0; i < primitivesLength; i++) {
+                var primitive = primitives[i];
+                var indexAccessorId = primitive.indices;
+                var indexAccessor = accessors[indexAccessorId];
+                var indicesArray = readAccessor(gltf, indexAccessor);
+                
+                Tipsify.tipsify({
+                    indices: indicesArray.data,
+                    maximumIndex: indexAccessor.max,
```
lilleyse: `max` is an array so you need to send in the 0th element.

lib/tipsifyGltf.js
```diff
@@ -0,0 +1,31 @@
...
+            var mesh = meshes[meshId];
+            var primitives = mesh.primitives;
+            var primitivesLength = primitives.length;
+            for (var i = 0; i < primitivesLength; i++) {
+                var primitive = primitives[i];
+                var indexAccessorId = primitive.indices;
+                var indexAccessor = accessors[indexAccessorId];
+                var indicesArray = readAccessor(gltf, indexAccessor);
+                
+                Tipsify.tipsify({
```
lilleyse: Just calling the function isn't enough, you'll have to also set the values back to the buffer stored in `extras._pipeline.source`. Maybe you could write a `writeAccessor` type function to handle writing back into the accessor.

specs/lib/tipsifyGltfSpec.js
```diff
@@ -0,0 +1,20 @@
...
+var gltfPath = './specs/data/tipsifyGltf/Cesium_Ground.gltf';
+
+describe('tipsifyGltf', function() {
+    it('reorders indices', function(done) {
+        readGltf(gltfPath, function(gltf) {
+            addDefaults(gltf);
+            var copyGltf = clone(gltf);
+            tipsifyGltf(gltf);
+
+            expect(gltf).not.toEqual(copyGltf);
```
lilleyse: I was suspicious that this passed, and it looks like it always will. I commented the tipsify line and it passes, meaning clone is always changing something about the gltf. `expect(clone(gltf)).toEqual(copyGltf);` seems to be accurate though, so you should also make that change in `gltfPipelineSpec.js` and anywhere else that does that. For this particular test though, I would just compare the indices alone, not the whole gltf.

lib/tipsifyGltf.js
JoshuaStorm: I was looking at the GeometryPipeline as a guide and it seemed to modify in place, am I missing something here?

lilleyse: It works a bit differently in Cesium because the indices array sent in is the one that eventually gets used. But in this case the array is just a temporary object and won't be saved with the gltf.

specs/lib/tipsifyGltfSpec.js
JoshuaStorm: I think I'll do this clone fix real quick in a separate PR so it isn't held back by the rest of this PR.

lilleyse: Ok sounds good.

lib/tipsifyGltf.js
JoshuaStorm: I am currently working on a writeAccessor function, but am at a bit of standstill, which may be caused by an initial misunderstanding of my own:

Currently, I am grabbing the indices from each accessor and tipsifying each set of indices relative to only the indices within that accessor--I am fairly certain this is the way we want to do it.

However, I need to eventually write these tipsified indices to the buffer state, but I cannot do that in the same scope as reading in the accessor, because at that scope I am only looking a specific bufferView and not the entire buffer.

The only solution I am seeing is creating some form of symbol table with keys of the initial byte location and values of the bufferView indices arrays and then writing that to a buffer _after the tipsify loop_. That is a somewhat cumbersome solution, though.

In other words, I'd be doing more of a writeBuffer outside of the loop instead of multiple writeAccessors within the loop since I don't know of a way to just change a single bufferView without rewriting the buffer each time.

Am I missing something? Just wanted to make sure before I go ahead with the cumbersome solution.

I hope I am not missing anything super obvious here. Thanks!

lilleyse: As we talked about offline, just writing back into the buffer at the correct offset is fine to do.

JoshuaStorm:
Updated. Still need to update tipsifyGltfSpec

lilleyse:
Have you tried running an exported model through the model inspector? It doesn't seem to load correctly, though there are no errors.

JoshuaStorm:
Alright, I think I am good to go for review on this now.

lib/gltfPipeline.js
```diff
@@ -40,19 +41,21 @@ function processJSONWithExtras(gltfWithExtras, options, callback) {
     var stats = new OptimizationStatistics();
 
     addDefaults(gltfWithExtras, stats);
-
+
```
lilleyse: Remove this whitespace.

lib/gltfPipeline.js
```diff
@@ -40,19 +41,21 @@ function processJSONWithExtras(gltfWithExtras, options, callback) {
...
     addDefaults(gltfWithExtras, stats);
-
+    
     removeUnused(gltfWithExtras, stats);
     if (options.printStats) {
         printStats(stats);
     }
     convertDagToTree(gltfWithExtras);
-
+
```
lilleyse: Same

lib/writeAccessor.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+var byteLengthForComponentType = require('./byteLengthForComponentType');
+var numberOfComponentsForType = require('./numberOfComponentsForType');
+var writeBufferComponentType = require('./writeBufferComponentType');
+
+module.exports = writeAccessor;
+
+function writeAccessor(gltf, accessor, dataArray) { // Argument name?
```
lilleyse: The name is fine, you can remove the comment.

lib/writeAccessor.js
```diff
@@ -0,0 +1,26 @@
...
+
```
lilleyse: Remove empty line.

lib/writeAccessor.js
```diff
@@ -0,0 +1,26 @@
...
+    var accessorType = accessor.type;
+    var bufferView = gltf.bufferViews[accessor.bufferView];
+
+    var attributeCount = accessor.count;
+    var componentCount = numberOfComponentsForType(accessorType);
+    var componentType = accessor.componentType;
+    var componentBytes = byteLengthForComponentType(componentType);
+    var currentBufferIndex = accessor.byteOffset + bufferView.byteOffset;
+
+    var buffer = gltf.buffers[bufferView.buffer].extras._pipeline.source;
```
lilleyse: For organization purposes just put this under the `bufferView` line.

lib/writeAccessor.js
```diff
@@ -0,0 +1,26 @@
...
+
+
```
lilleyse: Remove extra line.

specs/lib/tipsifyGltfSpec.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+var tipsifyGltf = require('../../lib/tipsifyGltf');
+var addDefaults = require('../../lib/addDefaults');
+var readGltf = require('../../lib/readGltf');
+var writeAccessor = require('../../lib/writeAccessor');
```
lilleyse: This include isn't used.

specs/lib/tipsifyGltfSpec.js
```diff
@@ -0,0 +1,24 @@
...
+var readAccessor = require('../../lib/readAccessor');
+
+var gltfPath = './specs/data/tipsifyGltf/chinesedragon.gltf';
+
+describe('tipsifyGltf', function() {
+    it('reorders indices', function(done) {
+        readGltf(gltfPath, function(gltf) {
+            addDefaults(gltf);
+            var indexAccessorId = gltf.meshes[Object.keys(gltf.meshes)].primitives[0].indices;
```
lilleyse: Should be `Object.keys(gltf.meshes)[0]`.

lilleyse:
I'm starting to think that `tipsifyGltf.js` should be renamed to be more descriptive of the stage, like `cacheOptimizeIndices.js` or something.

JoshuaStorm:
Referencing back to #1 

"Stage for _pre_ and _post_ vertex cache optimization. Use Cesium's implementation."

Should cache optimization be applied prior to and at the end of the pipeline?

lilleyse:
The pre/post terminology is just describing the technique used by the gpu. (https://www.opengl.org/discussion_boards/showthread.php/167481-Vertex-cache-optimization). Only doing it once in our pipeline is fine.

lib/cacheOptimizeIndices.js
```diff
@@ -0,0 +1,32 @@
...
+            var primitivesLength = primitives.length;
+            for (var i = 0; i < primitivesLength; i++) {
+                var primitive = primitives[i];
+                var indexAccessorId = primitive.indices;
+                var indexAccessor = accessors[indexAccessorId];
+                var indicesArray = readAccessor(gltf, indexAccessor);
+
+                indicesArray.data = Tipsify.tipsify({
+                    indices: indicesArray.data,
+                    maximumIndex: undefined,
```
lilleyse: My comment may have gotten lost, but you can remove this line.

lilleyse:
Make sure to also change the spec name and the spec/data folder name.

lilleyse:
Looks good!

specs/lib/cacheOptimizeIndicesSpec.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+var cacheOptimizeIndices = require('../../lib/cacheOptimizeIndices');
+var addDefaults = require('../../lib/addDefaults');
+var readGltf = require('../../lib/readGltf');
+var readAccessor = require('../../lib/readAccessor');
+
+var gltfPath = './specs/data/cacheOptimizeIndices/chinesedragon.gltf';
```
pjcozzi: How many triangles is this model?  To keep the repo small and the tests fast, why not just run this on a box?  The Cesium function is already unit tested, we just need to test the integration.

lib/cacheOptimizeIndices.js
```diff
@@ -0,0 +1,31 @@
...
+            var primitivesLength = primitives.length;
+            for (var i = 0; i < primitivesLength; i++) {
+                var primitive = primitives[i];
+                var indexAccessorId = primitive.indices;
+                var indexAccessor = accessors[indexAccessorId];
+                var indicesArray = readAccessor(gltf, indexAccessor);
+
+                indicesArray.data = Tipsify.tipsify({
+                    indices: indicesArray.data,
+                    cacheSize: 24
```
pjcozzi: This should be a parameter that we expose.

specs/lib/cacheOptimizeIndicesSpec.js
```diff
@@ -0,0 +1,23 @@
...
+    it('reorders indices', function(done) {
+        readGltf(gltfPath, function(gltf) {
+            addDefaults(gltf);
+            var indexAccessorId = gltf.meshes[Object.keys(gltf.meshes)[0]].primitives[0].indices;
+            var indicesBefore = readAccessor(gltf, gltf.accessors[indexAccessorId]);
+            
+            cacheOptimizeIndices(gltf);
+            
+            var indicesAfter = readAccessor(gltf, gltf.accessors[indexAccessorId]);
+            expect(indicesBefore).not.toEqual(indicesAfter);
```
pjcozzi: This should check for specific values, and should also check vertices since they may change for the pre-vertex shader cache.

specs/lib/cacheOptimizeIndicesSpec.js
JoshuaStorm: The cache optimization seems to have no effect on the usual box we use for testing -- I assumed this was reasonable since there are so few indices. Let me see if I can figure out how to reorder the indices such that they will be moved in a known manner with the cache optimization so I can avoid using the dragon model.

pjcozzi:
> The cache optimization seems to have no effect on the usual box we use for testing -- I assumed this was reasonable since there are so few indices. Let me see if I can figure out how to reorder the indices such that they will be moved in a known manner with the cache optimization so I can avoid using the dragon model.

Or just try another simple test model if we have one.

specs/lib/cacheOptimizeIndicesSpec.js
JoshuaStorm: > should also check vertices since they may change for the pre-vertex shader cache.

Right now I believe I am only doing the post vertex cache optimization with tipsify. Another pull request will implement the pre vertex cache optimization

PR #82, opened 2016-06-13
Added generateNormals stage

lilleyse:
This adds support for generating normals if they do not exist.


pjcozzi:
@lilleyse consider just using the function in Cesium: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/GeometryPipeline.js#L1059-L1192

Or improve the version in Cesium and use that.

lilleyse:
Yeah I'll likely go that route. @JoshuaStorm is going to write a gltf primitive <-> Geometry converter so that we can use Cesium functions like these easily.

JoshuaStorm:
@lilleyse: In `generateNormalsSpec`, shouldn't we be able to check for equality between `box_normals` and `box_no_normals` after it is run through `generateNormals`? or am I misunderstanding something?

lilleyse:
I'm not sure if the normals in the `box_normals` model would be the same as those that are autogenerated.

JoshuaStorm:
I think I might be misunderstanding:

These are just vertex normals in the geometric sense, correct? If that is the case, shouldn't they be derived, in other words be the same whether they are provided or calculated? 

Sorry for the pedantic question, I just want to make sure I am understanding this. 

The current result I am getting is shown below, which I am somewhat suspicious is the correct normal values but the indices are out of order somehow.

generatednormals
.

lib/generateNormals.js
```diff
@@ -0,0 +1,152 @@
...
+            for (j = 0; j < normalsLength; j++) {
+                normals[j] = new Cartesian3();
+            }
+
+            // Add face normal to each vertex normal
+            for (j = 0; j < trianglesLength; ++j) {
+                var index1 = indices[j * 3 + 0];
+                var index2 = indices[j * 3 + 1];
+                var index3 = indices[j * 3 + 2];
+                var faceNormal = getFaceNormal(positions[index1], positions[index2], positions[index3], scratchNormal);
```
lasalvavida: Maybe I'm misreading, but does this just set the vertex normal to be the face normal? That won't be the correct vector.

pjcozzi:
@JoshuaStorm to debug you can tweak the fragment shader (even in the Model Inspector) so the output color is the normal.

The computed normal may not be the same as a provided normal, as a user could provide whatever they want.

lib/generateNormals.js
lilleyse: It adds all the face normals that the vertex is a part of, and then normalizes.

lasalvavida: Oh, I see thank you

JoshuaStorm:
Looking more into this, the issue does seem to be what we discussed offline @pjcozzi. In both this implementation and the `GeometryPipeline` implementation of calculating normals, the face normals are just averaged resulting in odd normals at sharp edge vertices. In the case above, the cube is only made up of these sharp edge vertices, resulting in its especially weird normals. 

I plan on looking more into the duplicating vertices at edges solution that you also mentioned, @pjcozzi, but I was wondering if I should plan on implementing it here (and not use the GeometryPipeline) or if I should plan on implementing it in `Cesium.GeometryPipeline`.

pjcozzi:
@JoshuaStorm the implementation should go into Cesium proper, then gltf-pipeline will pick it up in the next release via npm.  You can, of course, code it in this repo to test it, but you'll only want commits to go to the Cesium repo.

JoshuaStorm:
I am at a bit of a loss if anyone else has any ideas they'd be appreciated!

As a naive test to confirm the shared vertices issue, I iterated through the indices and in any case where there was multiple references to the same vertex, I duplicated that vertex and had the respective index refer to the new duplicate--effectively getting rid of shared vertices by simply duplicating all shared vertices.

Then generating normals for that, I ended up with this box:

weirdnormals

With RGB mapped to normal value

colorweirdnormals


I'm wondering if it is possibly for some of the vertices to be in the incorrect winding order?

pjcozzi:
Maybe, perhaps hack up the code so only 3 of the 6 faces are in the model.

lasalvavida:
@JoshuaStorm I took a stab at this and its pretty close now.

normal-box-front

The back normals look like they might be off.

normal-box-back

Could you take a look at this?

JoshuaStorm:
@lasalvavida Your box seems to mirror that of the included `box_normals.gltf`. Diffing the two files seems to confirm this as they only differ in a few naming schemes and a bit of buffer data (presumably from the extra data from duplicating the vertices). I'm assuming this means generate normals is doing its job properly? Perhaps not optimally

I could start implementing this solution into Cesium's `GeometryPipeline` and converting this to use the GometryPipeline implementation if you're happy with this.

A quick question about the code: what exactly is the purpose for finding the center of the model?

lasalvavida:
@JoshuaStorm It was some scratch code we can probably take out. I was using it to check if the normals were facing the center of the model and to flip them out if they were. However, it never got called, so I took it out and must have missed the center calculation.

I'd say before doing anything, we need an answer as to why the faces are shaded black. Is it the shader, or is there still something wrong with the data?

lilleyse:
I'm getting a crash when running the dragon model through:


[dragon_no_normals.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/348369/dragon_no_normals.gltf.txt)

JoshuaStorm:
@lilleyse the crash from `dragon_no_normals.gltf` appears to be due to the index values exceeding 16 bit values when using "hard" normal generation. I initially tried detecting this an bumping it to 32UInts but it turns out indices are required to be `UNSIGNED_BYTE` or `UNSIGNED_SHORT` (wish I thought to check the spec first...)

Perhaps I could find a way to remove duplicate vertex/normal combinations within the `generateNormals` stage before we try to write to the accessor?

lasalvavida:
@JoshuaStorm You're better off splitting off the excess into a new primitive. Once it is implemented,[#619](https://github.com/KhronosGroup/glTF/pull/619) will be an option as well.

JoshuaStorm:
Updated using `GeometryPipeline.computeNormal` 

This still results in unusual corner cases like the box, but works well for models like the dragon. This even works well for models with UVs.

I'll add a few pictures in a moment:

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -1,10 +1,89 @@
 'use strict';
+var findAccessorMinMax = require('./findAccessorMinMax');
 var writeAccessor = require('./writeAccessor');
+var mergeBuffers = require('./mergeBuffers');
+var uninterleaveAndPackBuffers = require('./uninterleaveAndPackBuffers');
+var numberOfComponentsForType = require('./numberOfComponentsForType');
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+var WebGLConstants = Cesium.WebGLConstants;
```
lilleyse: Reorganize these based on the node guide.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -1,10 +1,89 @@
...
+
 
 module.exports = cesiumGeometryToGltfPrimitive;
 
+function createAttributeSemantic(gltf, primitive, semantic, attributeLength) {
+    var type;
+    var bytesPerComponent;
+
+    switch (semantic.replace(/([0-9]|_)/g, '')) {
```
lilleyse: This regex doesn't work for `TEXCOORD_0` (Also fix in `gltfPrimitiveToCesiumGeometry`).

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -1,10 +1,89 @@
...
+        case 'POSITION':
+            type = 'VEC3';
+            bytesPerComponent = 8;
+            break;
```
lilleyse: While the Cesium geometry positions are doubles, they should be converted back to floats for gltf `bytesPerComponent = 4`.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -1,10 +1,89 @@
...
+        case 'TEXCOORD':
+            type = 'VEC2';
+            bytesPerComponent = 4;
+    }
+
+    var componentsPerType = numberOfComponentsForType(type);
+
+    var bufferLength = attributeLength * componentsPerType * bytesPerComponent;
+    var buffer = new Buffer(bufferLength);
```
lilleyse: Unless I'm missing something, this buffer isn't getting filled with data.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -1,10 +1,89 @@
...
+        byteOffset : 0,
+        byteStride : 0,
+        componentType : WebGLConstants.FLOAT,
+        count : attributeLength,
+        type : type
+    };
+
+    gltf.accessors[accessorId] = accessor;
+
+    // Find min/max for normal accessor
```
lilleyse: Change the comment since it happens for all attribute accessors.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -1,10 +1,89 @@
...
+    var minMax = findAccessorMinMax(gltf, accessor);
+    accessor.min = minMax.min;
+    accessor.max = minMax.max;
+
+    primitive.attributes[semantic] = accessorId;
+    mergeBuffers(gltf, 'buffer_0');
+    uninterleaveAndPackBuffers(gltf);
```
lilleyse: The merge and interleave should happen at the end of `cesiumGeometryToGltfPrimitive` just so we don't call this for every attribute.

lib/generateNormals.js
```diff
@@ -0,0 +1,68 @@
...
+var gltfPrimitiveToCesiumGeometry = require('./gltfPrimitiveToCesiumGeometry');
+var cesiumGeometryToGltfPrimitive = require('./cesiumGeometryToGltfPrimitive');
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+var WebGLConstants = Cesium.WebGLConstants;
+var GeometryPipeline = Cesium.GeometryPipeline;
+
+module.exports = generateNormals;
+
+
```
lilleyse: Remove extra line.

lib/generateNormals.js
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Generates normals for primitives if they do not exist.
+ *
+ * @param gltf
+ * @param {boolean} isHard
```
lilleyse: Remove this.

lib/generateNormals.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+function generateNormals(gltf) {
+    var generatedNormals = false;
```
lilleyse: Remove.

lib/generateNormals.js
```diff
@@ -0,0 +1,68 @@
...
+        var positionAccessorId = primitive.attributes.POSITION;
+        var normalAccessorId = primitive.attributes.NORMAL;
+        var indicesAccessorId = primitive.indices;
+        var positionAccessor = gltf.accessors[positionAccessorId];
+        var position = [];
+        readAccessor(gltf, positionAccessor, position);
+
+        var options = {
+            attributeLength : position.length
+        };
```
lilleyse: This options thing is somewhat confusing, and it would be better is `cesiumGeometryToGltfPrimitive` didn't need to use it at all. Can't it get the attribute length from the Cesium geometry?

lib/generateNormals.js
```diff
@@ -0,0 +1,68 @@
...
+
+        if (primitive.mode === WebGLConstants.TRIANGLES && defined(indicesAccessorId) && defined(positionAccessorId) && !defined(normalAccessorId)) {
+            generatedNormals = true;
```
lilleyse: And remove this.

lib/gltfPipeline.js
```diff
@@ -25,6 +26,7 @@ var quantizeAttributes = require('./quantizeAttributes');
 var writeGltf = require('./writeGltf');
 var writeBinaryGltf = require('./writeBinaryGltf');
 var writeSource = require('./writeSource');
+var generateNormals = require('./generateNormals');
```
lilleyse: Remove this (already included).

lib/gltfPipeline.js
```diff
@@ -58,9 +60,12 @@ function processJSON(gltf, options, callback) {
 }
 
 function processJSONWithExtras(gltfWithExtras, options, callback) {
+    var hard = defaultValue(options.hardNormals, false);
```
lilleyse: Remove this.

lib/gltfPipeline.js
```diff
@@ -58,9 +60,12 @@ function processJSON(gltf, options, callback) {
...
     var stats = new OptimizationStatistics();
-    addDefaults(gltfWithExtras, options);
+    generateNormals(gltfWithExtras);
```
lilleyse: I forget, is there a reason `generateNormals` is before `addDefaults`?

lib/gltfPipeline.js
```diff
@@ -58,9 +60,12 @@ function processJSON(gltf, options, callback) {
...
+    addDefaults(gltfWithExtras, stats);
     removeUnused(gltfWithExtras, stats);
+    generateNormals(gltfWithExtras, hard);
```
lilleyse: Remove this.

lib/gltfPrimitiveToCesiumGeometry.js
```diff
@@ -7,9 +7,39 @@ var Geometry = Cesium.Geometry;
...
+            }
+            return packed;
+        case 'Cartesian4':
+            packed = new Array(length * 4);
+            for (i = 0; i < length; ++i) {
+                Cartesian4.pack(values.data[i], packed, i * 4);
+            }
+            return packed;
+    }
+}
```
lilleyse: Now that Cesium 1.23 is on npm, we can use the `CartesianX.packArray` functions instead.

lilleyse:
Remove `box_uvs_no_normals.gltf`, `dragon_no_normals.gltf`, and `dragon_uvs_no_normals.gltf`.

lib/cesiumGeometryToGltfPrimitive.js
JoshuaStorm: Are you sure about this? Just doing a quick

`var semantic = 'TEXCOORD_0'`
`semantic.replace(/([0-9]|_)/g, '')`
returns `'TEXCOORD'` as we want. Did an issue occur on a specific gltf you passed in?

lilleyse: Never mind, I forgot about the `g`

lib/cesiumGeometryToGltfPrimitive.js
JoshuaStorm: That's definitely a problem. Thanks!

JoshuaStorm: Took me a bit to remember what my code was doing, but I do believe the buffer is eventually written to: `createAttributeSemantic` creates the buffer that `mapGeometryAttributeToPrimitive` writes to. 

Is this not actually writing to the buffer?

lilleyse: Ah I see now, just leave a comment explaining what's going on.

lib/gltfPipeline.js
JoshuaStorm: No particular reason, that I recall

lilleyse: Ok then after makes more sense.

JoshuaStorm:
Updated. I'm not sure why the PR build is failing since `encodeImages` is untouched by this PR

lasalvavida:
It's not you; it's in master; I just opened #136 to fix it

lasalvavida:
@JoshuaStorm pull master; it should be fixed now

JoshuaStorm:
Good to go when someone has a chance to review. @lilleyse are we trying to get this in before publishing to npm?

lilleyse:
Yup this and ao. I'll review soon.

lib/generateNormals.js
```diff
@@ -0,0 +1,60 @@
...
+    var primitives = getPrimitives(gltf);
+    var primitivesLength = primitives.length;
+    for (var i = 0; i < primitivesLength; i++) {
+        var primitive = primitives[i];
+        var positionAccessorId = primitive.attributes.POSITION;
+        var normalAccessorId = primitive.attributes.NORMAL;
+        var indicesAccessorId = primitive.indices;
+        var positionAccessor = gltf.accessors[positionAccessorId];
+        var position = [];
+        readAccessor(gltf, positionAccessor, position);
```
lilleyse: You can remove these three lines.

lib/gltfPipeline.js
```diff
@@ -61,7 +62,8 @@ function processJSON(gltf, options, callback) {
 
 function processJSONWithExtras(gltfWithExtras, options, callback) {
     var stats = new OptimizationStatistics();
-    addDefaults(gltfWithExtras, options);
+    generateNormals(gltfWithExtras);
```
lilleyse: Sorry I wasn't clear with this, but `generateNormals` should go after the initial stages. Probably after `printStats` is good.

lilleyse:
test
**Box with UVs and no normals** (The normal accessor maxes are pretty crazy for this. You should investigate if this is a Cesium issue or a gltf-pipeline isse)
test2
**Box with no UVs and no normals** Looks correct.
test3
**Dragon with no UVs and no normals** Looks correct.
test4
**Dragon with UVs and no normals** The dragon body is correct but the spheres are not. I'm pretty sure the body has UVs and the spheres don't, so there may be some edge case going on. The min/max values for the accessor are really strange here.
[gltfs.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/366275/gltfs.zip)

JoshuaStorm:
Update on the current state of this, I've been smashing my head into this for a bit, so if someone wants to give it a look with fresh eyes I'd appreciate it, especially since this is holding back publishing. I wouldn't be surprised if I am overlooking something obvious.

**Box with UVs and no normals**: Same results as before
box_uvs_no_normals

**Models with no normals and no UVs**: still working same as before, properly (or at least seemingly)

**Dragon with UVs and no normals**: The body is still generating normals fine, but the second primitive is not (eyes + sphere)
dragon_uvs_no_normals


We suspected that the sphere and eyes had no UVs in the model, but they do in fact. However, even removing the UVs before trying to generate normals resulted in the same odd results. This is what has me most confused
butchered

**Shapes test**: Sean generated this for me, the cylinder and box have UVs, the cone and sphere do not. This has similar results to the dragon: when there are UVs, the primitives that have UVs generate normals properly but the primitives that do not have UVs generate the incorrect normals.
uvs_and_no_uvs


I may have just beat my head into this for too long and need to look at it with fresh eyes tomorrow.

JoshuaStorm:
Here are the models used. `butchered.gltf` is the model created by removing the dragon's body primitive and removing the other primitive's UVs

[Models.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/366856/Models.zip)

JoshuaStorm:
I think this is finally ready for review!

Thanks @lasalvavida for pointing out the issue with generating normals before the combine and remove stages.

lasalvavida:
If I pull master into this branch and run with `--ao` I get the following exception:

> Unhandled rejection TypeError: Cannot assign to read only property 'gltfWithExtras' of C:\Users\rtaglang\AGI\git\gltf-pipeline
    at processJSONWithExtras (c:\Users\rtaglang\AGI\git\gltf-pipeline\lib\gltfPipeline.js:92:34)
    at c:\Users\rtaglang\AGI\git\gltf-pipeline\lib\gltfPipeline.js:168:9
    at c:\Users\rtaglang\AGI\git\gltf-pipeline\lib\readGltf.js:43:17
    at tryCatcher (c:\Users\rtaglang\AGI\git\gltf-pipeline\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (c:\Users\rtaglang\AGI\git\gltf-pipeline\node_modules\bluebird\js\release\promise.js:504:31)
    at Promise._settlePromise (c:\Users\rtaglang\AGI\git\gltf-pipeline\node_modules\bluebird\js\release\promise.js:561:18)
    at Promise._settlePromise0 (c:\Users\rtaglang\AGI\git\gltf-pipeline\node_modules\bluebird\js\release\promise.js:606:10)
    at Promise._settlePromises (c:\Users\rtaglang\AGI\git\gltf-pipeline\node_modules\bluebird\js\release\promise.js:685:18)
    at Async._drainQueue (c:\Users\rtaglang\AGI\git\gltf-pipeline\node_modules\bluebird\js\release\async.js:138:16)
    at Async._drainQueues (c:\Users\rtaglang\AGI\git\gltf-pipeline\node_modules\bluebird\js\release\async.js:148:10)
    at Immediate.Async.drainQueues [as _onImmediate] (c:\Users\rtaglang\AGI\git\gltf-pipeline\node_modules\bluebird\js\release\async.js:17:14)
    at processImmediate [as _immediateCallback] (timers.js:383:17)

JoshuaStorm:
@lasalvavida What model was this produced with? Worked fine for `box_uv_no_normals` for me, awaiting `dragon_uvs_no_normals` now

lasalvavida:
The above seems to be a problem in my WebStorm, it works fine from bash so you can disregard it

specs/lib/gltfPipelineSpec.js
```diff
@@ -120,7 +120,7 @@ describe('gltfPipeline', function() {
         readGltf(gltfEmbeddedPath, options, function (gltf) {
             var initialUri = gltf.buffers.CesiumTexturedBoxTest.uri;
             processJSON(gltf, options, function () {
-                var finalUri = gltf.buffers.CesiumTexturedBoxTest.uri;
+                var finalUri = gltf.buffers.uri;
```
lasalvavida: Why did you make this change? If you do this you're comparing at different levels of the glTF hierarchy.

JoshuaStorm: Oops, this was a naive change since `processFile` will now merge buffers, so `gltf.buffers.CesiumTexturedBoxTest` will no longer exist.

This could now expect the old buffer to be undefined and compare initialUri with `gltf.buffer_0.uri` (the only buffer that should be on the gltf), but that doesn't seem to be perfect test either. Thoughts?

lasalvavida: Since in this test, there can only be one buffer (I assume). Just get the first buffer:
```javascript
var firstBufferId = Object.keys(gltf.buffers)[0];
var testBuffer = gltf.buffers[firstBufferId];
```

specs/lib/gltfPipelineSpec.js
```diff
@@ -118,9 +118,13 @@ describe('gltfPipeline', function() {
     it('will write sources from JSON', function(done) {
         var options = {};
         readGltf(gltfEmbeddedPath, options, function (gltf) {
-            var initialUri = gltf.buffers.CesiumTexturedBoxTest.uri;
+            var firstBufferId = Object.keys(gltf.buffers)[0];
+            var testBuffer = gltf.buffers[firstBufferId];
+            var initialUri = testBuffer.uri;
```
lasalvavida: I don't think you need to do this here, just after processJSON gets run, right? You do know the initial bufferId is CesiumTexturedBoxTest.

lib/gltfPipeline.js
```diff
@@ -64,7 +65,7 @@ function processJSON(gltf, options, callback) {
 
 function processJSONWithExtras(gltfWithExtras, options, callback) {
     var stats = new OptimizationStatistics();
-    addDefaults(gltfWithExtras, options);
+    addDefaults(gltfWithExtras, stats);
```
lasalvavida: Why did this change? I'm pretty sure this should still be options.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -1,10 +1,74 @@
...
 var writeAccessor = require('./writeAccessor');
 
+var WebGLConstants = Cesium.WebGLConstants;
+
 module.exports = cesiumGeometryToGltfPrimitive;
 
+// Creates a buffer location to write the new values to. mapGeometryAttributeToPrimitive does the actual writing
+function createAttributeSemantic(gltf, primitive, semantic, packedLength) {
+    var type;
+    switch (semantic.replace(/([0-9]|_)/g, '')) {
```
lasalvavida: @lilleyse Can have the final verdict on this one, but I'm pretty sure I remember saying not to do this in an earlier revision.

It's not the worst thing in the world, but you do make a new string every single time this statement executes which is a little wasteful.

This should probably be more like:
```javascript
if (semantic.indexOf('POSITION') === 0 || semantic.indexOf('NORMAL') === 0) {
    type = 'VEC3';
} else if (semantic.indexOf('TEXCOORD') === 0) {
    type = 'VEC2';
}
```

specs/lib/generateNormalsSpec.js
```diff
@@ -0,0 +1,35 @@
...
+
+var gltfNoNormalsPath = './specs/data/generateNormals/box_no_normals.gltf';
+var gltfNormalsPath = './specs/data/generateNormals/box_normals.gltf';
+
+describe('generateNormals', function(){
+    it('generates normals if they do not exist', function(done) {
+        var options = {};
+        readGltf(gltfNoNormalsPath, options, function(gltf){
+            var attributes = gltf.meshes.mesh_box.primitives[0].attributes;
+            var byteLengthBefore = gltf.buffers[Object.keys(gltf.buffers)[0]].byteLength;
```
lasalvavida: Same comment as in `gltfPipelineSpec`

lib/cesiumGeometryToGltfPrimitive.js
lilleyse: I don't have a strong preference either way, your suggestion is more compact and a little clearer, so I say go with that.

JoshuaStorm: Odd, I could have sworn I already made this change.

lasalvavida:
I will note there are some lighting artifacts on some more complex models:

lighting artifacts

but, I'm fairly certain that's a hard vs. soft normals issue.

lasalvavida:
A similar effect on buggy:

buggy-lighting-artifacts

lasalvavida:
I just wanted to include the above comments as a frame of reference for when we add support for generating hard normals. This looks great to me, thanks @lilleyse and @JoshuaStorm!

lilleyse:
Thanks for reviewing @lasalvavida

PR #84, opened 2016-06-14
Min and max must be enforced to prevent errors in quantized attributes

lasalvavida:
I was converting some of the glTF sample models to test quantized attributes and the 2CylinderEngine model has an accessor max value that is slightly less than the actual max in the data.

This can cause out of bounds errors when writing into the buffer since the encoded value ends up being greater than 65535.

### DIFF
--- a/lib/quantizeAttributes.js
+++ b/lib/quantizeAttributes.js
@@ -110,7 +110,7 @@ function quantizeAttributes(gltf, options) {
                     value = source.readFloatLE(i + j * componentByteLength);
                     value = Math.min(value, max[j]);
                     value = Math.max(value, min[j]);
-                    var encoded = Math.round((value - min[j]) * range / (max[j] - min[j]))
+                    var encoded = Math.round((value - min[j]) * range / (max[j] - min[j]));
                     source.writeUInt16LE(encoded, i + j * 2);
                 }
                 num++;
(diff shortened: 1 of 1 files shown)


pjcozzi:
@lasalvavida we should not need this workaround; instead, the model should be fixed since it doesn't follow the spec.

@lexaknyazev does the validator already account for this?

lexaknyazev:
> does the validator already account for this?

Binary data validation is work in progress. Shouldn't take much time to implement.

pjcozzi:
Very nice, sounds great!

lasalvavida:
@pjcozzi To provide a little more background, it's an issue of precision in this particular case. The 2CylinderEngine sample model has its max reported correctly, but truncated to 3 decimal places. The actual max ends up being greater than this because it wasn't rounded up. 

What do you think about maybe just throwing an exception here instead with a little more information?

lexaknyazev:
Looks like it's a good point for 1.0.1 spec update: how exporters must truncate / floor / ceil floats.

pjcozzi:
> Looks like it's a good point for 1.0.1 spec update: how must exporters truncate / floor / ceil floats.

Agreed, do you want to open up a pull request into the 1.0.1 glTF branch with the proposed wording?

> What do you think about maybe just throwing an exception here instead with a little more information?

The validator will check for this so I would not test it here just like we don't check for the hundreds of other spec rules.  However, if this becomes a very common problem in the glTF ecosystem, then we could use the workaround here, but this is the exact type of ecosystem problem that glTF is meant to fix.

lasalvavida:
I'm going to close this for now. We can revisit it if this becomes a serious issue.

lexaknyazev:
See https://github.com/KhronosGroup/glTF/pull/628.

Issue #85, opened 2016-06-15
BrainStem sample model combinePrimitives throws exception


lasalvavida:
Probably related to #83

> buffer.js:835
>     throw new TypeError('value is out of bounds');
>     ^
> 
> TypeError: value is out of bounds
>     at checkInt (buffer.js:835:11)
>     at Buffer.writeUInt16LE (buffer.js:893:5)
>     at mergeAccessors (C:\Users\rtaglang\AGI\git\gltf-pipeline\lib\combinePrimitives.js:233:36)
>     at combinePrimitiveGroup (C:\Users\rtaglang\AGI\git\gltf-pipeline\lib\combinePrimitives.js:91:36)
>     at combineMeshPrimitives (C:\Users\rtaglang\AGI\git\gltf-pipeline\lib\combinePrimitives.js:49:33)
>     at combinePrimitives (C:\Users\rtaglang\AGI\git\gltf-pipeline\lib\combinePrimitives.js:19:43)
>     at processJSONWithExtras (C:\Users\rtaglang\AGI\git\gltf-pipeline\lib\gltfPipeline.js:49:5)
>     at C:\Users\rtaglang\AGI\git\gltf-pipeline\lib\gltfPipeline.js:63:9
>     at C:\Users\rtaglang\AGI\git\gltf-pipeline\lib\readGltf.js:42:13
>     at C:\Users\rtaglang\AGI\git\gltf-pipeline\lib\loadGltfUris.js:18:9

lasalvavida:
Pretty sure this happens because the BrainStem model has more than 65535 total vertices, so when combining, the index value exceeds the possible value.

combinePrimitives should be able to make multiple primitives to prevent this issue. I'll work on this once #88 gets merged.

pjcozzi:
Let's also have a default option to use `OES_element_index_uint`, see https://github.com/KhronosGroup/glTF/pull/619.

lasalvavida:
Closed by #162.

PR #89, opened 2016-06-16
Merge duplicate vertices

lasalvavida:
Adds stage for merging duplicate vertices from the #1 roadmap.

### DIFF
--- a/lib/mergeDuplicateVertices.js
+++ b/lib/mergeDuplicateVertices.js
@@ -104,9 +104,8 @@ function mergeDuplicateVerticesInMesh(gltf, mesh) {
                     }
                     seenIndices[index] = true;
                 }
+                indexByteOffset += indexByteStride;
             }
-            indexByteOffset += indexByteStride;
-
         }
     }
 }
--- a/lib/mergeDuplicateVertices.js
+++ b/lib/mergeDuplicateVertices.js
@@ -3,12 +3,12 @@ var Cesium = require('cesium');
 var DeveloperError = Cesium.DeveloperError;
 var defined = Cesium.defined;
 
-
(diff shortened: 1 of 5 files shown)


lilleyse:
For additional testing purposes, do we have any glTF models that have duplicate vertices?

lib/mergeDuplicateVertices.js
```diff
@@ -0,0 +1,116 @@
...
+var writeBufferComponentType = require('./writeBufferComponentType');
+var removeUnusedVertices = require('./removeUnusedVertices');
+
+module.exports = mergeDuplicateVertices;
+
+/**
+ * Looks for duplicated vertices in a glTF hierarchy and merges them,
+ * modifying indices to account for the change if they exist.
+ *
+ * @param {Object} [gltf] A glTF hierarchy.
```
lilleyse: `gltf` isn't optional, so remove the square braces.

lilleyse:
The code and tests look good to me, I just want to test on a real use case before merging. I'll try to create a model for that.

lasalvavida:
@lilleyse It looks like BrainStem has duplicate vertices, so I'm running some tests on that now.

lilleyse:
Ok nice.

lasalvavida:
@lilleyse Take a look at these changes when you get a chance. The [BrainStem model](https://github.com/KhronosGroup/glTF/tree/master/sampleModels/BrainStem/glTF-Embedded) should go from 15,430 KB down to 4,437 KB (with combine primitives commented out because of #85) with no detrimental effects.

lilleyse:
Changes looks good. Also that's a great improvement! I confirmed that BrainStem works.

PR #90, opened 2016-06-16
Gltf to geometry

JoshuaStorm:
Added to helper methods for converting between gltf and Cesium.Geometry data types. Should be helpful for reusing Cesium.GeometryPipeline code.


lib/modeToPrimitiveType.js
```diff
@@ -0,0 +1,24 @@
...
+        case 2:
+            return PrimitiveType.LINE_LOOP;
+        case 3:
+            return PrimitiveType.LINE_STRIP;
+        case 4:
+            return PrimitiveType.TRIANGLES;
+        case 5:
+            return PrimitiveType.TRIANGLE_STRIP;
+        case 6:
+            return PrimitiveType.TRIANGLE_FAN;
```
lilleyse: Instead of 0, 1, 2, etc use the `WebGLConstants` enums like in `byteLengthForComponentType.js`.

lib/readGltfGeometry.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
```
lilleyse: The words `read` and `write` for these two files are slightly confusing. Maybe `convert` is better.

lib/readGltfGeometry.js
```diff
@@ -0,0 +1,44 @@
...
+    var positionAccessor = gltf.accessors[positionId];
+    var componentsPerAttribute = numberOfComponentsForType(positionAccessor.type);
+    var values = readAccessor(gltf, positionAccessor);
+    
+    var packedValues = [];
+    for (var i = 0; i < values.data.length; i++) {
+        packedValues.push(values.data[i].x);
+        packedValues.push(values.data[i].y);
+        packedValues.push(values.data[i].z);
+    }
```
lilleyse: You can use `Cartesian3.packArray` instead.

lib/readGltfGeometry.js
```diff
@@ -0,0 +1,44 @@
...
+    
+    var position = new GeometryAttribute({
+        'componentDatatype' : ComponentDatatype.FLOAT,
+        'componentsPerAttribute' : componentsPerAttribute,
+        'values' : new Float64Array(packedValues)
```
lilleyse: Remove the quote marks when creating objects. Usually we only use quotes if we are faking JSON data. This applies to a bunch of areas in this PR.

lib/writeGeometryToGltf.js
```diff
@@ -0,0 +1,17 @@
...
+    var positionAccessor = gltf.accessors[positionId];
+
+    var indicesId = primitive.indices;
+    var indicesAccessor = gltf.accessors[indicesId];
+
+    var values = geometry.attributes.position.values;
+
+    writeAccessor(gltf, positionAccessor, values);
+    writeAccessor(gltf, indicesAccessor, geometry.indices);
+}
```
lilleyse: Can you also include support for normals here since I will be using it? First check that the primitive has a NORMAL attribute though.

specs/lib/readGltfGeometrySpec.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+var readGltfGeometry = require('../../lib/readGltfGeometry');
+var readGltf = require('../../lib/readGltf');
+var addDefaults = require('../../lib/addDefaults')
```
lilleyse: Missing semicolon.

specs/lib/readGltfGeometrySpec.js
```diff
@@ -0,0 +1,19 @@
...
+var gltfPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf';
+
+describe('readGltfGeometry', function() {
+    it('returns a geometry', function(done) {
+        readGltf(gltfPath, function(gltf) {
+            addDefaults(gltf);
+            var primitive = gltf.meshes[Object.keys(gltf.meshes)[0]].primitives[0];
+            var geometry = readGltfGeometry(gltf, primitive);
+            
+            expect(geometry).toBeDefined();
```
lilleyse: You should flesh this test out more.

specs/lib/writeGeometrytToGltfSpec.js
```diff
@@ -0,0 +1,50 @@
...
+var readAccessor = require('../../lib/readAccessor');
+var readGltf = require('../../lib/readGltf');
+var addDefaults = require('../../lib/addDefaults');
+var writeGeometryToGltf = require('../../lib/writeGeometryToGltf');
+var Cesium = require('cesium');
+var GeometryAttribute = Cesium.GeometryAttribute;
+var Geometry = Cesium.Geometry;
+
+var gltfPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf';
+var values = {'0': -0.5, '1': -0.5, '2': 0.5, '3': 0.5, '4': -0.5, '5': 0.5, '6': -0.5, '7': 0.5, '8': 0.5,
```
lilleyse: Just make this a regular array.

lib/readGltfGeometry.js
JoshuaStorm: I was also wondering if I should drop the `gltf` and go for `primitive` so: `convertPrimitiveToGeometry` and `convertGeometryToPrimitive`

or even just `primitiveToGeometry` and `geometryToPrimitive`

lib/readGltfGeometry.js
JoshuaStorm: `Cartesian3.packArray` does not seem to be exposed by Cesium. However, I will change it to at least use `Cartesian3.pack` in the loop instead, for the time being.

lilleyse: Yeah never mind, those functions aren't released in Cesium yet.

lib/readGltfGeometry.js
lilleyse: Those last two work for me.

JoshuaStorm:
Good for another look over now!

lib/geometryToPrimitive.js
```diff
@@ -0,0 +1,17 @@
...
+    var positionAccessor = gltf.accessors[positionId];
+
+    var indicesId = primitive.indices;
+    var indicesAccessor = gltf.accessors[indicesId];
+
+    var values = geometry.attributes.position.values;
+
+    writeAccessor(gltf, positionAccessor, values);
+    writeAccessor(gltf, indicesAccessor, geometry.indices);
+}
```
lilleyse: Can you support normals in here as well?

lib/primitiveToGeometry.js
```diff
@@ -0,0 +1,57 @@
...
+module.exports = primitiveToGeometry;
+
+function primitiveToGeometry(gltf, primitive) {
+    var positionId = primitive.attributes.POSITION;
+    var positionAccessor = gltf.accessors[positionId];
+    var componentsPerAttribute = numberOfComponentsForType(positionAccessor.type);
+    var values = readAccessor(gltf, positionAccessor);
+    
+    var packedValues = [];
+    packedValues.length = values.data.length * 3;
```
lilleyse: Change this to `var packedValues = new Array(values.data.length * 3);`
You could also factor out `values.data.length` into its own variable and use it here and in the for loop below.

lib/primitiveToGeometry.js
```diff
@@ -0,0 +1,57 @@
...
+
+    for (var i = 0; i < values.data.length; ++i) {
+        Cartesian3.pack(values.data[i], packedValues, i * 3);
+    }
+    
+    var position = new GeometryAttribute({
```
lilleyse: `positionAttribute` is a clearer name

lib/primitiveToGeometry.js
```diff
@@ -0,0 +1,57 @@
...
+        componentDatatype : ComponentDatatype.FLOAT,
+        componentsPerAttribute : componentsPerAttribute,
+        values : new Float64Array(packedValues)
+    });
+
+    var normalId = primitive.attributes.NORMAL;
+    var normalAccessor = gltf.accessors[normalId];
+    var normal = readAccessor(gltf, normalAccessor);
```
lilleyse: The normal attribute will not always exist, so you need to check for that.

lib/primitiveToGeometry.js
```diff
@@ -0,0 +1,57 @@
...
+    
+    var indicesId = primitive.indices;
+    var indicesAccessor = gltf.accessors[indicesId];
+    var indices = readAccessor(gltf, indicesAccessor);
+    var primitiveType = modeToPrimitiveType(primitive.mode);
+
+    var geometry = new Geometry({
+        attributes : {
+            position : position,
+            normal : new Float32Array(packedNormal)
```
lilleyse: This should also be a `GeometryAttribute` like positions.

specs/lib/primitiveToGeometrySpec.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+var primitiveToGeometry = require('../../lib/primitiveToGeometry');
+var readGltf = require('../../lib/readGltf');
+var addDefaults = require('../../lib/addDefaults');
+var readAccessor = require('../../lib/readAccessor');
+var Cesium = require('cesium');
+var Cartesian3 = Cesium.Cartesian3;
+
+
```
lilleyse: Remove extra line.

specs/lib/primitiveToGeometrySpec.js
```diff
@@ -0,0 +1,37 @@
...
+describe('primitiveToGeometry', function() {
+    it('returns a geometry', function(done) {
+        readGltf(gltfPath, function(gltf) {
+            addDefaults(gltf);
+            var primitive = gltf.meshes[Object.keys(gltf.meshes)[0]].primitives[0];
+            var geometry = primitiveToGeometry(gltf, primitive);
+            
+            expect(geometry.attributes.position.values).toEqual(positions);
+            expect(geometry.attributes.normal).toEqual(normals);
```
lilleyse: Same comment as above, normal should be an attribute so you'll want to check `normal.values` here.

specs/lib/primitiveToGeometrySpec.js
```diff
@@ -0,0 +1,37 @@
...
+            expect(geometry.indices).toEqual(indices);
+            expect(geometry.primitiveType).toEqual(primitiveType);
+            expect(geometry).toBeDefined();
```
lilleyse: You don't need this check anymore.

JoshuaStorm:
I think I've covered everything now for whenever you're ready to review.

lilleyse:
Looks good!

lib/modeToPrimitiveType.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+var Cesium = require('cesium');
+var PrimitiveType = Cesium.PrimitiveType;
+var WebGLConstants = Cesium.WebGLConstants;
+
+module.exports = modeToPrimitiveType;
+
+function modeToPrimitiveType(mode) {
```
pjcozzi: Do we need this function?  Aren't the `WebGLConstants` and `PrimitiveType` values the same?

See https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/PrimitiveType.js

lib/primitiveToGeometry.js
```diff
@@ -0,0 +1,66 @@
...
+var Cesium = require('cesium');
+var Geometry = Cesium.Geometry;
+var GeometryAttribute = Cesium.GeometryAttribute;
+var ComponentDatatype = Cesium.ComponentDatatype;
+var Cartesian3 = Cesium.Cartesian3;
+var defined = Cesium.defined;
+
+module.exports = primitiveToGeometry;
+
+function primitiveToGeometry(gltf, primitive) {
```
pjcozzi: Given that these are all fundamental graphics terms, throughout, I think we should be explicit about what is glTF and what is Cesium, e.g., `gltfPrimitiveToCesiumGeometry`.

lib/primitiveToGeometry.js
```diff
@@ -0,0 +1,66 @@
...
+            values: new Float32Array(packedNormals)
+        });
+    }
+    
+    var indicesId = primitive.indices;
+    var indicesAccessor = gltf.accessors[indicesId];
+    var indices = readAccessor(gltf, indicesAccessor);
+    var primitiveType = modeToPrimitiveType(primitive.mode);
+
+    var geometry = new Geometry({
```
pjcozzi: What about texture coordinates and other attributes?  Perhaps we should warn if they are dropped?

JoshuaStorm: I imagine there is no reason they couldn't be included (am I missing something?). I could add them now, but I just figured there was little reason to at this point since these methods would be private and only be used for specific portions of the pipeline that are based on positions/normals.

I'll go ahead and start working on these comments for a new PR

JoshuaStorm: Post-VS just relies on reordering indices. 

Pre-VS does re-order attributes, but as is this seems to work for textured models, but I imagine we are not gaining the full optimization since pre-VS is not reordering all attributes.

pjcozzi: > Pre-VS does re-order attributes, but as is this seems to work for textured models, but I imagine we are not gaining the full optimization since pre-VS is not reordering all attributes.

I'm not sure how this works at all.  Indices are per vertex, not per attribute so all attributes in a vertex need to be moved.

JoshuaStorm: > Indices are per vertex, not per attribute so all attributes in a vertex need to be moved.

Testing more models I'm starting to see this. I'll make sure I get a PR to resolve this before implementing any of this into the pipeline

PR #91, opened 2016-06-16
added image processing extras

likangning93:
In anticipation of ambient occlusion and other future image processing stages, I added extras for use with jimp (https://www.npmjs.com/package/jimp) in the form of a gltf-level scratch `Jimp` image and `Jimp` copies of all the existing images, stored as `jimpImage` in `extras._pipeline.`

Since this won't be used for every pipeline configuration, I tried to make it as optional as possible.

I also added relevant specs.

### DIFF
--- a/lib/encodeImages.js
+++ b/lib/encodeImages.js
@@ -1,0 +1,52 @@
+'use strict';
+var Jimp = require('jimp');
+
+module.exports = encodeImages;
+var async = require('async');
+
+function encodeImages(gltfWithExtras, callback) {
+    var images = gltfWithExtras.images;
+    var imageIDs = Object.keys(images);
+    async.each(imageIDs, function(imageID, asyncCallback){
+        if (images.hasOwnProperty(imageID)) {
+            var image = images[imageID];
+            var pipelineExtras = image.extras._pipeline;
+            // re-encode the jimp image here to a buffer that can be put in source
+            var mime;
+            switch(pipelineExtras.extension) {
+                case '.png':
+                    mime = Jimp.MIME_PNG;
+                    break;
+                case '.jpeg':
+                    mime = Jimp.MIME_JPEG;
+                    break;
+                case '.jpg':
+                    mime = Jimp.MIME_JPEG;
+                    break;
+                case '.bmp':
+                    mime = Jimp.MIME_BMP;
+                    break;
+            }
+
+            pipelineExtras.jimpImage.getBuffer(mime, function(err, imageBuffer) {
+                if (err) {
+                    throw err;
+                }
+                pipelineExtras.source = imageBuffer;
+                asyncCallback();
+            });
+        }
+    }, function(err) {
+        if (err) {
+            if (callback) {
+                callback(err);
+            }
+            else{
+                throw err;
+            }
+        }
+        else if (callback) {
+            callback(gltfWithExtras);
+        }
+    });
+}
--- a/lib/encodeImages.js
+++ b/lib/encodeImages.js
@@ -1,11 +1,11 @@
 'use strict';
+var async = require('async');
 var Jimp = require('jimp');
 
 module.exports = encodeImages;
-var async = require('async');
 
-function encodeImages(gltfWithExtras, callback) {
-    var images = gltfWithExtras.images;
+function encodeImages(gltf, callback) {
+    var images = gltf.images;
     var imageIDs = Object.keys(images);
     async.each(imageIDs, function(imageID, asyncCallback){
         if (images.hasOwnProperty(imageID)) {
@@ -44,15 +44,8 @@ function encodeImages(gltfWithExtras, callback) {
         }
     }, function(err) {
         if (err) {
-            if (callback) {
-                callback(err);
-            }
-            else{
-                throw err;
-            }
+            throw err;
         }
-        else if (callback) {
-            callback(gltfWithExtras);
-        }
+        callback(gltf);
     });
 }
--- a/lib/encodeImages.js
+++ b/lib/encodeImages.js
@@ -11,30 +11,36 @@ function encodeImages(gltfWithExtras, callback) {
         if (images.hasOwnProperty(imageID)) {
             var image = images[imageID];
             var pipelineExtras = image.extras._pipeline;
-            // re-encode the jimp image here to a buffer that can be put in source
-            var mime;
-            switch(pipelineExtras.extension) {
-                case '.png':
-                    mime = Jimp.MIME_PNG;
-                    break;
-                case '.jpeg':
-                    mime = Jimp.MIME_JPEG;
-                    break;
-                case '.jpg':
-                    mime = Jimp.MIME_JPEG;
-                    break;
-                case '.bmp':
-                    mime = Jimp.MIME_BMP;
-                    break;
-            }
 
-            pipelineExtras.jimpImage.getBuffer(mime, function(err, imageBuffer) {
-                if (err) {
-                    throw err;
+            if (pipelineExtras.imageChanged) {
+                // re-encode the jimp image here to a buffer that can be put in source
+                var mime;
+                switch(pipelineExtras.extension) {
+                    case '.png':
+                        mime = Jimp.MIME_PNG;
+                        break;
+                    case '.jpeg':
+                        mime = Jimp.MIME_JPEG;
+                        break;
+                    case '.jpg':
+                        mime = Jimp.MIME_JPEG;
+                        break;
+                    case '.bmp':
+                        mime = Jimp.MIME_BMP;
+                        break;
                 }
-                pipelineExtras.source = imageBuffer;
+
+                pipelineExtras.jimpImage.getBuffer(mime, function(err, imageBuffer) {
+                    if (err) {
+                        throw err;
+                    }
+                    pipelineExtras.source = imageBuffer;
+                    asyncCallback();
+                });
+            }
+            else {
                 asyncCallback();
-            });
+            }
         }
     }, function(err) {
         if (err) {
--- a/lib/gltfPipeline.js
+++ b/lib/gltfPipeline.js
@@ -26,12 +26,12 @@ module.exports = {
 
 function processJSON(gltf, options, callback) {
     addPipelineExtras(gltf);
-    loadGltfUris(gltf, options.basePath, function(err, gltf) {
+    loadGltfUris(gltf, options, function(err, gltf) {
         if (err) {
             throw err;
         }
         processJSONWithExtras(gltf, options, callback);
-    }, options.imageProcess);
+    });
 }
 
 function processJSONWithExtras(gltfWithExtras, options, callback) {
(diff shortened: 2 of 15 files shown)


lilleyse:
I'm a bit torn, because while it's faster to make the image processing an optional step, it makes the code a bit more confusing. It would be kind of nice if image processing was completely hidden from `gltfPipeline.js`, and instead it always decoded in `loadGltfUris` and encodes in `writeSource` (but only if the data was changed, this can be marked by a flag in the image's `extras` that is set by the AO stage/atlas stage etc).

An alternative is to continue to allow `readGltf` and `loadGltfUris` to take the `imageProcess` flag, but not have an `encodeImages` stage in `gltfPipeline`, instead always encode in `writeSource` using the flag idea above.

Let me know what you think, or if I'm overlooking something.

pjcozzi:
Can't this be handled with each stage declaring its dependencies?  See #75.

likangning93:
I really like option 2 for re-encoding. I also toyed with the idea of
changing 'loadGltfUris' to take an options parameter instead of 'basePath',
which might let us hide the image processing more. I'll give it a try.

likangning93:
Updated! As discussed offline, we decided to keep the `encodeImages` stage where it is for now. It seems like `writeSource` (which handles moving extras back into the json) is currently tied to `writeGltf` and not to `writeBinaryGltf`, and we may want to move it to come before the actual file writing options. When that happens we will move `encodeImages` over to that. `encodeImages` also now checks for an `imageChanged` flag before encoding.

I also updated `loadGltfUris` to take in an options parameter instead of just `basePath` and updated all relevant specs accordingly.

likangning93:
merged in master.

lib/encodeImages.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+var Jimp = require('jimp');
+
+module.exports = encodeImages;
+var async = require('async');
```
lilleyse: Fix the ordering here.

lib/encodeImages.js
```diff
@@ -0,0 +1,58 @@
...
+
+function encodeImages(gltfWithExtras, callback) {
```
lilleyse: Change `gltfWithExtras` to `gltf` to follow the other stages.

lib/encodeImages.js
```diff
@@ -0,0 +1,58 @@
...
+            if (callback) {
+                callback(err);
+            }
+            else{
+                throw err;
+            }
+        }
+        else if (callback) {
+            callback(gltfWithExtras);
+        }
```
lilleyse: Right now the callback to `encodeImage` expects a gltf object, not an error. Here you can just simplify to

```javascript
if (err) {
    throw err;
}
callback(gltf)
```

lib/loadGltfUris.js
```diff
@@ -8,18 +8,37 @@ var DeveloperError = Cesium.DeveloperError;
...
+                }
+                gltf.extras._pipeline.jimpScratch = image;
+                callback(err, gltf);
+            });
+        };
+    }
+
     async.each(['buffers', 'images', 'shaders'], function(name, asyncCallback) {
-        loadURI(gltf, basePath, name, asyncCallback);
+        loadURI(gltf, basePath, name, asyncCallback, imageProcess);
```
lilleyse: Usually the callback should be the last argument.

lib/readGltf.js
```diff
@@ -10,14 +10,13 @@ var DeveloperError = Cesium.DeveloperError;
 
 module.exports = readGltf;
 
-function readGltf(gltfPath, callback) {
+function readGltf(gltfPath, options, callback, imageProcess) {
```
lilleyse: You don't use `imageProcess` anymore right?

lib/readGltf.js
```diff
@@ -35,11 +34,11 @@ function readGltf(gltfPath, callback) {
...
             addPipelineExtras(gltf);
         }
-        loadGltfUris(gltf, filePath, function(err, gltf) {
+        loadGltfUris(gltf, options, function(err, gltf) {
             if (err) {
                 throw err;
             }
             callback(gltf);
-        });
+        }, imageProcess);
```
lilleyse: Same comment here.

lib/readGltf.js
```diff
@@ -10,14 +10,13 @@ var DeveloperError = Cesium.DeveloperError;
...
 module.exports = readGltf;
 
-function readGltf(gltfPath, callback) {
+function readGltf(gltfPath, options, callback, imageProcess) {
     if (!defined(gltfPath)) {
         throw new DeveloperError('Input path is undefined.');
     }
     
     var fileExtension = path.extname(gltfPath);
-    var filePath = path.dirname(gltfPath);
```
lilleyse: I know this change is now everywhere, but I think `options.basePath` should just be an optional argument for `readGltf`. Almost all of the time the `basePath` can be computed like this.

specs/lib/encodeImagesSpec.js
```diff
@@ -0,0 +1,85 @@
...
+var encodeImages = require('../../lib/encodeImages');
+var loadGltfUris = require('../../lib/loadGltfUris');
+var dataUriToBuffer = require('data-uri-to-buffer');
+var imageUri = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAICAIAAABLbSncAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAADsMAAA7DAcdvqGQAAADTSURBVBhXAcgAN/8B49/cCAQAyukB2vAB+vz/Ig7+QR0B+PwAAezj3LDfAqnZ/wMB/xwPBwUEAiMK91Uj/gLN6wGj1fs9IyEwGxoUCPotFQC75g7rASECvd/6KxwkVCHFUiTX9P4Xq9L8ZjUD+vWyAaHK+CUX+pujFBYRH1NR2vUAANrLX8zXyAJCHOWnsj3d7frR4+XLymz24Y6ZuKI7LGUE/vcBEAglAQTR/P/9nbmV8fUAYURiTjRqAeXi6AgFDMDWptPiwP///isdPEIsXvr8+Tj0Y5s8qCp8AAAAAElFTkSuQmCC';
+var basePath = './specs/data/boxTexturedUnoptimized/';
+var clone = require('clone');
+var Jimp = require('jimp');
+
+describe('encodeImages', function() {
+    var imageData;
```
lilleyse: This var isn't used.

specs/lib/encodeImagesSpec.js
```diff
@@ -0,0 +1,85 @@
...
+        loadGltfUris(gltfClone, options, function() {
+            var pipelineExtras0001 = gltfClone.images.Image0001.extras._pipeline;
+            var pipelineExtras0002 = gltfClone.images.Image0002.extras._pipeline;
+
+            encodeImages(gltfClone, function() {
+                expect(bufferEqual(pipelineExtras0001.source, imageBuffer)).toBe(true);
+                expect(bufferEqual(pipelineExtras0002.source, imageBuffer)).toBe(true);
+                done();
+            });
+        }, true);
```
lilleyse: Remove this last parameter.

specs/lib/encodeImagesSpec.js
```diff
@@ -0,0 +1,85 @@
...
+                    expect(image.bitmap.height).toEqual(10);
+                    expect(jimpImage001.bitmap.height).toEqual(10);
+                    Jimp.read(pipelineExtras0002.source, function(jimpErr, image) {
+                        expect(image.bitmap.height).toEqual(8);
+                        expect(jimpImage002.bitmap.height).toEqual(8);
+                        done();
+                    });
+                });
+            });
+        }, true);
```
lilleyse: And this.

specs/lib/gltfPipelineSpec.js
```diff
@@ -91,7 +101,10 @@ describe('gltfPipeline', function() {
         var spy = spyOn(fs, 'writeFile').and.callFake(function(file, data, callback) {
             callback();
         });
-        var options = { 'binary' : true };
+        var options = {
+            basePath: basePath,
+            'binary' : true
```
lilleyse: Remove quotation marks.

lib/readGltf.js
likangning93: It's still needed to signal that a `jimp` needs to get added.

lilleyse: I mean, isn't it now `options.imageProcess`?

likangning93: I kind of unpacked the options further up, should I just be doing `options.xxx`?

likangning93: Ah. I was looking at the wrong file.

likangning93:
Updated!

likangning93:
@pjcozzi integrating image processing into `runStages` might be a little messy since the encoding and decoding have to happen at the beginning and end of the pipeline and it seems like `runStages` handles the stages that are more synchronous.

I imagine we could add something before runStages that checks if the pipeline needs image processing extras, but that could get messy if we have to add anything else that wraps the pipeline like this.

specs/lib/cacheOptimizeIndicesSpec.js
```diff
@@ -5,6 +5,7 @@ var readGltf = require('../../lib/readGltf');
 var readAccessor = require('../../lib/readAccessor');
 var writeAccessor = require('../../lib/writeAccessor');
 
+var basePath = './specs/data/boxTexturedUnoptimized/';
```
lilleyse: There are a few lines like this that aren't used. Otherwise this PR looks ready to me.

likangning93:
ok. Updated!

lasalvavida:
> @pjcozzi integrating image processing into runStages might be a little messy since the encoding and decoding have to happen at the beginning and end of the pipeline and it seems like runStages handles the stages that are more synchronous.

We should adopt a promise style architecture for both synchronous and asynchronous stages; it would allow all stages to be run in the same fashion.

pjcozzi:
> We should adopt a promise style architecture for both synchronous and asynchronous stages; it would allow all stages to be run in the same fashion.

Yes, please add it to #1.

lib/gltfPipeline.js
```diff
@@ -54,11 +55,16 @@ function processJSONWithExtras(gltfWithExtras, options, callback) {
     if (options.quantize) {
         quantizeAttributes(gltfWithExtras);
     }
-    callback(gltfWithExtras);
+    if (options.imageProcess) {
+        encodeImages(gltfWithExtras, callback);
+    }
```
pjcozzi: `} else {`

package.json
```diff
@@ -38,6 +38,7 @@
     "data-uri-to-buffer": "0.0.4",
     "datauri": "^1.0.0-alpha.5",
     "image-size": "^0.5.0",
+    "jimp": "^0.2.24",
```
pjcozzi: Add this to the third-party section of [LICENSE.md](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/LICENSE.md).  In general, can we please update that section to match the libraries here in `dependencies`?

Issue #93, opened 2016-06-17
gltfPipeline.js not writing sources in all cases


lilleyse:
When the output of `gltfPipeline.js` is JSON and not a file, the `extras._pipeline.source` are not written back to the glTF. We should call `writeSource` in `processFile` and `processJSON`.

@JoshuaStorm Do you want to look into this?

JoshuaStorm:
Sure, I'll look into it now.

lilleyse:
Fixed in #94

PR #94, opened 2016-06-17
Fix bug where gltfPipeline was not writingSources in all cases 

JoshuaStorm:
Addresses #93 

`gltfPipeline` now writeSources properly when using `processJSON` or `processFile`. Had to move some logic around which caused a bit more copy-pasted code instead of just reusing the `processJSON` method as we were before, but it was only a few lines so it should be fine.

### DIFF
--- a/lib/gltfPipeline.js
+++ b/lib/gltfPipeline.js
@@ -34,7 +34,7 @@ function writeSources(gltf, outputPath, options, callback) {
     var embedImage = options.embedImage;
     var basePath = path.dirname(outputPath);
     async.each(['buffers', 'images', 'shaders'], function(name) {
-        writeSource(gltf, basePath, name, embed, embedImage, callback)
+        writeSource(gltf, basePath, name, embed, embedImage, callback);
     });
 }
 
--- a/lib/gltfPipeline.js
+++ b/lib/gltfPipeline.js
@@ -29,23 +29,26 @@ module.exports = {
     processFileToDisk : processFileToDisk
 };
 
-function writeSources(gltf, outputPath, options, callback) {
-    var embed = options.embed;
-    var embedImage = options.embedImage;
-    var basePath = path.dirname(outputPath);
-    async.each(['buffers', 'images', 'shaders'], function(name) {
-        writeSource(gltf, basePath, name, embed, embedImage, callback);
+function writeSources(gltf, callback) {
+    var embed = true;
+    var embedImage = true;
+    async.each(['buffers', 'images', 'shaders'], function(name, asyncCallback) {
+        writeSource(gltf, undefined, name, embed, embedImage, asyncCallback);
+    }, function() {
+        callback();
     });
 }
 
-function processJSON(gltf, outputPath, options, callback) {
+function processJSON(gltf, options, callback) {
     addPipelineExtras(gltf);
     loadGltfUris(gltf, options, function(err, gltf) {
         if (err) {
             throw err;
         }
         processJSONWithExtras(gltf, options, function(gltf) {
-            writeSources(gltf, outputPath, options, callback(gltf));
+            writeSources(gltf, function() {
+                callback(gltf);
+            });
         });
     });
 }
@@ -77,10 +80,12 @@ function processJSONWithExtras(gltfWithExtras, options, callback) {
     }
 }
 
-function processFile(inputPath, outputPath, options, callback) {
+function processFile(inputPath, options, callback) {
     readGltf(inputPath, options, function(gltf) {
         processJSONWithExtras(gltf, options, function(gltf) {
-            writeSources(gltf, outputPath, options, callback(gltf));
+            writeSources(gltf, function() {
+                callback(gltf);
+            });
         });
     });
 }
@@ -116,6 +121,7 @@ function processFileToDisk(inputPath, outputPath, options, callback) {
         });
     });
 }
+
 function printStats(stats) {
     process.stdout.write('Nodes removed: ' + stats.numberRemoved.nodes + '\n');
     process.stdout.write('Skins removed: ' + stats.numberRemoved.skins + '\n');
--- a/lib/gltfPipeline.js
+++ b/lib/gltfPipeline.js
@@ -1,4 +1,6 @@
 'use strict';
+var path = require('path');
+var async = require('async');
 var addDefaults = require('./addDefaults');
 var addPipelineExtras = require('./addPipelineExtras');
 var removeUnused = require('./removeUnused');
@@ -16,6 +18,7 @@ var loadGltfUris = require('./loadGltfUris');
 var quantizeAttributes = require('./quantizeAttributes');
 var cacheOptimizeIndices = require('./cacheOptimizeIndices');
 var encodeImages = require('./encodeImages');
+var writeSource = require('./writeSource');
 var Cesium = require('cesium');
 var defaultValue = Cesium.defaultValue;
 
@@ -26,13 +29,24 @@ module.exports = {
     processFileToDisk : processFileToDisk
 };
 
-function processJSON(gltf, options, callback) {
+function writeSources(gltf, outputPath, options, callback) {
+    var embed = options.embed;
+    var embedImage = options.embedImage;
+    var basePath = path.dirname(outputPath);
+    async.each(['buffers', 'images', 'shaders'], function(name) {
+        writeSource(gltf, basePath, name, embed, embedImage, callback)
+    });
+}
+
+function processJSON(gltf, outputPath, options, callback) {
     addPipelineExtras(gltf);
     loadGltfUris(gltf, options, function(err, gltf) {
         if (err) {
             throw err;
         }
-        processJSONWithExtras(gltf, options, callback);
+        processJSONWithExtras(gltf, options, function(gltf) {
+            writeSources(gltf, outputPath, options, callback(gltf));
+        });
     });
 }
 
@@ -63,10 +77,10 @@ function processJSONWithExtras(gltfWithExtras, options, callback) {
     }
 }
 
-function processFile(inputPath, options, callback) {
+function processFile(inputPath, outputPath, options, callback) {
     readGltf(inputPath, options, function(gltf) {
         processJSONWithExtras(gltf, options, function(gltf) {
-            callback(gltf);
+            writeSources(gltf, outputPath, options, callback(gltf));
         });
     });
 }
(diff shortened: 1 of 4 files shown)


specs/lib/gltfPipelineSpec.js
```diff
@@ -114,13 +114,47 @@ describe('gltfPipeline', function() {
...
+            callback();
+        });
+        var options = {
+            createDirectory : false,
+            embed : false
+        };
+        readGltf(gltfPath, options, function(gltf) {
+            processJSON(gltf, outputPath, options, function() {
+                process.nextTick(function() {
+                    expect(spy.calls.count()).toEqual(4);
```
lasalvavida: @JoshuaStorm Can you add a check in these tests making sure that source actually does get updated? Maybe make a change to it while it is in pipeline extras then decode the uri after and compare.

lasalvavida: Actually disregard this, it isn't necessary.

lasalvavida:
This looks good to me. @lilleyse is there a good use case to test this with to make sure it works properly?

lib/gltfPipeline.js
```diff
@@ -26,13 +29,24 @@ module.exports = {
...
 };
 
-function processJSON(gltf, options, callback) {
+function writeSources(gltf, outputPath, options, callback) {
+    var embed = options.embed;
+    var embedImage = options.embedImage;
+    var basePath = path.dirname(outputPath);
+    async.each(['buffers', 'images', 'shaders'], function(name) {
+        writeSource(gltf, basePath, name, embed, embedImage, callback);
+    });
```
lilleyse: You need a callback for when the async finishes, and instead you should send in `asyncCallback` to `writeSource`. Check out how `writeGltf` handles this.

lib/gltfPipeline.js
```diff
@@ -26,13 +29,24 @@ module.exports = {
...
+}
+
+function processJSON(gltf, outputPath, options, callback) {
```
lilleyse: `outputPath` doesn't really make sense as an argument for `processJSON` since all we want is JSON back. It looks like it was added so that `writeSources` would work, but instead `writeSources` could just always set `embed` to true.

lib/gltfPipeline.js
```diff
@@ -26,13 +29,24 @@ module.exports = {
...
     addPipelineExtras(gltf);
     loadGltfUris(gltf, options, function(err, gltf) {
         if (err) {
             throw err;
         }
-        processJSONWithExtras(gltf, options, callback);
+        processJSONWithExtras(gltf, options, function(gltf) {
+            writeSources(gltf, outputPath, options, callback(gltf));
```
lilleyse: The callbacks need to be fixed a bit. Normally you could just pass in `callback`, not `callback(gltf)`, but in this case `writeSource` doesn't supply the gltf back, so here you may want:

```javascript
processJSONWithExtras(gltf, options, function(gltf) {
    writeSources(gltf, outputPath, options, function() {
        callback(gltf);
    });
});
```

lib/gltfPipeline.js
```diff
@@ -63,10 +77,10 @@ function processJSONWithExtras(gltfWithExtras, options, callback) {
     }
 }
 
-function processFile(inputPath, options, callback) {
+function processFile(inputPath, outputPath, options, callback) {
     readGltf(inputPath, options, function(gltf) {
         processJSONWithExtras(gltf, options, function(gltf) {
-            callback(gltf);
+            writeSources(gltf, outputPath, options, callback(gltf));
```
lilleyse: Same note here.

lib/gltfPipeline.js
```diff
@@ -84,17 +98,24 @@ function writeFile(gltf, outputPath, options, callback) {
...
-    processJSON(gltf, options, function(gltf) {
-        writeFile(gltf, outputPath, options, callback);
+    addPipelineExtras(gltf);
+    loadGltfUris(gltf, options, function(err, gltf) {
+        if (err) {
+            throw err;
+        }
+        processJSONWithExtras(gltf, options, function(gltf) {
+            writeFile(gltf, outputPath, options, callback);
+        });
```
lilleyse: Why the change here and in `processFileToDisk`?

lib/gltfPipeline.js
```diff
@@ -84,17 +98,24 @@ function writeFile(gltf, outputPath, options, callback) {
...
 function processFileToDisk(inputPath, outputPath, options, callback) {
-    processFile(inputPath, options, function(gltf) {
-        writeFile(gltf, outputPath, options, callback);
+    readGltf(inputPath, options, function(gltf) {
+        processJSONWithExtras(gltf, options, function(gltf) {
+            writeFile(gltf, outputPath, options, callback);
+        });
     });
 }
-
```
lilleyse: Keep this line.

specs/lib/gltfPipelineSpec.js
```diff
@@ -114,13 +114,47 @@ describe('gltfPipeline', function() {
         });
     });
 
+    it('will write sources from JSON', function(done) {
```
lilleyse: This test and the other may not be needed based on the comments above.

specs/lib/gltfPipelineSpec.js
lilleyse: I think this is the right idea for testing this.

lilleyse:
> This looks good to me. @lilleyse is there a good use case to test this with to make sure it works properly?

Mainly just your other comment, running a gltf through `processJson` or `processFile` and seeing that its data changed.

lib/gltfPipeline.js
JoshuaStorm: I may have misunderstood how we wanted to resolve this issue. I thought we wanted the capability to output non-embedded files when using `processJSON` and `processFile`. If we are fine with them always being embedded, then that does simplify the solution (RE: your comment below - the only reason I refactored `processJSONToDisk` and `processFileToDisk` was to allow non-embedded).

lib/gltfPipeline.js
JoshuaStorm: I was wondering why the syntax I used above wouldn't work: Can a function in JS not be passed as an arguments with arguments in the function itself?

lib/gltfPipeline.js
lilleyse: Really its not a perfect solution either way, but I think its more user friendly to not write the files. Also I probably just didn't think through it enough when I created the issue.

lib/gltfPipeline.js
lilleyse: What will happen is the result of the `callback(gltf)` will be sent to `writeSources`, which could be anything and is probably not itself a callback.

JoshuaStorm:
Hit a bit of a snag, but this good for another look now.

likangning93:
Does this ensure that `writeSources` also happens for the data flow that ends up going through `writeBinaryGltf`? I think that was also an issue, but I might have missed something.

lib/gltfPipeline.js
lilleyse: This can change back to its original version now, right?

lib/writeSource.js
```diff
@@ -24,7 +24,7 @@ function writeSource(gltf, basePath, name, embed, embedImage, callback) {
                 var extension = object.extras._pipeline.extension;
 
                 // Write the source object as a data or file uri depending on the embed flag
-                if (embed && (embedImage || name !== 'images')) {
+                if (embed && (embedImage || name !== 'images') || basePath === undefined) {
```
lilleyse: Use `Cesium.defined` here.

specs/lib/gltfPipelineSpec.js
```diff
@@ -104,22 +104,46 @@ describe('gltfPipeline', function() {
             callback();
         });
         var options = {
-            createDirectory : false
+            createDirectory : false,
```
lilleyse: Remove comma.

specs/lib/gltfPipelineSpec.js
```diff
@@ -104,22 +104,46 @@ describe('gltfPipeline', function() {
...
+    });
+
+    it('will write sources from JSON', function(done) {
+        var options = {};
+        readGltf(gltfPath, options, function (gltf) {
+            var initialSource = gltf['buffers'].CesiumTexturedBoxTest.extras._pipeline.source;
+            processJSON(gltf, options, function () {
+                var finalSource = gltf['buffers'].CesiumTexturedBoxTest.extras._pipeline.source;
+                expect(initialSource).not.toEqual(finalSource);
+                done();
```
lilleyse: For these two tests it's more accurate to compare the data uris since we are checking that it wrote the data back to those.

specs/lib/gltfPipelineSpec.js
```diff
@@ -104,22 +104,46 @@ describe('gltfPipeline', function() {
...
+                expect(initialSource).not.toEqual(finalSource);
                 done();
             });
         });
     });
 
     it('will add image processing extras if this is a pipeline with image processing', function(done) {
         var options = {
-            imageProcess: true
+            imageProcess : true,
```
lilleyse: Remove comma.

lilleyse:
> Does this ensure that writeSources also happens for the data flow that ends up going through writeBinaryGltf? I think that was also an issue, but I might have missed something.

`writeBinaryGltf` and `writeFile` should both handle writing `extras._pipeline.source` back to the gltf, the changes in this PR only deal with the function that return a gltf as JSON.

JoshuaStorm:
Updated.

> writeBinaryGltf and writeFile should both handle writing extras._pipeline.source back to the gltf, the changes in this PR only deal with the function that return a gltf as JSON.

Is there currently an issue with `writeBinaryGltf` not having a writeSource step? I could look at that issue next.

lilleyse:
> Is there currently an issue with writeBinaryGltf not having a writeSource step? I could look at that issue next.

No issue with that, it's working fine for me.

lib/gltfPipeline.js
lilleyse: Never mind, these functions can't use `processJson` anymore due to the recent changes.

specs/lib/gltfPipelineSpec.js
```diff
@@ -107,19 +107,43 @@ describe('gltfPipeline', function() {
...
+        });
+    });
+
+    it('will write sources from JSON', function(done) {
+        var options = {};
+        readGltf(gltfPath, options, function (gltf) {
+            var initialUri = gltf['buffers'].CesiumTexturedBoxTest.uri;
+            processJSON(gltf, options, function () {
+                var finalUri = gltf['buffers'].CesiumTexturedBoxTest.uri;
+                expect(initialUri).not.toEqual(finalUri);
```
lilleyse: The initial uri will be a filepath to the .bin file, so the comparison will always pass. A quick fix for these 2 tests is to read from `gltfEmbeddedPath` instead of `gltfPath`.

JoshuaStorm:
These are good for review whenever you're ready

lilleyse:
Looks good to me!

PR #95, opened 2016-06-17
Write gltf changes

lilleyse:
A few random fixes and changes related to writing gltfs.

### DIFF
--- a/lib/getBinaryGltf.js
+++ b/lib/getBinaryGltf.js
@@ -30,7 +30,7 @@ function updateBinaryObject(gltf, pipelineExtras, name, state) {
                     state.currentBinaryView++;
                 }
 
-                var objectSource = object.extras._pipeline.source;
+                var objectSource = new Buffer(object.extras._pipeline.source);
                 var bufferViewId = 'binary_bufferView' + state.currentBinaryView;
                 KHR_binary_glTF.bufferView = bufferViewId; //Create bufferview
                 bufferViews[bufferViewId] = {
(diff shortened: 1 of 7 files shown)


lib/getBinaryGltf.js
```diff
@@ -30,7 +30,7 @@ function updateBinaryObject(gltf, pipelineExtras, name, state) {
                     state.currentBinaryView++;
                 }
 
-                var objectSource = object.extras._pipeline.source;
+                var objectSource = new Buffer(object.extras._pipeline.source);
```
lilleyse: `Buffer.concat` below didn't like concatting a shader string with a buffer, so I just always wrap the source in a Buffer.

lib/writeBinaryGltf.js
```diff
@@ -10,18 +10,25 @@ function writeBinaryGltf(gltf, outputPath, createDirectory, callback) {
...
     if (outputExtension !== '.glb') {
-        outputPath = path.basename(outputPath, outputExtension) + '.glb';
+        outputPath = path.join(path.dirname(outputPath), path.basename(outputPath) + '.glb');
     }
     // Create the output directory if specified
     if (createDirectory) {
         outputPath = path.join(path.dirname(outputPath), 'output', path.basename(outputPath));
-        mkdirp.sync(path.dirname(outputPath));
     }
+    mkdirp.sync(path.dirname(outputPath));
```
lilleyse: This change includes all the deleted specs. I thought it was weird that the pipeline won't export a gltf file unless `createDirectory` is true, especially since there are cases where you don't want the file to be put in an output folder.

lib/writeBinaryGltf.js
```diff
@@ -10,18 +10,25 @@ function writeBinaryGltf(gltf, outputPath, createDirectory, callback) {
...
 
-    var glb = getBinaryGltf(gltf, callback);
```
lilleyse: This callback was throwing off callbacks that expect an error in the first argument, but instead got the header, scene, body instead.

pjcozzi:
@lasalvavida do you want to review this?

lasalvavida:
These changes look good to me.

PR #96, opened 2016-06-17
Naive AO is ready for a first look!

likangning93:
Run:

`node ./bin/gltf-pipeline.js ./specs/data/ambientOcclusion/cube_over_ground.gltf --ao_diffuse=true --ao_samples=128 --ao_rayDepth=1.0 --ao_resolution=64 -o hello_AO.gltf`

This takes a couple seconds on my machine, but it's a really small model.

This version of naiveAO outputs a gltf with a separate image for each primitive and adds materials and textures as needed. It won't work with models that don't have any diffuse-textured materials, though, since it tries to find reference materials and textures to clone.

#### Tasks for this PR
* [x] improve memory efficiency by raytracing "on the fly" instead of precomputing and storing all the sample points
* [ ] improve sampling to avoid artifacts near the edges of triangles in UV space
* [ ] profile the code to get an idea of how to make speed improvements

#### Tasks for upcoming PRs
* [ ] evaluate speed benefits using a uniform grid for faster raytracing
* [ ] add an option to bake the AO to vertices or a separate texture and automatically modify relevant shader code
* [ ] evaluate the speed and memory benefits of moving the raytracing math to basic types instead of `Cartesian3`
* [ ] GPU-accelerate with CUDA or WebGL rasterization, e.g. [hemicube](http://the-witness.net/news/2010/09/hemicube-rendering-and-integration/)
   * [ ] Expand to prebaking IBL?
* [ ] UV preprocessing, checking, and cleanup: currently, this stage only works with primitives that already have normalized, non-overlapping UVs

### DIFF
--- a/bin/gltf-pipeline.js
+++ b/bin/gltf-pipeline.js
@@ -16,7 +16,13 @@ if (process.argv.length < 3 || defined(argv.h) || defined(argv.help)) {
         '  -b --binary, write binary glTF file.\n' +
         '  -s --separate, writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.\n' +
         '  -t --separateImage, write out separate textures, but embed geometry/animation data files, and shader files.\n' +
-        '  -q, quantize the attributes of this model.\n';
+        '  -q, quantize the attributes of this model.\n' +
+        '  --ao_diffuse, bake ambient occlusion into the diffuse texture.\n' +
+        '  --ao_separate, bake ambient occlusion into a separate texture and modify the shader to use it.\n' +
+        '  --ao_scene, specify which scene to bake AO for.\n' +
+        '  --ao_rayDepth, ray distance for raytraced ambient occlusion.\n' +
+        '  --ao_resolution, resolution along one dimension for each AO texture.\n' +
+        '  --ao_samples, sample count for ambient occlusion\n';
     process.stdout.write(help);
     return;
 }
@@ -31,6 +37,12 @@ var binary = defaultValue(defaultValue(argv.b, argv.binary), false);
 var separate = defaultValue(defaultValue(argv.s, argv.separate), false);
 var separateImage = defaultValue(defaultValue(argv.t, argv.separateImage), false);
 var quantize = defaultValue(defaultValue(argv.q, argv.quantize), false);
+
+var ao_diffuse = defaultValue(argv.ao_diffuse, false);
+var ao_scene = argv.ao_scene;
+var ao_rayDepth = defaultValue(argv.ao_rayDepth, 1.0);
+var ao_resolution = defaultValue(argv.ao_resolution, 128);
+var ao_samples = defaultValue(argv.ao_samples, 16);
 
 if (!defined(gltfPath)) {
     throw new DeveloperError('Input path is undefined.');
@@ -49,7 +61,13 @@ var options = {
     binary : binary,
     embed : !separate,
     embedImage : !separateImage,
-    quantize : quantize
+    quantize : quantize,
+    ao_diffuse : ao_diffuse,
+    ao_scene : ao_scene,
+    ao_rayDepth : ao_rayDepth,
+    ao_resolution : ao_resolution,
+    ao_samples : ao_samples,
+    imageProcess : ao_diffuse
 };
 
 processFileToDisk(gltfPath, outputPath, options);
--- a/bin/gltf-pipeline.js
+++ b/bin/gltf-pipeline.js
@@ -18,7 +18,6 @@ if (process.argv.length < 3 || defined(argv.h) || defined(argv.help)) {
         '  -t --separateImage, write out separate textures, but embed geometry/animation data files, and shader files.\n' +
         '  -q, quantize the attributes of this model.\n' +
         '  --ao_diffuse, bake ambient occlusion into the diffuse texture.\n' +
-        '  --ao_separate, bake ambient occlusion into a separate texture and modify the shader to use it.\n' +
         '  --ao_scene, specify which scene to bake AO for.\n' +
         '  --ao_rayDepth, ray distance for raytraced ambient occlusion.\n' +
         '  --ao_resolution, resolution along one dimension for each AO texture.\n' +
@@ -38,11 +37,13 @@ var separate = defaultValue(defaultValue(argv.s, argv.separate), false);
 var separateImage = defaultValue(defaultValue(argv.t, argv.separateImage), false);
 var quantize = defaultValue(defaultValue(argv.q, argv.quantize), false);
 
-var ao_diffuse = defaultValue(argv.ao_diffuse, false);
-var ao_scene = argv.ao_scene;
-var ao_rayDepth = defaultValue(argv.ao_rayDepth, 1.0);
-var ao_resolution = defaultValue(argv.ao_resolution, 128);
-var ao_samples = defaultValue(argv.ao_samples, 16);
+var aoOptions = {
+    runAO : defaultValue(argv.ao_diffuse, false),
+    scene : argv.ao_scene,
+    rayDepth : defaultValue(argv.ao_rayDepth, 1.0),
+    resolution : defaultValue(argv.ao_resolution, 128),
+    numberSamples : defaultValue(argv.ao_samples, 16)
+};
 
 if (!defined(gltfPath)) {
     throw new DeveloperError('Input path is undefined.');
@@ -62,12 +63,8 @@ var options = {
     embed : !separate,
     embedImage : !separateImage,
     quantize : quantize,
-    ao_diffuse : ao_diffuse,
-    ao_scene : ao_scene,
-    ao_rayDepth : ao_rayDepth,
-    ao_resolution : ao_resolution,
-    ao_samples : ao_samples,
-    imageProcess : ao_diffuse
+    aoOptions : aoOptions,
+    imageProcess : aoOptions.runAO
 };
 
 processFileToDisk(gltfPath, outputPath, options);
--- a/bin/gltf-pipeline.js
+++ b/bin/gltf-pipeline.js
@@ -37,13 +37,15 @@ var separate = defaultValue(defaultValue(argv.s, argv.separate), false);
 var separateImage = defaultValue(defaultValue(argv.t, argv.separateImage), false);
 var quantize = defaultValue(defaultValue(argv.q, argv.quantize), false);
 
-var aoOptions = {
-    runAO : defaultValue(argv.ao_diffuse, false),
-    scene : argv.ao_scene,
-    rayDepth : defaultValue(argv.ao_rayDepth, 1.0),
-    resolution : defaultValue(argv.ao_resolution, 128),
-    numberSamples : defaultValue(argv.ao_samples, 16)
-};
+var aoOptions;
+if (argv.ao_diffuse) {
+    aoOptions = {
+        scene : argv.ao_scene,
+        rayDepth : defaultValue(argv.ao_rayDepth, 1.0),
+        resolution : defaultValue(argv.ao_resolution, 128),
+        numberSamples : defaultValue(argv.ao_samples, 16)
+    };
+}
 
 if (!defined(gltfPath)) {
     throw new DeveloperError('Input path is undefined.');
@@ -64,7 +66,7 @@ var options = {
     embedImage : !separateImage,
     quantize : quantize,
     aoOptions : aoOptions,
-    imageProcess : aoOptions.runAO
+    imageProcess : defined(aoOptions)
 };
 
 processFileToDisk(gltfPath, outputPath, options);
--- a/lib/bakeAmbientOcclusion.js
+++ b/lib/bakeAmbientOcclusion.js
@@ -10,6 +10,7 @@ var Cartesian3 = Cesium.Cartesian3;
 var Matrix3 = Cesium.Matrix3;
 var Matrix4 = Cesium.Matrix4;
 var Quaternion = Cesium.Quaternion;
+var WebGLConstants = Cesium.WebGLConstants;
 var Jimp = require('jimp');
 
 var Ray = Cesium.Ray;
@@ -33,29 +34,28 @@ var quaternionScratch = new Quaternion();
 var matrix3Scratch = new Matrix3();
 
 function bakeAmbientOcclusion(gltf, options) {
-    // requires each mesh to occur only once in the scene
+    // Requires each mesh to occur only once in the scene
     var sceneID = defaultValue(options.scene, gltf.scene);
     if (!defined(sceneID)) {
         sceneID = Object.keys(gltf.scenes)[0];
     }
     var scene = gltf.scenes[sceneID];
 
-    // generate triangle soup
+    // Generate triangle soup
     var raytracerScene = generateRaytracerScene(gltf, scene, options);
 
+    // Raytrace to a new texture for each primitive
     var parameters = {
-        meshes: gltf.meshes,
         raytracerScene: raytracerScene,
-        resolution: options.resolution,
-        primitiveFunction: raytracePrimitiveTexels
+        resolution: options.resolution
     };
+    NodeHelpers.forEachPrimitiveInScene(gltf, scene, raytracePrimitiveTexels, parameters);
 
-    // raytrace to a new texture for each primitive
-    NodeHelpers.forEachPrimitiveInScene(gltf, scene, parameters);
-
-    ////////// add to the gltf //////////
+    // Add to the gltf
     bakeToDiffuse(gltf, scene, options.resolution, raytracerScene);
 }
+
+////////// adding to the gltf //////////
 
 function bakeToDiffuse(gltf, scene, resolution, raytracerScene) {
     // find material with a diffuse texture parameter to clone as needed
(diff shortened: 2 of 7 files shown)


likangning93:
Ongoing things that we'll have to deal with:
- speed - the naive raytrace gets much slower as the number of triangles increases
- memory efficiency - models like the milk truck crash the pipeline for me even with about 6 gb of memory allocated. This is a major problem.

likangning93:
Please refrain from taking a look too deeply before next week. I might try to make some memory efficiency improvements before Monday, since frankly right now this is pretty alarming. I think the gist of the problem is in how I handle the texel points that the algorithm casts rays from. Currently at a high level the stage raytraces like this:
- generate a soup of world-space triangles
- rasterize every triangle in UV space to generate a massive cache of world-space `texelPoints`
- for every `texelPoint`, cast out the requisite number of feeler rays

I think a big part of the problem is in the `texelPoint` cache - this thing must be huge, and it's not even necessary, mostly just an artifact from reading too much about hemicube irradiance baking.

There may also be some space efficiency problems in storing the triangles using `cartesian3` but I'll tackle the `texelPoint` cache problem first.

pjcozzi:
This is exciting news!

Can you please:
* [ ] Post example with/without AO screenshots
* [ ] Add a tasklist to the top of this PR with items that you want to do for this PR, e.g., memory efficiency, and a tasklist with roadmap items for after this PR, e.g., uniform grid, optimal AO map size, etc.

likangning93:
`cube_over_ground.gltf"` with and without AO:
with_and_without_ao

pjcozzi:
Cool, thanks @likangning93!

pjcozzi:
>  improve sampling to avoid artifacts near the edges of triangles

I don't know that it is needed, but this could be of interest: http://jcgt.org/published/0002/01/05/

> Tasks for upcoming PRs

Also add
* [ ] GPU-accelerate with CUDA or WebGL rasterization, e.g. [hemicube](http://the-witness.net/news/2010/09/hemicube-rendering-and-integration/)
   * [ ] Expand to prebaking IBL?

These could also be 565 projects.

likangning93:
So with the `texelPoint` cache removed, running the AO pipeline with a bigger model like `CesiumMilkTruck.gltf` no longer causes dramatic and unsustainable increases in memory use, and the process can get to raytracing without crashing. I also want to profile if removing this cache also helps with speed, but I'll need to find a suitable model since just `time node ...` doesn't show any improvements for tiny models like the one I have in the specs folder. I'm going to run this on `CesiumMilkTruck.gltf` overnight just to see what happens.

likangning93:
I think something is wrong with how `CesiumMilkTruck.gltf`'s UVs are being read. I'll take another look tomorrow.

likangning93:
The problem with the milk truck model is that its UVs aren't normalized, they're much bigger and get normalized in the shader using a hardcoded scaling parameter. The AO code currently does a pixel march over each triangle's AABB in UV space, so if the UV coordinates aren't normalized the march is 1) wrong and 2) seems to find far more "pixels" than it should. I'll throw an error for now if  the pixel march code detects non-normalized UVs.

lasalvavida:
@likangning93 You should be able to [normalize](https://cesiumjs.org/Cesium/Build/Documentation/Cartesian3.html#.normalize) them pretty easily if that makes sense.

likangning93:
Thanks Rob!

pjcozzi:
Can we also add a stage to un-encode oct-encoded normals?

lasalvavida:
> Can we also add a stage to un-encode oct-encoded normals?

I can certainly decode the normal data easily enough, but un-modifying the shader is a pretty complex problem.

pjcozzi:
> I can certainly decode the normal data easily enough, but un-modifying the shader is a pretty complex problem.

Ah, right, I would not bother now since we have control of the pipeline and can just oct-encode after AO.

likangning93:
Some changes I need to make along with the first round of comments:
- raytracing and baking shouldn't be done for every primitive in the gltf, it should be done for every primitive in the scene
- remove some unnecessary memory allocations and move some more expensive math out of loops

lib/bakeAmbientOcclusion.js
```diff
@@ -0,0 +1,616 @@
...
+            NodeHelpers.depthFirstTraversal(allNodes[rootNodeName], allNodes, addNodeToSoup, parameters);
+        }
+    }
+
+    return raytracerScene;
+}
+
+function addNodeToSoup(parameters, node) {
+    var transform = node.extras._pipeline.flatTransform;
+    var mesheIDs = node.meshes;
```
lasalvavida: Nitpick: should probably be meshIDs

lib/bakeAmbientOcclusion.js
```diff
@@ -0,0 +1,616 @@
...
+    for (var meshIndex in mesheIDs) {
+        if (mesheIDs.hasOwnProperty(meshIndex)) {
+            var meshID = mesheIDs[meshIndex];
```
lasalvavida: This is an array not a dictionary; iterate by length, not using `for in`

lib/bakeAmbientOcclusion.js
```diff
@@ -0,0 +1,616 @@
...
+            var meshData = parameters.meshes[meshID];
+            var primitives = meshData.primitives;
+            for (var primitiveID in primitives) {
+                if (primitives.hasOwnProperty(primitiveID)) {
+                    processPrimitive(parameters, meshID + '_' + primitiveID, primitives[primitiveID], transform);
```
lasalvavida: Same comment.

lib/bakeAmbientOcclusion.js
```diff
@@ -0,0 +1,616 @@
...
+    var aoBuffer = {
+        resolution: resolution,
+        samples: new Array(resolution * resolution).fill(0.0),
+        count: new Array(resolution * resolution).fill(0.0)
+    };
+
+    raytracerScene.aoBufferByPrimitive[meshPrimitiveID] = aoBuffer;
+
+    // Read each triangle's Cartesian3s using the index buffer
+    for (var i = 0; i < numTriangles; i++) {
```
lasalvavida: Not all primitives are triangles, so you need to check the primitive's mode before you can make this assumption.

lib/bakeAmbientOcclusion.js
likangning93: Updated, changed in `NodeHelpers.js`

likangning93:
Updated! I think this is actually ready for a deeper look now.

bin/gltf-pipeline.js
```diff
@@ -49,7 +61,13 @@ var options = {
     binary : binary,
     embed : !separate,
     embedImage : !separateImage,
-    quantize : quantize
+    quantize : quantize,
+    ao_diffuse : ao_diffuse,
+    ao_scene : ao_scene,
+    ao_rayDepth : ao_rayDepth,
+    ao_resolution : ao_resolution,
+    ao_samples : ao_samples,
```
lilleyse: Perhaps pack all the AO properties into an object. Also this makes it less verbose in `gltfPipeline`.

bin/gltf-pipeline.js
```diff
@@ -49,7 +61,13 @@ var options = {
...
+    imageProcess : ao_diffuse
```
lilleyse: Does `imageProcess` still need to be true if `--ao-separate` is set? Maybe this type of thing will be handled later.

lib/bakeAmbientOcclusion.js
```diff
@@ -0,0 +1,586 @@
...
+var Matrix4 = Cesium.Matrix4;
+var Quaternion = Cesium.Quaternion;
+var WebGLConstants = Cesium.WebGLConstants;
+var Jimp = require('jimp');
+
+var Ray = Cesium.Ray;
+var baryCentricCoordinates = Cesium.barycentricCoordinates;
+var readAccessor = require('./readAccessor');
+var NodeHelpers = require('./NodeHelpers');
+
```
lilleyse: The includes could be organized just a little more.

lib/bakeAmbientOcclusion.js
```diff
@@ -0,0 +1,586 @@
...
+    var exampleMaterialID;
+    var exampleTextureID;
+    var exampleImageID;
+
+    for (var materialID in materials) {
+        if (materials.hasOwnProperty(materialID)) {
+            var material = materials[materialID];
+            if (defined(material.values) && defined(material.values.diffuse)) {
+                if (typeof material.values.diffuse === "string") {
```
lilleyse: Use single quotes.

lib/bakeAmbientOcclusion.js
```diff
@@ -0,0 +1,586 @@
...
+
+function postProcessAO(aoBuffer, dataResolution, goalResolution, jimpAO) {
+    // Copy the data over to the jimp
+    var value = 0.0;
+    jimpAO.resize(dataResolution, dataResolution);
+    for (var x = 0; x < dataResolution; x++) {
+        for (var y = 0; y < dataResolution; y++) {
+            var dataIdx = dataResolution * y + x;
+            value = 255.0 * (aoBuffer.samples[dataIdx] / aoBuffer.count[dataIdx]);
+            jimpAO.bitmap.data[(dataResolution * y + x) * 4 + 3] = value;
```
lilleyse: Replace `dataResolution * y + x` with `dataIdx`?

lib/bakeAmbientOcclusion.js
```diff
@@ -0,0 +1,586 @@
...
+    var raytracerScene = parameters.raytracerScene;
+    var bufferDataByAccessor = raytracerScene.bufferDataByAccessor;
+    var indices = bufferDataByAccessor[primitive.indices].data;
+    var positions = bufferDataByAccessor[primitive.attributes.POSITION].data;
+    var normals = bufferDataByAccessor[primitive.attributes.NORMAL].data;
+    var uvs = bufferDataByAccessor[primitive.attributes.TEXCOORD_0].data;
+    var numTriangles = indices.length / 3;
+    var transform = node.extras._pipeline.flatTransform;
+    var inverse = new Matrix4();
+    inverse = Matrix4.inverse(transform, inverse);
```
lilleyse: `inverse` isn't used right now.

specs/lib/bakeAmbientOcclusionSpec.js
```diff
@@ -0,0 +1,525 @@
...
+                            "_pipeline": {}
+                        }
+                    }
+                ]
+            }
+        }
+    };
+
+
+
```
lilleyse: Extra spacing here.

specs/lib/bakeAmbientOcclusionSpec.js
```diff
@@ -0,0 +1,525 @@
...
+                addPipelineExtras(boxOverGroundGltf);
+                loadGltfUris(boxOverGroundGltf, loadGltfUriOptions, function(err, gltf) {
+                    if (err) {
+                        throw err;
+                    }
+                    done();
+                });
+            }
+        });
+    });
```
lilleyse: Should only call `done()` when both are done.

specs/lib/bakeAmbientOcclusionSpec.js
```diff
@@ -0,0 +1,525 @@
...
+    var point3 = new Cartesian3(0.0, 1.0, 0.0);
+
+    var tetrahedron = [
+        {positions: [point0, point1, point2]},
+        {positions: [point0, point1, point3]},
+        {positions: [point1, point2, point3]},
+        {positions: [point2, point0, point3]}
+    ];
+
+    function testContainmentAndFitCartesian3(min, max, cartesian3s) {
```
lilleyse: Is this supposed to be used for something right now?

specs/lib/bakeAmbientOcclusionSpec.js
```diff
@@ -0,0 +1,525 @@
...
+        }
+        return true;
+    }
+
+    it('correctly processes a basic 2-triangle square primitive', function() {
+        var scene = testGltf.scenes[testGltf.scene];
+        var options = {
+            "rayDepth" : 0.1,
+            "resolution" : 10
+        };
```
lilleyse: Don't use quotes for these.

specs/lib/bakeAmbientOcclusionSpec.js
```diff
@@ -0,0 +1,525 @@
...
+                buffer: aoBuffer
+            });
+        }
+
+        for (i = 0; i < 6; i++) {
+          var texel = texelPoints[i];
+          bakeAmbientOcclusion.computeAmbientOcclusionAt(
+            texel.position, texel.normal, 16, 4,
+            tetrahedron, 0.001, 10.0, aoBuffer, texel.index);
+        }
```
lilleyse: Spacing is off in this block.

specs/lib/bakeAmbientOcclusionSpec.js
```diff
@@ -0,0 +1,525 @@
...
+        var samples = aoBuffer.samples;
+        var counts = aoBuffer.count;
+        for (i = 0; i < 6; i++) {
+            expect(samples[i]).toEqual(16.0);
+            expect(counts[i]).toEqual(16);
+        }
+    });
+
+    it('generates various levels of occlusion for samples in the mouth of an open tetrahedron', function() {
+        var openTetrahedron = [tetrahedron[1], tetrahedron[2], tetrahedron[3]];
```
lilleyse: This is not being used, but is it supposed to?

specs/lib/bakeAmbientOcclusionSpec.js
```diff
@@ -0,0 +1,525 @@
...
+                buffer: aoBuffer
+            }
+        ];
+
+        for (var i = 0; i < 3; i++) {
+          var texel = texelPoints[i];
+          bakeAmbientOcclusion.computeAmbientOcclusionAt(
+            texel.position, texel.normal, 16, 4,
+            tetrahedron, 0.001, 10.0, aoBuffer, texel.index);
+        }
```
lilleyse: Spacing off here too.

lilleyse:
Pretty minor comments here, I like the organization of the AO stage, it was easy to follow. As you already noted in your comments, it would be good to handle the case of creating ao textures even if no diffuse textures are present. Great start here!

bin/gltf-pipeline.js
likangning93: `imageProcess` is needed at both the beginning and the end of the pipeline and will probably be needed in the pre-write stage, so I think it would be better to start using it now even though it's basically the same as the AO flag.

specs/lib/bakeAmbientOcclusionSpec.js
likangning93: Ah. This is left over from when I was generating that `texelPoint` cache, it's not used anymore.

likangning93:
Updated!

lib/gltfPipeline.js
```diff
@@ -53,17 +53,7 @@ function processJSONWithExtras(gltfWithExtras, options, callback) {
...
-    if (options.ao_diffuse) {
-        var aoOptions = {
-            scene: options.ao_scene,
-            rayDepth: options.ao_rayDepth,
-            numberSamples: options.ao_samples,
-            resolution: options.ao_resolution
-        };
-        bakeAmbientOcclusion(gltfWithExtras, aoOptions);
-    }
+    bakeAmbientOcclusion(gltfWithExtras, options.aoOptions);
```
lilleyse: I think it's a little clearer and more in-line with the other stages if we check if we can run the ao stage before going into the function.

```javascript
if (defined(options.aoOptions) && options.aoOptions.runAO) {
    bakeAmbientOcclusion(gltfWithExtras, options.aoOptions);
}
```

And possibly just remove `runAO` altogether. As long as the ao options exist, it should be able to run the stage. This would require some edits in `gltf-pipeline.js`.

specs/lib/bakeAmbientOcclusionSpec.js
```diff
@@ -0,0 +1,501 @@
...
+                boxGltf = JSON.parse(data);
+                addPipelineExtras(boxGltf);
+                loadGltfUris(boxGltf, loadGltfUriOptions, function(err, gltf) {
+                    if (err) {
+                        throw err;
+                    }
+                    loadModel2();
+                });
+            }
+        });
```
lilleyse: Hm just noticed, this area could be simplified by using `readGltf` if that works.

lib/gltfPipeline.js
likangning93: Oh that makes sense... no idea why someone would have aoOptions but then not run the AO.

likangning93:
Updated!

lilleyse:
All good to me. Are there any other items on the checklist you plan on finishing before this is ready? They can all be separate PRs if you prefer.

In terms of the ao workflow, we should have a dedicated ao branch and all ao-related pull requests would go into that branch. When ao is "done", we'll merge the ao branch into master.

So once this is ready, can you close this PR, create an `ao-master` branch, then create a new PR with all these commits that targets `ao-master` (and just reference back to this PR so we know where all the code came from).

likangning93:
Last notes before closing:
- I'm putting conservative rasterization aside for a few days to try baking per-vertex, since not all our models will have texture coordinates or textures
- here are some results from baking:
[models.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/326448/models.zip)
    - the cube model in here demonstrates the need for conservative rasterization, you can see that no AO was baked to texels on some edges
    - as a low bar for efficiency, the building model (normals mangled by blender) took 139 minutes, 47.361 seconds to bake at `128` samples per texel, `512x512` AO texels wit ha ray depth of `2.0`. This model is 5521 triangles

likangning93:
moved to https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/102

PR #97, opened 2016-06-20
Add support for attribute compression

lasalvavida:
Currently this relies on Cesium resolving the built-in shader functions, so it won't work in other glTF engines. 

Changing the `includeBuiltIns` flag to true will fully resolve the shader source in the pipeline, making it compatible with other engines, but then if you use the model in Cesium you will get shader errors since it tries to add the functions in and they already exist. 
@pjcozzi Is that a bug in Cesium?


lasalvavida:
For example, here's a milk truck model that crashes Cesium.

[CesiumMilkTruck-optimized.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/323711/CesiumMilkTruck-optimized.zip)

pjcozzi:
Does this still need updates to README.md and integration into the pipeline?

lib/compressTextureCoordinates.js
```diff
@@ -0,0 +1,172 @@
...
+var findAccessorMinMax = require('./findAccessorMinMax');
+var getAccessorByteStride = require('./getAccessorByteStride');
+var numberOfComponentsForType = require('./numberOfComponentsForType');
+var readBufferComponentType = require('./readBufferComponentType');
+var writeBufferComponentType = require('./writeBufferComponentType');
+var uninterleaveAndPackBuffers = require('./uninterleaveAndPackBuffers');
+
+module.exports = compressTextureCoordinates;
+
+function compressTextureCoordinates(gltf) {
```
pjcozzi: Include some doc here like this: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/AttributeCompression.js#L233

lasalvavida:
@pjcozzi I didn't want to add pipeline integration until the concerns in my original comment were addressed.

lib/compressTextureCoordinates.js
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+    uninterleaveAndPackBuffers(gltf);
+}
+
+function canCompress(accessor) {
+    return accessor.type === 'VEC2';
+}
+
+function getTextureCoordinateAccessors(gltf) {
```
pjcozzi: Can this and `getNormalAccessors` be replaced with a generic function?  For example [removeObject](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/removeObject.js) is used instead of custom functions.

pjcozzi: Actually, it could be more elegant and faster (by avoiding the intermediate data structure) to have this new function take a function that is called per accessor when it is discovered.  In general, I think helpers like this that help walk the glTF data structures could simplify a lot of code in gltf-pipeline.

lib/octEncodeNormals.js
```diff
@@ -0,0 +1,175 @@
...
+    var techniques = gltf.techniques;
+    var technique = techniques[techniqueId];
+    var attributes = technique.attributes;
+    var parameters = technique.parameters;
+    var normalParameterName;
+    for (var parameterName in parameters) {
+        if (parameters.hasOwnProperty(parameterName)) {
+            var parameter = parameters[parameterName];
+            var semantic = parameter.semantic;
+            if (semantic === 'NORMAL') {
```
pjcozzi: I don't think we can reuse this semantic since the spec defines `NORMAL` as floating-point vec3.

https://github.com/KhronosGroup/glTF/blob/master/specification/README.md#parametersemantic

`_NORMAL` or `_ENCODED_NORMAL` or `_OCT_ENDCODED_NORMAL` would be OK.

lib/compressTextureCoordinates.js
```diff
@@ -0,0 +1,172 @@
...
+    var technique = techniques[techniqueId];
+    var attributes = technique.attributes;
+    var parameters = technique.parameters;
+    var texCoordParameterName;
+    for (var parameterName in parameters) {
+        if (parameters.hasOwnProperty(parameterName)) {
+            var parameter = parameters[parameterName];
+            var semantic = parameter.semantic;
+            if (defined(semantic)) {
+                if (semantic.indexOf('TEXCOORD') === 0) {
```
pjcozzi: Same comment.

lib/octEncodeNormals.js
lasalvavida: Agreed. It's probably worth noting that I'm pretty sure the AGI model converter leaves the semantic as NORMAL when it applies oct-encoding so that might be something for @tfili to look at.

pjcozzi:
> Changing the includeBuiltIns flag to true will fully resolve the shader source in the pipeline, making it compatible with other engines, but then if you use the model in Cesium you will get shader errors since it tries to add the functions in and they already exist. 

We'll want this to be `true` (and not even exposed).  Perhaps just post-process the generated shaders so the specific `czm_` functions are renamed to `gltf_`.  If we modify Cesium to, for example, not add a `czm_` function that is already declared, assets that go through this pipeline stage would require that version of Cesium.

lib/octEncodeNormals.js
pjcozzi: Yes.  It doesn't break the Cesium implementation, but doesn't strictly follow the spec.

lasalvavida: Actually, I'm not seeing where it explicitly states that in the spec. All of the uniforms have explicit types, but the attributes do not.

pjcozzi: Ah, yes, you are right.  Let's just leave it as is; I don't think we need to tighten up the spec here because, for example, `NORMAL` for 2D would be vec2.

lasalvavida:
This should be ready for another look. I've tested all different combinations on buggy and milktruck. They all work, and I've also included a file size comparison

| Model | Original | Baseline | Quantized | Oct-Encoded | Compressed | Oct-Encoded/Quantized | Oct-Encoded/Compressed | Oct-Encoded/Quantized/Compressed
| :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| Buggy.gltf | 10,402 KB | 6,050 KB | 3,911 KB | 4,203 KB | 6,050 KB | 3,136 KB | 6,050 KB | 4,203 KB | 4,203 KB |
| CesiumMilkTruck.gltf | 1,371 KB | 1,366 KB | 1,308 KB | 1,329 KB | 1,351 KB | 1,293 KB | 1,314 KB | 1,293 KB |

lasalvavida:
@pjcozzi If you get a chance, could you take a look at this?

pjcozzi:
The numbers look good! https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/97#issuecomment-228421144

lib/getAccessorsForSemantic.js
```diff
@@ -0,0 +1,35 @@
...
+ * found that uses a particular semantic.
+ *
+ * @param {Object} gltf The glTF hierarchy
+ * @param {String} semantic Matches against the beginning of the semantic string for the primitives' attributes
+ * @param {Function} callback This gets called when a match is found with the arguments (gltf, primitive, matchedSemantic, accessorId)
+ * @param {Function} callback This gets called when the iteration completes
+ */
+function getAccessorsForSemantic(gltf, semantic, callback, done) {
+    var meshes = gltf.meshes;
+    for (var meshId in meshes) {
```
pjcozzi: Do you want to cherry-pick `NodeHelpers.forEachPrimitiveInScene` from @likangning93's AO branch and use it here?  There are several places in gltf-pipeline that could use it now.

lasalvavida: Sounds good to me

lasalvavida: As discussed offline, this didn't really make sense to do in practice, but I am going to change it so that is uses promises instead of a done callback.

lasalvavida:
@pjcozzi This is updated to use promises and is ready

pjcozzi:
Perhaps not related to this branch, but any thoughts on this:
```
$ node ./bin/gltf-pipeline.js Cesium_Air.glb -b -o Cesium_Air.glb
buffer.js:622
    throw new RangeError('index out of range');
    ^

RangeError: index out of range
    at checkOffset (buffer.js:622:11)
    at Buffer.readUInt16LE (buffer.js:668:5)
    at readBufferComponentType (c:\source\gltf-pipeline\lib\readBufferComponentT
ype.js:16:27)
    at findAccessorMinMax (c:\source\gltf-pipeline\lib\findAccessorMinMax.js:33:
33)
    at accessorDefaults (c:\source\gltf-pipeline\lib\addDefaults.js:24:30)
    at addDefaults (c:\source\gltf-pipeline\lib\addDefaults.js:561:5)
    at processJSONWithExtras (c:\source\gltf-pipeline\lib\gltfPipeline.js:61:5)
    at c:\source\gltf-pipeline\lib\gltfPipeline.js:141:9
    at c:\source\gltf-pipeline\lib\readGltf.js:45:13
    at c:\source\gltf-pipeline\lib\loadGltfUris.js:36:9
```
I got this for all .glb files I tried.

pjcozzi:
> I got this for all .glb files I tried.

At least for the aircraft, the .gltf file worked OK.

lasalvavida:
@pjcozzi Im pretty sure that's probably #106

pjcozzi:
These two commands produced the same models:
```
node ./bin/gltf-pipeline.js Cesium_Air.gltf -q -b -o Cesium_Air_q.glb
node ./bin/gltf-pipeline.js Cesium_Air.gltf -q -n -c -b -o Cesium_Air_qnc.glb
```
Shouldn't the second model be smaller due to normal and texture coordinate compression?

pjcozzi:
> @pjcozzi Im pretty sure that's probably #106

Are all the glTF sample models up-to-date?  If not, can we update them?

Can you also open a PR into Cesium to update the models included there?

lasalvavida:
@pjcozzi, can you link to the Cesium_Air.gltf you're using for that test?

lasalvavida:
> Are all the glTF sample models up-to-date? If not, can we update them?

All of the sample models should be fine; the current converter doesn't output models with this error anymore, and I did regenerate them all when I was fixing schema errors. I will double check though.

It seems like the cesium models are the ones that need updating.

pjcozzi:
Here's Cesium Man with `-q`:

image

```
node ./bin/gltf-pipeline.js Cesium_Man.glb -q -b -o Cesium_Man_q.glb
```

pjcozzi:
> Here's Cesium Man with -q:

The Monster model, which also has skins, doesn't have this problem.

pjcozzi:
> @pjcozzi, can you link to the Cesium_Air.gltf you're using for that test?

https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Apps/SampleData/models/CesiumAir/Cesium_Air.gltf

> It seems like the cesium models are the ones that need updating.

Yes, that is what I suspect.

lasalvavida:
> These two commands produced the same models:

>> @pjcozzi, can you link to the Cesium_Air.gltf you're using for that test?

>https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Apps/SampleData/models/CesiumAir/Cesium_Air.gltf

That model already has oct-encoded normals which is why there's no change.

pjcozzi:
> That model already has oct-encoded normals which is why there's no change.

Ah, right!  Good news.

lasalvavida:
@pjcozzi The cesium man issue should be fixed in #119.

pjcozzi:
> @pjcozzi The cesium man issue should be fixed in #119.

Looks good, I merged it into this branch.

Let me know when you finish https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/97#issuecomment-229447996 and I'll do final testing.

lasalvavida:
A few updates:

I've made it so that `JOINT` semantics are excluded from quantization by default. There's too many potential issues with trying to make sure that decoding produces the correct `int` value. Even with the #119 fix, RiggedFigure is broken. I've opened #120 for a better way to compress `JOINT`.

Quantized BrainStem breaks the model inspector, but it works fine in Cesium; I'll open an issue in the model inspector, but it doesn't affect this pull request or the cesium one.

Compressing Texture coordinates on the BoxTextured model produces this:
boxtextured-min

I'm still trying to figure that out.

pjcozzi:
Thanks for the careful testing!

I merged https://github.com/AnalyticalGraphicsInc/cesium/pull/3891 and will keep this open until the BoxTextured model is fixed.

lasalvavida:
@pjcozzi The issue seems to be that the texture coordinate compression only works on texture coordinates bounded in the floating point range [0.0, 1.0]. However, BoxTextured uses the repeating part of its texture to wrap around the rest of the box; breaking the compression. I put in a check on the accessor min and max for this.

pjcozzi:
@lasalvavida I think that is the best approach.

@tfili is this what you did for COLLADA?

lasalvavida:
@pjcozzi I've tested all of the glTF models and all of the Cesium ones. This should be ready for whatever final testing you want to do.

pjcozzi:
Nice!

PR #98, opened 2016-06-20
Added function for CesiumRTC

lasalvavida:
@lilleyse

### DIFF
--- a/lib/addCesiumRTC.js
+++ b/lib/addCesiumRTC.js
@@ -1,6 +1,7 @@
 'use strict';
 var Cesium = require('cesium');
 var Cartesian3 = Cesium.Cartesian3;
+var DeveloperError = Cesium.DeveloperError;
 var Ellipsoid = Cesium.Ellipsoid;
 var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
@@ -26,8 +27,14 @@ function addCesiumRTC(gltf, options) {
     if (defined(options)) {
         var positionArray = [];
         var ellipsoid = defaultValue(options.ellipsoid, Ellipsoid.WGS84);
-        var position = defaultValue(options.position,
-            Cartesian3.fromRadians(options.longitude, options.latitude, options.height, ellipsoid));
+        var position = options.position;
+        if (!defined(position)) {
+            if (defined(options.longitude) && defined(options.latitude && defined(options.height))) {
+                position = Cartesian3.fromRadians(options.longitude, options.latitude, options.height, ellipsoid);
+            } else {
+                throw new DeveloperError('Either a position or lat/long/height must be provided');
+            }
+        }
         Cartesian3.pack(position, positionArray);
         var extensions = gltf.extensions;
         if (!defined(extensions)) {
--- a/lib/addCesiumRTC.js
+++ b/lib/addCesiumRTC.js
@@ -2,41 +2,58 @@
 var Cesium = require('cesium');
 var Cartesian3 = Cesium.Cartesian3;
 var Ellipsoid = Cesium.Ellipsoid;
+var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
 
 var addExtensionsUsed = require('./addExtensionsUsed');
 
 module.exports = addCesiumRTC;
 
+/**
+ * Positions the model relative to a center point using the CESIUM_RTC extension.
+ * Either specify options.position directly or calculate it from longitude, latitude and height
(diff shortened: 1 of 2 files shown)


lilleyse:
Should the positions be transformed relative to the center point too?

lib/addCesiumRTC.js
```diff
@@ -0,0 +1,42 @@
...
+var Ellipsoid = Cesium.Ellipsoid;
+var defined = Cesium.defined;
+
+var addExtensionsUsed = require('./addExtensionsUsed');
+
+module.exports = addCesiumRTC;
+
+function addCesiumRTC(gltf, options) {
+    var positionArray = [];
+    var position = Cartesian3.fromRadians(options.longitude, options.latitude, options.height, Ellipsoid.WGS84);
```
lilleyse: Seems easier to just compute the center from the position attribute min/max. It's a little more restrictive, but this makes it easier to add this stage to `gltfPipeline`.

lilleyse:
As we talked about offline, I had a misunderstanding of the RTC extension. My model is specified in wgs84 already so I need to do some preprocessing before using this stage.

lib/addCesiumRTC.js
```diff
@@ -0,0 +1,59 @@
...
+ * @param {Number} [options.latitude] The latitude to map onto the ellipsoid.
+ * @param {Number} [options.height] Distance from the ellipsoid's surface.
+ * @param {Ellipsoid} [options.ellipsoid=Ellipsoid.WGS84] The ellipsoid to use for lat-long mapping.
+ */
+function addCesiumRTC(gltf, options) {
+    if (defined(options)) {
+        var positionArray = [];
+        var ellipsoid = defaultValue(options.ellipsoid, Ellipsoid.WGS84);
+        var position = defaultValue(options.position,
+            Cartesian3.fromRadians(options.longitude, options.latitude, options.height, ellipsoid));
```
lilleyse: If `options.position` is defined it will still call `Cartesian3.fromRadians` but with undefined values.

lasalvavida: Yup, good catch

lasalvavida:
Updated

lilleyse:
Looks good.

lib/addCesiumRTC.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Positions the model relative to a center point using the CESIUM_RTC extension.
+ * Either specify options.position directly or calculate it from longitude, latitude and height
+ * along with an ellipsoid (WGS84 if not specified).
+ *
+ * @param {Object} gltf A javascript object holding a glTF hierarchy.
+ * @param {Object} [options] Defines the relationship with the center position
+ * @param {Cartesian3} [options.position] Places the model relative to a particular position.
+ * @param {Number} [options.longitude] The longitude to map onto the ellipsoid.
```
pjcozzi: We are usually explicit that longitude/latitude is in radians and height is in meters.

lib/addCesiumRTC.js
```diff
@@ -0,0 +1,66 @@
...
+function addCesiumRTC(gltf, options) {
+    if (defined(options)) {
+        var positionArray = [];
+        var ellipsoid = defaultValue(options.ellipsoid, Ellipsoid.WGS84);
+        var position = options.position;
+        if (!defined(position)) {
+            if (defined(options.longitude) && defined(options.latitude && defined(options.height))) {
+                position = Cartesian3.fromRadians(options.longitude, options.latitude, options.height, ellipsoid);
+            } else {
+                throw new DeveloperError('Either a position or lat/long/height must be provided');
```
pjcozzi: Don't abbreviate the parameters here.

pjcozzi:
> As we talked about offline, I had a misunderstanding of the RTC extension. My model is specified in wgs84 already so I need to do some preprocessing before using this stage.

Exactly, but I don't know that this stage is useful in practice here.  Although it is appropriate for our 3D Tiles pipeline; it is too niche for the glTF Pipeline, and just increases its complexity with a stage that most users will not know what to do with.

I suggest that we remove it.

lasalvavida:
> I suggest that we remove it.

I'm not sure that's the best avenue. This is a glTF extension and I think that having this utility could be useful. Maybe just make the documentation a little clearer on what it is for?

pjcozzi:
This stage alone is not useful since it doesn't update the vertex data.  It's not a priority for us to add that to gltf-pipeline now and most users would not need it, so I would remove this partial stage and reconsider adding a complete stage if users need it in practice.

PR #100, opened 2016-06-20
gltfPrimitiveToCesiumGeometry fixes

JoshuaStorm:
Now handles texture coordinate attribute. See: #90 

As of now, it is only handling one set of texture coordinates. I was unable to find a gltf model with multiple texture coordinate sets for one primitive. Do we have any?

Addressed some other small comments from #90 as well.

I'm also wondering where I should be putting the pre and post VS optimizations within the pipeline. I imagine the pre can replace a portion of what is now labeled the `cacheOptimization` step. When do we want to perform the post VS optimization?


lasalvavida:
@JoshuaStorm One of your tests is failing here and you do have a few jsHint errors.

JoshuaStorm:
@lasalvavida Addressed

specs/lib/gltfPrimitiveToCesiumGeometrySpec.js
```diff
@@ -11,20 +11,24 @@ var positions = new Float64Array([ -0.5, -0.5, 0.5, 0.5, -0.5, 0.5, -0.5, 0.5, 0
 var normals = new Float32Array([ 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 0, 1, 0, 0, 1, 
     0, 0, 1, 0, 0, 1, 0, 0, -1, 0, 0, -1, 0, 0, -1, 0, 0, -1, 0, -1, 0, 0, -1, 0, 0, -1, 0, 0, -1, 0, 0, 0, 0, -1, 0, 0,
     -1, 0, 0, -1, 0, 0, -1 ]);
+var sts = new Float32Array([ 6, 0, 5, 0, 6, 0.9999998807907104, 5, 0.9999998807907104, 4, 0, 5, 0, 4, 1, 5, 1,
+    2, 0, 1, 0, 2, 1, 1, 1, 3, 0, 4, 0, 3, 1, 4, 1, 3, 0, 2, 0, 3, 1, 2, 1, 0, 0, 0, 0.9999998807907104, 1, 0, 1,
+    0.9999998807907104 ]);
 var indices = new Uint32Array([ 0, 1, 2, 3, 2, 1, 4, 5, 6, 7, 6, 5, 8, 9, 10, 11, 10, 9, 12, 13, 14, 15, 14, 13, 16, 17,
     18, 19, 18, 17, 20, 21, 22, 23, 22, 21 ]);
 var primitiveType = 4;
```
lilleyse: I should have brought this up before, but instead can you extract these values from the gltf like in line 71-92 of `cesiumGeometryToGltfPrimitiveSpec`, just so if the test model changes, the test will still pass.

specs/lib/cesiumGeometryToGltfPrimitiveSpec.js
```diff
@@ -49,30 +58,48 @@ describe('geometryToPrimitive', function() {
...
+           var packedNormals = new Array(positionLength * 3);
+           for (var i = 0; i < normalLength; ++i) {
+               Cartesian3.pack(newNormals[i], packedNormals, i * 3);
+           }
+
+           var textureLength = newCoordinates.length;
+           var packedCoordinates = new Array(textureLength * 2);
+           for (var i = 0; i < positionLength; ++i) {
+               Cartesian2.pack(newCoordinates[i], packedCoordinates, i * 2);
+           }
```
lilleyse: Add `var i;` at the top to avoid multiple declarations.

lilleyse:
That's all, looks good to me otherwise.

pjcozzi:
> As of now, it is only handling one set of texture coordinates

Are we sure that we can't just handle all attributes in a generic manner?  For the ones where the semantics are know, we can map glTF to Cesium, but for the others, I _thought_ Cesium would allow generic attributes.

JoshuaStorm:
> Are we sure that we can't just handle all attributes in a generic manner? For the ones where the semantics are know, we can map glTF to Cesium

I am not sure what the benefit would be to handling all attributes in a generic manner since only the attributes of known semantics will be changed when we use the geometry pipeline optimizations and our only intention is to use these geometries in intermediate steps of the gltf pipeline. Wouldn't we just be reading attributes to rewrite them exactly the same later?

pjcozzi:
> I am not sure what the benefit would be to handling all attributes in a generic manner since only the attributes of known semantics will be changed when we use the geometry pipeline optimizations and our only intention is to use these geometries in intermediate steps of the gltf pipeline. Wouldn't we just be reading attributes to rewrite them exactly the same later?

For example, look at [GeometryPipeline.reorderForPreVertexCache](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/GeometryPipeline.js#L369-L393).  It moves each attribute.

JoshuaStorm:
Updated. I expect this PR will still need a bit of work after this refactoring, let me know if you think I'm on the right track for what you had in mind @pjcozzi.

JoshuaStorm:
Actually found a bug while doing further testing. **Not** ready for review, my bad.

JoshuaStorm:
As of now, `gltfPrimitiveToCesiumGeometry` and `cesiumGeometryToGltfPrimitive` seem to work as expected. However, when adding a step in the pipeline to apply `Cesium.GeometryPipeline` steps, certain primitives seem to have issues.

When putting a geometry through `GeometryPipeline.reorderForPostVertexCache` only and rewriting it to the gltf, the model seems to work fine.

When putting a geometry through `GeometryPipeline.reorderForPostVertexCache` only and rewriting it ot the gltf, the model seems fine.

However, when putting a gltf through both, it seems to mangle some primitives (see below). If I flip the order of `GeometryPipeline.reorderForPostVertexCache` and `GeometryPipeline.reorderForPostVertexCache`, other primitives are mangled.

I've tried a few things and double checked that `gltfPrimitiveToCesiumGeometry` and `cesiumGeometryToGltfPrimitive` work as expected but haven't found what may be causing this issue.

cacheoptimize

JoshuaStorm:
Update to the last comment: I believe the bugs are all in my use of the `GeometryPipeline` functions and not any of the code in this PR.

pjcozzi:
@lilleyse can you please help @JoshuaStorm with the Cesium API if needed?

JoshuaStorm:
Had to rebase due to pulling and accidentally pushing from one of Rob's PRs for testing.

As Sean and I discussed offline: The scope of this PR is decreasing to just implementing the `GeometryPipeline.reorderForPostVertexCache` since `reorderForPreVertexCache` is creating issues that we need to think over more before implementing fully (which we might put on the backburner for now). 

A brief description of the issue: the pre vertex cache optimization shifts around both the primitive's attributes and primitive's indices. This is fine when only one primitive uses each attributes accessor, but in many cases, when the attributes accessors are shared (when multiple primitives share vertices), the model will get mangled as the shared vertices will be shifted each time a primitive accesses it resulting in indices referring to the wrong vertices.

We have a couple ideas for possible solutions, but none of which are perfect and involve quite a bit of work, so Sean and I figured that would be appropriate for another PR.

lasalvavida:
@JoshuaStorm, I split off the `createAccessorUsageTables` utility for you as discussed offline in my buffer-fixes branch from #105. I would cache primitives by their index accessor, then iterate over the tables. If a table only has a single index accessor, you are free to perform cache optimization on that index accessor and all the related attribute accessors in the table.

lilleyse:
@JoshuaStorm Do you want to try the pre-vertex cache with the new code before I look into this?

JoshuaStorm:
Yeah, I'll give it a shot and let you know when this is good for review

JoshuaStorm:
Good for review now!

Handling selective pre vertex cache optimization now! Works on all models I have tested. In order to do selective pre vertex cache optimization, I map the accessors provided by `createAccessorUsageTables` to primitives which requires a check for equality between the accessors. I use a naive O(N * M) comparison check in order to do this, but this seems fine since N and M will be fairly small. For example, the entire `cacheOptimization` step only adds 4583ms to the `gltfPipeline` runtime when processing the `BrainStem.gltf` model.

lib/cacheOptimization.js
```diff
@@ -0,0 +1,84 @@
...
+    // Iterate through each usage table
+    for (var j = 0; j < usageTables.length; j++) {
+        var indexAccessors = usageTables[j].indexAccessors;
+        var attributeAccessors = usageTables[j].attributeAccessors;
+        var keys = Object.keys(indexAccessors);
+        // If a set of attributes accessors is only used by one indices accessor
+        if (keys.length === 1) {
+            var indicesId = keys[0];
+            // look through to see if an independent primitive exists for those attributes
+            primitive = getAccessorsPrimitive(gltf, attributeAccessors, indicesId);
```
lasalvavida: Doing this search every time is really inefficient. Iterate over the primitives and put them into a dicitionary of arrays where the keys are the index accessor ids before you go through the usage tables. Then here, you can just look up the primitives in ~O(1).

lib/createAccessorUsageTables.js
```diff
@@ -129,7 +129,7 @@ function createAccessorUsageTables(gltf) {
                         for (attribute in attributes) {
                             if (attributes.hasOwnProperty(attribute)) {
                                 attributeAccessorId = attributes[attribute];
-                                table.attributeAccessors[attributeAccessorId] = attribute;
+                                table.attributeAccessors[attributeAccessorId] = true;
```
lasalvavida: Thank you for correcting that

lib/cacheOptimization.js
JoshuaStorm: Good call, I didn't think of that for some reason. 
Thanks!

lib/gltfPrimitiveToCesiumGeometry.js
```diff
@@ -0,0 +1,101 @@
...
+    switch(attributeId) {
+        case 'POSITION':
+            attributeName = 'position';
+            finalValues = new Float64Array(packed);
+            break;
+        case 'NORMAL':
+            attributeName = 'normal';
+            finalValues = new Float32Array(packed);
+            break;
+        case 'TEXCOORD_0':
```
lasalvavida: If you take a look at the glTF spec, valid attributes are of the form `SEMANTIC` or `SEMANTIC_#`. You need to check if the attributeId starts with a semantic, equality won't work for all cases. For example, this could be `TEXCOORD` instead of `TEXCOORD_0` in a valid glTF file.

JoshuaStorm: Since Cesium Geometries only directly handle one `position`, `normal`, and `st`, I believe it would be sufficient to just handle equality of `SEMANTIC` or `SEMANTIC_0` and just allow the default option for generic Geometry attributes. 

Is there any circumstance where the first `SEMANTIC_#` # wouldn't be 0?

JoshuaStorm:
Updated, still seeking thoughts in regards to @lasalvavida's comment:

> If you take a look at the glTF spec, valid attributes are of the form `SEMANTIC` or `SEMANTIC_#`. You need to check if the attributeId starts with a semantic, equality won't work for all cases. For example, this could be `TEXCOORD` instead of `TEXCOORD_0` in a valid glTF file.


Since Cesium Geometries only directly handle one `position`, `normal`, and `st`, I believe it would be sufficient to just handle equality of `SEMANTIC` or `SEMANTIC_0` and just allow the default option for generic Geometry attributes in all other cases.

Is there any circumstance where the first SEMANTIC_# # wouldn't be 0?

lib/cacheOptimization.js
```diff
@@ -0,0 +1,107 @@
...
+            var primitives = mesh.primitives;
+            var primitivesLength = primitives.length;
+            for (var i = 0; i < primitivesLength; i++) {
+                var primitive = primitives[i];
+                var indicesId = primitive.indices;
+
+                var primitivesOfIndicesId;
+                if (!dictionary[indicesId]) {
+                    primitivesOfIndicesId = [];
+
```
lasalvavida: extra whitespace

lib/cacheOptimization.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return dictionary;
+}
+
+function getIndependentPrimitive(dictionary, indicesId, attributeAccessors) {
+    var primitivesOfIndicesId = dictionary[indicesId];
+    for (var primitiveId in primitivesOfIndicesId) {
+        if (primitivesOfIndicesId.hasOwnProperty(primitiveId)) {
+            var primitive = primitivesOfIndicesId[primitiveId];
+            var accessors = primitive.attributes;
```
lasalvavida: Watch your naming here. Since the attributes are `SEMANTIC` : `accessorId` it would be better to call this `attributes` and iterate over a variable named `semantic` or `attributeSemantic`.

lib/cacheOptimization.js
```diff
@@ -0,0 +1,107 @@
...
+
+            var equalCount = 0;
+            // Naive check to see if this primitive's attributes match
+            for (var accessorId in accessors) {
+                if (accessors.hasOwnProperty(accessorId)) {
+                    for (var attributeAccessor in attributeAccessors) {
+                        if (attributeAccessor === accessors[accessorId]) {
+                            equalCount++;
+                        }
+                    }
```
lasalvavida: This second loop can be avoided. Just do something like this:
```javascript
for (var attributeSemantic in attributes) {
    if (attributes.hasOwnProperty(attributeSemantic)) {
        var accessorId = attributes[attributeSemantic];
        if (defined(attributeAccessors[accessorId])) {
            ...
        }
    }
}
```

lasalvavida:
@JoshuaStorm Take a look at the [specification](https://github.com/KhronosGroup/glTF/tree/master/specification#meshes). 

Specifically:
> Valid attribute semantic property names include POSITION, NORMAL, TEXCOORD, COLOR, JOINT, and WEIGHT. Attribute semantic property names can be of the form [semantic]_[set_index], e.g., TEXCOORD_0, TEXCOORD_1, etc.

The short answer to your question is yes. Maybe there are two primitives with different materials, and the author decided to make them use TEXCOORD_0 and TEXCOORD_1 to make it easier to tell them apart. That is still valid glTF. There's also nothing stopping someone from making it TEXCOORD_25, or POSITION_3 if they want to.

Just doing a check for whether the listed semantic starts with the one you're looking for should be fine for all cases.

```javascript
if (attributeSemantic.indexOf(semantic) === 0) {
   ...
}
```

JoshuaStorm:
Updated, I hope I didn't miss anything.

I am actually handling semantics a bit differently now to avoid possible issues with different semantic naming schemes. Currently when going from gltf primitive <-> Cesium Geometry, I am just using the first semantic to map accordingly and just having the rest be generic attributes. This works fine for now, but I may need to modify it slightly once I change #82 to use the GeometryPipeline.

lib/cacheOptimization.js
```diff
@@ -0,0 +1,107 @@
...
+            var primitive = primitivesOfIndicesId[primitiveId];
+            var attributes = primitive.attributes;
+
+            var equalCount = 0;
+            // Naive check to see if this primitive's attributes match
+            for (var semantic in attributes) {
+                if (attributes.hasOwnProperty(semantic)) {
+                    var accessorId = attributes[semantic];
+                    if (defined(attributeAccessors[accessorId])) {
+                        if (attributeAccessors[semantic] === attributes[semantic]) {
```
lasalvavida: What is this testing for? `attributeAccessors` doesn't have semantic keys. I think just the defined check should be sufficient here.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -0,0 +1,55 @@
...
+var writeAccessor = require('./writeAccessor');
+
+module.exports = cesiumGeometryToGltfPrimitive;
+
+// Get the first gltf attribute semantic of that type
+function getFirstAttributeSemantic(primitive, semantic) {
+    var attributes = primitive.attributes;
+    for (var attributeSemantic in attributes) {
+        if (attributes.hasOwnProperty(attributeSemantic)) {
+            if (attributeSemantic.replace(/[0-9]|_/g, '') === semantic) {
```
lasalvavida: This does work, but using `indexOf` is preferable since it doesn't create a new string.

lib/cacheOptimization.js
JoshuaStorm: Must've accidentally ctrl-zed that back at some pooint, oops.

lib/cesiumGeometryToGltfPrimitive.js
JoshuaStorm: Is there a similar function that removes the index without creating a new string?

JoshuaStorm:
Aside from a final review, this should be good to go.

lib/cacheOptimization.js
```diff
@@ -0,0 +1,105 @@
...
+var GeometryPipeline = Cesium.GeometryPipeline;
+var defined = Cesium.defined;
+
+module.exports = cacheOptimization;
+
+// Helper method to map accessor collections from the usageTable to an independent primitive (if there is one)
+function createIndicesToAttributeDictionary(gltf) {
+    var dictionary = {};
+    var meshes = gltf.meshes;
+    for (var meshId in meshes) {
```
pjcozzi: Can this use `NodeHelpers.forEachPrimitiveInScene`?

lib/cacheOptimization.js
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+}
+
+function cacheOptimization(gltf, cacheSize) {
+    // perform post vertex cache optimization
+    var primitive;
+    var geometry;
+    var meshes = gltf.meshes;
+    for (var meshId in meshes) {
```
pjcozzi: Same comment.

pjcozzi:
@lasalvavida for your final testing, could you please do some performance testing with a model with a large number of vertices and no textures (like Happy Buddha or the Dragon from http://graphics.stanford.edu/data/3Dscanrep/)?

In Cesium, turn off the globe, stars, and sky, (or use the Model Inspector) and see what the FPS is with and without cache optimization.  You may need to load several of the same model to notice.  A 10% win would be good.  It is probably less.

For numbers from the original paper, see http://gfx.cs.princeton.edu/gfx/pubs/Sander_2007_%3ETR/index.php

JoshuaStorm:
@pjcozzi:
To test I loaded 100 Dragon models in Cesium.

Before pipeline:
~20FPS

Through the pipeline without cache optimization: 
~21FPS

Through the pipeline with cache optimization:
~44FPS

>A 10% win would be good. It is probably less.

I've checked triple checked that I'm not making a mistake here,  but the performance improvements seem better than we expected.

lasalvavida:
@JoshuaStorm Good job, that's awesome! I will try to duplicate soon.

pjcozzi:
Despite this being a heavily vertex-bound test case, I still did not expect something this good!

lasalvavida:
@JoshuaStorm When I run the chinese dragon model through the pipeline with and without cache optimization enabled, I see no change. Could you please attach the models you used for your test here?

JoshuaStorm:
Odd, here's the model I used:

[chinesedragon.gltf.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/341958/chinesedragon.gltf.zip)

lib/cacheOptimization.js
JoshuaStorm: Is this currently in master? I cannot seem to find this function.

lasalvavida: It is not; it is in the ao branch. I think forego using it for now since it is async and doesn't have a way to notify when it completes currently.

lasalvavida:
@JoshuaStorm I meant can you attach the ones with and without cache optimization?

JoshuaStorm:
@lasalvavida 

Sorry about that: below is the cache optimized, the original input, and the original input through the pipeline without the cache optimization.

[Archive.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/342097/Archive.zip)

lasalvavida:
@JoshuaStorm The model I was using had split primitives, which is why there was no change. That is the desired behavior. I tested with the models you provided and got the same results, great job!

PR #105, opened 2016-06-22
A few buffer logic fixes

lasalvavida:
These fixes come from integrating pipeline into 3D Tiles processing and having some conversion problems.

Removes an unnecessary 'if' wrapping when copying data chunks from removeUnusedVertices. It shouldn't ever be copying data past the end of the segment, so if it is we want it to fail because something else is wrong.

uninterleaveAndPackBuffers had an issue with naming collision in bufferViews, and this should fix that.

@lilleyse Could you take a look at this when you get a chance and maybe run a few random sample models through it to make sure they don't break?


lasalvavida:
I have some test failures that need fixing first.

lasalvavida:
That should fix the tests, this is ready for a look.

lib/uninterleaveAndPackBuffers.js
```diff
@@ -98,7 +121,9 @@ function packAccessors(gltf, accessorsToPack, sourceBuffer, packBuffer, bufferVi
         var offset = byteOffset;
         var count = accessor.count;
         accessor.byteStride = 0;
-        accessor.bufferView = bufferViewId;
+        // Store as temp for now, finalize later
+        var extras = getPipelineExtras(accessor);
```
lilleyse: Is this comment something you will address before this PR is done?

lasalvavida: This refers to the fact that it temporarily stores the new bufferViewId in extras and then permanently assigns it at the end to avoid naming conflicts. It isn't a TODO.

lasalvavida:
@lilleyse This is ready for a look. It refactors removeUnusedIndices to cover all of the possible edge cases and adds a test for the cross-dependency issue I was having. 

The tilesets I was testing on process correctly now, and the BrainStem model still has the same result from merging the duplicates; all the old tests still pass.

I added uint32 index support for the sponza model as well.

JoshuaStorm:
@lasalvavida Build is currently failing

lasalvavida:
@JoshuaStorm Yup, sorry about that, should be fixed now

lib/byteLengthForComponentType.js
```diff
@@ -25,6 +28,7 @@ function byteLengthForComponentType(componentType) {
         case WebGLConstants.UNSIGNED_SHORT:
             return 2;
         case WebGLConstants.FLOAT:
+        case WebGLConstants.UNSIGNED_INT:
```
lilleyse: Also update `readBufferComponentType` and `writeBufferComponentType` with the new type.

lasalvavida: Updated

lib/readBufferComponentType.js
```diff
@@ -14,6 +14,8 @@ function readBufferComponentType(buffer, componentType, byteOffset) {
             return buffer.readInt16LE(byteOffset);
         case WebGLConstants.UNSIGNED_SHORT:
             return buffer.readUInt16LE(byteOffset);
+        case WebGLConstants.UNSIGNED_INT:
+            return buffer.readUInt32LE(value, byteOffset);
```
lilleyse: Remove `value'.

lilleyse:
I was about to comment that the sponza model wasn't loading in Cesium, but the recent change fixes that, so that's good!

lib/readBufferComponentType.js
lasalvavida: Whoops, thanks for catching that

lib/createAccessorUsageTables.js
```diff
@@ -0,0 +1,168 @@
...
+                        for (attribute in attributes) {
+                            if (attributes.hasOwnProperty(attribute)) {
+                                attributeAccessorId = attributes[attribute];
+                                // Make sure the attribute accessors haven't been assigned somewhere else
+                                var attributeTableIndex = accessorTableMapping[attributeAccessorId];
+                                if (defined(attributeTableIndex) && attributeTableIndex !== tableIndex) {
+                                    // Merge them into this one if they have been
+                                    mergeUsageTables(tables, accessorTableMapping, tableIndex, attributeTableIndex);
+                                }
+                                table.attributeAccessors[attributeAccessorId] = true;
```
JoshuaStorm: Is there any reason you wouldn't be able to use the `attributeAccessors` attribute type as the value instead of just setting it to true? In order to use `createAccessorUsageTables` in PR #100 I would need the attribute types associated with the accessors and it seems like that would be a good spot to sneak them in.
Thanks!

lilleyse:
Tested a few models, seems good to me.

lib/createAccessorUsageTables.js
lasalvavida: the attributeAccessor value is now assigned to be the semantic

JoshuaStorm: Thanks! My bad on the mixed up terminology

lasalvavida:
@lilleyse This should be ready, merge whenever you're ready

PR #107, opened 2016-06-24
Comment out combinePrimitives and add a TODO for now

lasalvavida:


lilleyse:
Ok this is fine for now.

PR #109, opened 2016-06-24
Force .gltf extension in writeGltf and standardize some tests

lilleyse:
When loading binary and exporting as non-binary, the output extension was still .glb. This PR forces the output extension to be .gltf. I also cleaned up some of the paths for the `gltfPipeline` tests.

@lasalvavida Take a look when you have a chance.


lasalvavida:
Tested, .glb file now output as .gltf unless specified otherwise. Thanks @lilleyse!

PR #111, opened 2016-06-24
Find accessor min and max from data if we're quantizing to avoid errors

lasalvavida:
@lilleyse For the converter


lilleyse:
Tested, looks good.

lilleyse:
Actually a test is failing.

lasalvavida:
@lilleyse this should be good now

pjcozzi:
Can we add a comment about why we are doing this?  As a workaround for invalid models, see #84.

I expect that this workaround will not be needed once the exporters/converters are updated.

PR #113, opened 2016-06-27
Added option to bake AO to vertices

likangning93:
I added an option to bake AO to vertices! After a lot of experimentation, I ended up with this approach:

By default AO is computed at each triangle center, adding the contribution to each of the triangle's vertices, and then averaging the contribution in each vertex over the number of samples that contributed to it.

To produce better results over models with low tesselation, the triangles can be supersampled on a grid. Basically, each triangle gets "rasterized" into its own plane in model space, and then each "pixel" in the triangle gets raytraced. I think this is actually superior to naive tesselation since many building models will have triangles of varying sizes, and this kind of enforces a uniform number of samples per unit area. It also conveniently lets us use a lot of the existing code for baking AO to textures.

Here's the straight AO for a building model with a resolution of 5 sample points per unit area and 16 rays per sample point.

building_reso_5

It took almost 4 hours to bake because the building is quite big in model space (so lots of sample points), but hopefully speeding up intersection tests will get that down to something reasonable. Also, a version with just 1 sample point per unit area rendered in around 8 minutes and doesn't look too different unless you flip between identical images of the two. Blended with dynamic lighting, the difference would probably be pretty low on this model. When things are faster we should do some experiments to assess the runtime/quality tradeoff.

building_reso_2


pjcozzi:
Sounds great!  @likangning93 when you are finished AO, including the uniform grid optimizations, do you want to write a technical post for the Cesium blog?

likangning93:
I'd love to!

pjcozzi:
Given the approach and visual quality here, what do think about generating UVs?  Skip it for now and add it to a roadmap for later?

bin/gltf-pipeline.js
```diff
@@ -17,10 +18,17 @@ if (process.argv.length < 3 || defined(argv.h) || defined(argv.help)) {
...
+        '  --ao.shaderMode, for displaying vertex-baked occlusion. Valid settings are multiply, replace, and blend. Defaults to blend.\n' +
+        '  --ao.blendAmount, for displaying vertex-baked occlusion by blending with the original shader output. Defaults to 50%.\n' +
+        '  --ao.triangleAverage, bake ambient occlusion averaged over triangles to vertices. Defaults to false.\n' +
+        '  --ao.groundPlane, add a simulated ground plane to the scene for computing ambient occlusion. Defaults to false.\n' +
         '  --ao.scene, specify which scene to bake AO for. Defaults to the gltf default scene.\n' +
+        '  --ao.nearCull, near-distance at which to cull intersections when raytracing. Defaults to 0.0001 in world space.\n' +
         '  --ao.rayDepth, ray distance for raytraced ambient occlusion. Defaults to 1.0 units in world space.\n' +
-        '  --ao.resolution, number of texel samples along one dimension for each AO texture. Defaults to 128.\n' +
+        '  --ao.resolution: When baking to texture, this specifies the resolution of the AO texture that will be overlayed onto the diffuse texture. \n' +
```
pjcozzi: Should we break this into two arguments?

lib/bakeAmbientOcclusion.js
```diff
@@ -43,20 +48,323 @@ function bakeAmbientOcclusion(gltf, options) {
...
+function cloneJsonAsNeeded(id, cloneIds, items) {
+    var cloneID = cloneIds[id];
+    if (defined(cloneID)) {
+        if (cloneID === '') { // No clone exists yet. Make one!
+            cloneID = id + "_noAO";
+            items[cloneID] = clone(items[id]);
+            cloneIds[id] = cloneID;
+        }
+        return cloneID;
+    } else {
```
pjcozzi: Drop the `else`.

Just `return id;`

See https://github.com/AnalyticalGraphicsInc/cesium/tree/master/Documentation/Contributors/CodingGuide#functions

pjcozzi:
@likangning93 for testing, can you post a few typical command line examples?

pjcozzi:
@likangning93 you should have commit access to this repo (let me know if you do not) so you can work directly in this repo in a branch instead of in your fork.

lib/bakeAmbientOcclusion.js
```diff
@@ -584,3 +1074,43 @@ function sumBarycentric(barycentric, vector0, vector1, vector2, result) {
...
     return result;
 }
+
+var xAxis = new Cartesian3();
+var zAxis = new Cartesian3();
+var yAxis = new Cartesian3();
+// Returns the point's x/y coordinates in the triangle's plane where position0 is the origin
+// and position1 - position0 is the "x" axis.
+// http://paulbourke.net/geometry/pointlineplane/
+function flattenTriangle(positions, results) {
```
likangning93: The stuff from the Paul Bourke site is all just math, should I keep the reference here if there's no code?

likangning93:
Testing example that should run pretty quickly:
`node ./bin/gltf-pipeline.js ./cube_over_ground_tess.gltf --ao.vertex --ao.samples=16 --ao.shaderMode=replace --ao.resolution=12 --ao.triangleAverage --ao.rayDepth=1.0 --ao.groundPlane=true -o hello_AO_vertices_tess.gltf`

And for the building:
`node ./bin/gltf-pipeline.js ./building.gltf --ao.vertex --ao.samples=16 --ao.shaderMode=replace --ao.resolution=2 --ao.triangleAverage --ao.rayDepth=1.0 --ao.groundPlane=true -o building_AO.gltf`

I think these were the settings that took about 8 minutes, but I need to run it again to make sure. I'll update this when it's done.

[models.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/337540/models.zip)

lib/bakeAmbientOcclusion.js
pjcozzi: If you didn't port the code then it is OK to keep this link, but explicitly mention that it is a link to a reference so we don't question it in the future.

likangning93:
@pjcozzi I think we should skip UVs for now. As you mentioned as well, baking to vertices is much less costly in terms of payload than just UVs alone, and I also think that for some of the buildings the texture resolution would have to be pretty high to get good results.

bin/gltf-pipeline.js
likangning93: Yes, I think `density` is more accurate.

likangning93:
Updated! New commands are:

Testing example that should run pretty quickly:
`node ./bin/gltf-pipeline.js ./cube_over_ground_tess.gltf --ao.vertex --ao.samples=16 --ao.shaderMode=replace --ao.density=12 --ao.triangleAverage --ao.rayDepth=1.0 --ao.groundPlane=true -o hello_AO_vertices_tess.gltf`

And for the building:
`node ./bin/gltf-pipeline.js ./building.gltf --ao.vertex --ao.samples=16 --ao.shaderMode=replace --ao.density=1 --ao.triangleAverage --ao.rayDepth=2.0 --ao.groundPlane=true -o building_AO.gltf`

[models.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/337693/models.zip)

lib/bakeAmbientOcclusion.js
```diff
@@ -43,20 +48,323 @@ function bakeAmbientOcclusion(gltf, options) {
...
+                }
+                glslPassThrough += passUVThrough;
+                glslDiffuseColor = 'texture2D(' + glslDiffuseColor + ', v_ambientOcclusionUV)';
+            }
+            glslDiffuseColor += '.rgb';
+        }
+
+        switch(options.shaderMode) {
+            case 'multiply':
+                glslChangeColor = 'gl_FragColor.rgb *= (1.0 - v_ambientOcclusion); \n';
```
pjcozzi: Minor, but why not store `1.0 - v_ambientOcclusion` in the attribute data so we avoid this per-fragment subtraction.

lib/bakeAmbientOcclusion.js
```diff
@@ -43,20 +48,323 @@ function bakeAmbientOcclusion(gltf, options) {
...
+                glslChangeColor = 'gl_FragColor.rgb = mix(gl_FragColor.rgb, (1.0 - v_ambientOcclusion) * ' +
+                    glslDiffuseColor + ', ' + (options.blendAmount + 0.0) + '); \n';
+                break;
+        }
+
+        // Replace the shaders in the gltf
+        var program = gltf.programs[technique.program];
+        if (!shadersEdited[program.vertexShader]) {
+            shadersEdited[program.vertexShader] = true;
+            editShader(shaders[program.vertexShader], glslNewAttributes + glslNewVaryings, 'mainBeforeAO', glslPassThrough);
```
pjcozzi: Here and below, I don't think we can add `glslNewAttributes` and `glslNewVaryings` at the top of the GLSL code.  For example, if the shader has `#version`, it needs to be the first line before comments and whitespace.  Perhaps add these just before the `mainBeforeAO` function.

lib/bakeAmbientOcclusion.js
```diff
@@ -43,20 +48,323 @@ function bakeAmbientOcclusion(gltf, options) {
...
+            shadersEdited[shaderID] = false;
+        }
+    }
+
+    // For each technique,
+    var techniquesCount = techniquesIDsToLookAt.length;
+    for (var i = 0; i < techniquesCount; i++) {
+        var technique = gltf.techniques[techniquesIDsToLookAt[i]];
+
+        var glslNewAttributes = 'attribute float a_ambientOcclusion; \n'; // snippet for adding attributes
```
pjcozzi: For the roadmap, we should also consider just storing this, for example, as the `w` component of the position or normal to reduce the total number of attributes, which will mean fewer WebGL calls (at least without VAOs) and being less likely to hit the limit on the maximum number of vertex attributes.

No need to do this now though.

lib/bakeAmbientOcclusion.js
```diff
@@ -43,20 +48,323 @@ function bakeAmbientOcclusion(gltf, options) {
...
+        },
+        extras: {
+            _pipeline: {
+                source: new Buffer(new Float32Array(allAOData).buffer)
+            }
+        }
+    };
+
+    // add buffer view
+    gltf.bufferViews.aoBufferView = {
```
pjcozzi: What if a buffer view named `aoBufferView` already exists?

We should have a generic function that can return a new unique id given some prefix, e.g., `getUniqueId('aoBuffer')` might return `aoBuffer0` or `aoBuffer1`.  I think something like this is already used in a few places.

@lasalvavida may know.

Can we clean it up, unit test it, and use it everywhere?

Also, remember that ids are global, e.g., buffers and buffer views use the same namespace.

pjcozzi:
@likangning93 have you tried quantizing the AO attribute using @lasalvavida's recent work?  If so, how is the visual quality?

pjcozzi:
Since each index (in glTF and WebGL) references and entire vertex, not an individual attribute, we have to keep in mind that this approach has the potential to bloat vertex data, especially for flat areas like the sides of buildings (where the occlusion factor may be mostly zero anyway).  On the other hand, the texture approach requires just per-vertex UVs + the texture map.

Perhaps triangles can be adaptively sampled base on the occlusion factor range for the samples?

pjcozzi:
Try converting https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Specs/Data/Models/CesiumAir/Cesium_Air.gltf with
```
$ node ./bin/gltf-pipeline.js ./Cesium_Air.gltf --ao.vertex --ao.samples=16 --a
o.shaderMode=replace --ao.resolution=2 --ao.triangleAverage --ao.rayDepth=1.0 -
-ao.groundPlane=true -o building_AO.gltf
```
The shader doesn't compile when loading the generated model perhaps because the precision qualifier is no longer that the top of the shader.

pjcozzi:
The milk truck also throws an out of range error:
```
$ node ./bin/gltf-pipeline.js ./CesiumMilkTruck.glb --ao.vertex --ao.samples=16
 --ao.shaderMode=replace --ao.resolution=2 --ao.triangleAverage --ao.rayDepth=1
.0 --ao.groundPlane=true -o building_AO.gltf
buffer.js:622
    throw new RangeError('index out of range');
    ^

RangeError: index out of range
    at checkOffset (buffer.js:622:11)
    at Buffer.readUInt16LE (buffer.js:668:5)
    at readBufferComponentType (c:\source\gltf-pipeline\lib\readBufferComponentT
ype.js:16:27)
    at findAccessorMinMax (c:\source\gltf-pipeline\lib\findAccessorMinMax.js:33:
33)
    at accessorDefaults (c:\source\gltf-pipeline\lib\addDefaults.js:24:30)
    at addDefaults (c:\source\gltf-pipeline\lib\addDefaults.js:561:5)
    at processJSONWithExtras (c:\source\gltf-pipeline\lib\gltfPipeline.js:43:5)
    at c:\source\gltf-pipeline\lib\gltfPipeline.js:80:9
    at c:\source\gltf-pipeline\lib\readGltf.js:45:13
    at c:\source\gltf-pipeline\lib\loadGltfUris.js:36:9
```
Is this because it is using Binary glTF?

pjcozzi:
> @pjcozzi I think we should skip UVs for now. As you mentioned as well, baking to vertices is much less costly in terms of payload than just UVs alone, and I also think that for some of the buildings the texture resolution would have to be pretty high to get good results.

OK, we'll include it in the future roadmap.

likangning93:
cesium air and the gltf version of the milk truck both work with the recommended changes to the shader code. I get the same problem with the glb, but that doesn't seem to be part of the AO code since it runs into the same problem without any AO arguments. I might just have to merge in master

likangning93:
Should I also go ahead and implement `getUniqueId` ?
Also, I don't think I've tried this with quantized yet, but I will take a look.

pjcozzi:
The ray tracing doesn't take into account translucent or transparent triangles, correct?  That is fine, but should be well documented.

pjcozzi:
Likewise, we should also document that this is for a static scene with the default pose (animation, skins).

bin/gltf-pipeline.js
```diff
@@ -17,10 +18,16 @@ if (process.argv.length < 3 || defined(argv.h) || defined(argv.help)) {
         '  -s --separate, writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.\n' +
         '  -t --separateImage, write out separate textures, but embed geometry/animation data files, and shader files.\n' +
         '  -q, quantize the attributes of this model.\n' +
-        '  --ao.diffuse, bake ambient occlusion into the diffuse texture. Defaults to false.\n' +
+        '  --ao.vertex, bake ambient occlusion to vertices instead of the diffuse texture and modify the shader. Defaults to false.\n' +
```
pjcozzi: Exposing all these options is fine for the JavaScript/Node API and for advanced command-line users, but most of our users will not know how to tweak these.  They will understand:
* AO on or off
* Some type of scale for "slower and nice looking" and "faster and OK looking"
* Ground plane on or off

Is there a way to translate these user-friendly parameters to all the potential options?  Are any options not important enough to expose?  Can any of these options be generated based on the model, e.g., `density` based on the bounding box?

lib/bakeAmbientOcclusion.js
```diff
@@ -43,20 +48,323 @@ function bakeAmbientOcclusion(gltf, options) {
...
+            NodeHelpers.forEachPrimitiveInScene(gltf, otherScene, cloneShadingChain, parameters);
+        }
+    }
+
+    // Edit the shaders.
+    // This can mean tracking down the unlit diffuse color input from the technique.
+    addAoToShaders(gltf, Object.keys(parameters.techniqueCloneIDs), options);
+
+    // Edit the programs: add the ao attribute
+    for (var programID in parameters.programCloneIDs) {
```
pjcozzi: Here, below, and throughout, extract things like `parameters.programCloneIDs` out of the loop.

lib/bakeAmbientOcclusion.js
```diff
@@ -43,20 +48,323 @@ function bakeAmbientOcclusion(gltf, options) {
...
+        }
+    }
+
+    // Edit the techniques: add ao to attributes, add ao to parameters
+    for (var techniqueID in parameters.techniqueCloneIDs) {
+        if (parameters.techniqueCloneIDs.hasOwnProperty(techniqueID)) {
+            var technique = gltf.techniques[techniqueID];
+            technique.attributes.a_ambientOcclusion = 'vertex_ao';
+            technique.parameters.vertex_ao = {
+                semantic: 'VERTEX_AO',
```
pjcozzi: Is this semantic used anywhere?  If so, let's discuss.  If not, remove it since this semantic is not defined in the glTF spec.  It would need to be a custom semantic, e.g., `_AO` (starts with an underscore), but I am hesitant to start to introduce custom semantics in gltf-pipeline unless we have a use case.  If we do introduce them, e.g., so users can generate their own shaders at runtime, they need to be well documented so they don't collide with an app's semantics.

lib/bakeAmbientOcclusion.js
```diff
@@ -43,20 +48,323 @@ function bakeAmbientOcclusion(gltf, options) {
...
+
+    NodeHelpers.forEachPrimitiveInScene(gltf, scene, concatenateAoBuffers, parameters);
+
+    var primitiveOrder = parameters.primitiveOrder;
+    var allAOData = parameters.allAOData;
+    var allAODataLength = allAOData.length;
+
+    gltf.buffers.aoBuffer = {
+        input: {
+            byteLength : allAODataLength * 4, // (FLOAT). TODO: add option to bake to shorts
```
pjcozzi: Is quantization better than using shorts since we know the tight min/max (instead of `0...1`)?  If so, then we won't need shorts at all.

lib/bakeAmbientOcclusion.js
```diff
@@ -584,3 +1074,42 @@ function sumBarycentric(barycentric, vector0, vector1, vector2, result) {
     Cartesian3.add(result, sumBarycentricScratch, result);
     return result;
 }
+
+var xAxis = new Cartesian3();
+var zAxis = new Cartesian3();
+var yAxis = new Cartesian3();
+// Returns the point's x/y coordinates in the triangle's plane where position0 is the origin
+// and position1 - position0 is the "x" axis.
+function flattenTriangle(positions, results) {
```
pjcozzi: This could go into its one file in case it is useful elsewhere.

lib/bakeAmbientOcclusion.js
```diff
@@ -584,3 +1074,42 @@ function sumBarycentric(barycentric, vector0, vector1, vector2, result) {
...
+    results[2].x = pointLineDistanceParametric(positions[2], positions[0], xAxis);
+    results[2].y = pointLineDistanceParametric(positions[2], positions[0], yAxis);
+
+    return results;
+}
+
+// Returns the parametric position on the line closest to the given point.
+// Math reference: http://paulbourke.net/geometry/pointlineplane/
+var pointOnLineScratch = new Cartesian3();
+function pointLineDistanceParametric(point, linePosition, lineDirection) {
```
pjcozzi: Same comment.

lib/bakeAmbientOcclusion.js
likangning93: As discussed offline: it seems like we need the semantic, so we'll use `_OCCLUSION` and document it in case of conflicts with user provided custom semantics.

likangning93:
Updated. For hiding parameters: since we want to still expose these parameters but include dynamic defaults, can we discuss how this should be implemented? Also, should that be in a separate PR?

likangning93:
For "tweakless" AO parameters, how about:
> ao.texture = true, false. Default to false.
> ao.groundPlane = true, false. Defaults to false.
> ao.blendAmount = 0-1. Default to 0.5, but not applicable to texture AO
> ao.quality = low, medium, high. Use this to compute rays per sample point and sample point resolution.
> ao.scale = small, medium, large. Is the model a toy, a car, or a building? Use this to compute ray distance from the scene's bounding box size.
> ao.scene = which gltf scene should the AO be baked for? Defaults to the default scene.

pjcozzi:
> ao.blendAmount = 0-1. Default to 0.5, but not applicable to texture AO

The end user may think "blend with what?"  Is "darkness" or something like that appropriate?

> ao.scale = small, medium, large. Is the model a toy, a car, or a building? Use this to compute ray distance from the scene's bounding box size.

What does this mean though? 1cm, 1m, 1km?  What if I have a detailed bolt with 1M triangles and a city with 1M triangles?  Is the bolt small and the city large?

> ao.scene = which gltf scene should the AO be baked for? Defaults to the default scene.

Perhaps just drop this and do all scenes?  I don't think we have an option like this elsewhere in the pipeline.

likangning93:
For the blend amount, perhaps `ao.blendWithShaderLighting`, `ao.lightingContribution`, or `ao.ambientOcclusionContribution`?
I think `ao.ambientOcclusionContribution` best describes what the code currently does, but is it too technical?

For scale, should we assume that the model is "to scale" in some space? Like, if you specify `centimeters` we will treat the model coordinates as centimeters and `meters` gets `meter` units?
If not I was going to do something like:

- small: rays bottom out after 1/5 of the largest AABB dimension
- medium: rays bottom out after 1/10 of the largest AABB dimension
- large: rays bottom out after 1/20 of the largest AABB dimension

We would have to experiment with what proportions to use, but I think this best covers all potential models.
Or we could also have the user input an estimate of the model's largest dimension in meters or centimeters and then compute ray falloff from that.

If we bake for all scenes, primitives shared between scenes which would end up "double exposed" for AO. I'm not sure if it's a good idea for the AO stage to enforce primitive uniqueness.

pjcozzi:
> I think ao.ambientOcclusionContribution best describes what the code currently does, but is it too technical?

What does the users see when this is `0`?  When this is `1`?  This is a sliding scale between lighting and AO.  I still think something indicating "darkness" will make sense to users even though it is not at all precise.

`ambientOcclusionContribution` is OK if that is the best we have, but I think any user would have to read at least the tooltip for it.

> For scale, should we assume that the model is "to scale" in some space? Like, if you specify centimeters we will treat the model coordinates as centimeters and meters gets meter units?

glTF is always in meters, and we can compute the bounding box.

> If we bake for all scenes, primitives shared between scenes which would end up "double exposed" for AO. I'm not sure if it's a good idea for the AO stage to enforce primitive uniqueness.

Perhaps just use the default scene until someone asks otherwise; I don't know anyone using something other than the default scene yet in glTF.

likangning93:
Updated!
I cleaned up the parameters some and added a method to bakeAmbientOcclusion that processes basic inputs and advanced inputs. You can now bake ao from the command line tool with just a `--ao` flag.

I think we can leave the ray depth/ray distance parameter out of basic options entirely. Right now it defaults to 1.0 units, but we can also evaluate further down the line if it should be dynamic.

I also realized I've been using yargs wrong, yargs parses the `true` in something like `--option=true` as a string instead of a boolean, so I adjusted the `ao.*` flags to take this into account. New commands are:

Testing example that should run pretty quickly:
`node ./bin/gltf-pipeline.js ./cube_over_ground_tess.gltf --ao.quality=medium --ao.groundPlane --ao.triangleCenterOnly -o hello_AO_vertices_tess.gltf`

And for the building:
`node ./bin/gltf-pipeline.js ./building.gltf --ao.shaderMode=replace  --ao.rayDistance=2.0 --ao.groundPlane -o building_AO.gltf`

[models.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/337693/models.zip)

lib/bakeAmbientOcclusion.js
```diff
@@ -11,52 +11,428 @@ var Matrix3 = Cesium.Matrix3;
...
+// - numberRays: number of rays to cast from each sample point. Clamps to nearest smaller square. Default: 64, 36, 16
+// - triangleCenterOnly: for vertex baking mode, do not sample a triangle's area in addition to its center using a grid. Default: false
+
+// Advanced appearance settings:
+// - rayDistance: ray bottom-out distance in world-space units. Default: 1.0
+// - nearCull: near cull depth for ray-triangle collisions. Default: 0.0001 units in world space.
+// - shaderMode: how should the shader use per-vertex AO? Valid settings are multiply, replace, and blend. Default: blend
+// - scene: which scene to use when baking AO. Default: gltf default
+
+function bakeAmbientOcclusion(gltf, aoOptions) {
```
pjcozzi: Throughout, why pass separate `gltf` and `aoOptions`.  Why not just pass `options` like Cesium does?  https://github.com/AnalyticalGraphicsInc/cesium/tree/master/Documentation/Contributors/CodingGuide#options-parameters

bin/gltf-pipeline.js
```diff
@@ -17,11 +17,10 @@ if (process.argv.length < 3 || defined(argv.h) || defined(argv.help)) {
         '  -s --separate, writes out separate geometry/animation data files, shader files and textures instead of embedding them in the glTF file.\n' +
         '  -t --separateImage, write out separate textures, but embed geometry/animation data files, and shader files.\n' +
         '  -q, quantize the attributes of this model.\n' +
-        '  --ao.diffuse, bake ambient occlusion into the diffuse texture. Defaults to false.\n' +
-        '  --ao.scene, specify which scene to bake AO for. Defaults to the gltf default scene.\n' +
-        '  --ao.rayDepth, ray distance for raytraced ambient occlusion. Defaults to 1.0 units in world space.\n' +
-        '  --ao.resolution, number of texel samples along one dimension for each AO texture. Defaults to 128.\n' +
-        '  --ao.samples, sample count for ambient occlusion texel. Clamps to the nearest smaller perfect square. Defaults to 16.\n';
+        '  --ao.toTexture: Bake AO to existing diffuse textures instead of to vertices. Does not modify shaders.\n' +
```
pjcozzi: Update the README.md to include these.

lib/GeometryMath.js
```diff
@@ -0,0 +1,64 @@
...
+module.exports = {
+    sumBarycentric: sumBarycentric,
+    flattenTriangle: flattenTriangle,
+    pointLineDistanceParametric: pointLineDistanceParametric
+};
+
+// Functions for working with geometry like triangles and lines
+
+var sumBarycentricScratch = new Cartesian3();
+function sumBarycentric(barycentric, vector0, vector1, vector2, result) {
```
pjcozzi: Throughout this file, throw `DeveloperError` if a required property is undefined.  Include the Cesium pragmas so we can optimize them out in the future.

```javascript
        //>>includeStart('debug', pragmas.debug);
        if (!defined(spherical)) {
            throw new DeveloperError('spherical is required');
        }
        //>>includeEnd('debug');
```

@mramato are you doing this anywhere with your node work yet?

bin/gltf-pipeline.js
pjcozzi: Also, should this list include just `--ao`?

bin/gltf-pipeline.js
```diff
@@ -17,11 +17,10 @@ if (process.argv.length < 3 || defined(argv.h) || defined(argv.help)) {
...
+        '  --ao.groundPlane: Add a groundplane at the lowest point of the model for AO computation.\n' +
```
pjcozzi: What is the default?

bin/gltf-pipeline.js
pjcozzi: What is the default?

lib/GeometryMath.js
mramato: We don't use pragmas or optimize out `DeveloperError` in node code and I doubt we will anytime soon.  The pragma is specifically an AMD/requirejs thing so it doesn't exist in Node and even if we decided to do some sort of Node build, server side code needs more robust logging than client code; so we probably would keep them anyway.  So definitely check for and throw `DeveloperError`, but you can leave out the pragma.

pjcozzi:
The default ray distance is going to be a % of the bounding box as discussed offline.

lib/GeometryMath.js
pjcozzi: I still want to leave the pragma because we are doing more and more heavy math here and elsewhere and we will ultimately want to optimize them out.

mramato: But there are no Node build tools to optimize them out. Node code doesn't get combined/minified like browser code does and even if we did, we certainly wouldn't be using requirejs to do it (which is the only system that uses these pragmas).  Just like we don't leave TODOs or commented out code in master; we shouldn't leave these unused pragmas in either.

Even if we had some sort of Node build, server-side code is a different beast than client code and removing error checking is a bad idea.  I can understand your point about Math code, but as we've already seen in Cesium, NaNs still sneak in unexpectedly, especially when you are dealing with external data like the gltf pipeline will.  Server code should have more checks.

If for some reason we ever need the pragmas, it will be easy to find all of the places we are throwing developer error and add them back it; you could probably even do it with a regex.

Finally, I will add that if we want to perform extra error checking for non-production builds, the standard way to do this is by checking for the environment variable, `NODE_ENV=production`, and only run code if it's not set.  But that doesn't make sense for most developer errors, since you're just replacing one if check with another; but it can still be useful for performance-intensive error checks.

pjcozzi: Let's discuss offline when we have time.  The performance need here is perhaps even more so than Cesium.

likangning93:
Some updates:
- I updated a bunch of internal functions in `bakeAmbientOcclusion` to use options parameters. Self documenting code is nice! Unfortunately there were a couple really nasty functions that are pretty hot, in particular `trianglePixelMarch`.
- there is a pipeline conflict with cacheOptimization, see https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/122. It makes sense to bake AO after cacheOptimization anyway, though, and that gets around the problem.

I will address the dynamic default ray depth next.

likangning93:
Updated! Right now the default is to make the ray distance 20% of the scene's smallest AABB dimension. We should wait until the uniform grid is integrated to tune this, since otherwise doing many renders of different buildings will take too long.

pjcozzi:
> Unfortunately there were a couple really nasty functions that are pretty hot, in particular trianglePixelMarch.

To avoid implicit allocations when creating the object literal to pass as the options parameter, use a scratch variable.

> Updated! Right now the default is to make the ray distance 20% of the scene's smallest AABB dimension. 

This will not work well the more "non-square" a model is. 20% of the longest dimension would be safer.  You could also compute this per ray; perhaps 20% of the dimension that the ray projects onto most.  I dunno, could produce some interesting results.

lib/bakeAmbientOcclusion.js
```diff
@@ -327,19 +861,95 @@ function processPrimitive(primitive, meshPrimitiveID, parameters, node) {
...
+                    raytracerScene, pixelWidth, resolution, callback, parameters) {
+    var uMin = Math.min(uv0.x, uv1.x, uv2.x);
+    var vMin = Math.min(uv0.y, uv1.y, uv2.y);
+    var uMax = Math.max(uv0.x, uv1.x, uv2.x);
+    var vMax = Math.max(uv0.y, uv1.y, uv2.y);
+
+    // Perform a pixel march over the
+    // 0.0, 0.0 to width, width is the bottom left pixel
+    // 1.0-width, 1.0-width to 1.0, 1.0 is the top right pixel
+    // TODO: borrow from conservative rasterization: http://http.developer.nvidia.com/GPUGems2/gpugems2_chapter42.html
```
pjcozzi: Remove this TODO and add the link to the roadmap issue.

lib/cacheOptimization.js
```diff
@@ -0,0 +1,105 @@
+'use strict';
```
pjcozzi: @likangning93 did you merge `master` into `ao-master` and `ao-master` into this?  It looks like some changes are in this branch but not `ao-master`.

pjcozzi:
Just those comments.

@lilleyse please do final testing and merge and ready.

Great job @likangning93, and thanks for the detailed roadmap for future work!

lib/cacheOptimization.js
likangning93: I think I might have. Should I have merged into `ao-master` first?

likangning93: I did that because I needed `getUniqueId`.

pjcozzi: Just merge master into ao-master and then merge ao-master into this branch and everything will be as expected.

likangning93:
Updated!

@pjcozzi, I thought 20% of the minimum AABB dimension might be better for our use case since I figured most buildings would share a similar footprint but would vary widely in height, so using the minimum AABB dimension would keep the ray distances pretty uniform. I think most non-cubic models also tend to have detail "scaled" for the minimum dimension. For example, [Sponza's arches](http://likangning93.github.io/Project6-WebGL-Deferred-Shading/) are scaled so that there are a few in each minimum dimension and a bunch in each maximum dimension instead of 2 in each maximum dimension and none in each minimum.

lib/bakeAmbientOcclusion.js
```diff
@@ -39,6 +39,34 @@ var cartesian2Scratch = new Cartesian2();
 var quaternionScratch = new Quaternion();
 var matrix3Scratch = new Matrix3();
 var matrix4Scratch = new Matrix4();
+var trianglePixelMarchScratch = {
+    uv0 : cartesian2Scratch,
```
pjcozzi: No need to include the `Scratch` prefix here since it is already in `trianglePixelMarchScratch`; however, I don't know that I would use an object literal here since property access is likely to be slow.  See https://www.youtube.com/watch?v=XAqIpGU8ZZk

lib/bakeAmbientOcclusion.js
```diff
@@ -39,6 +39,34 @@ var cartesian2Scratch = new Cartesian2();
...
+    inverseTranspose : matrix4Scratch,
+    raytracerScene : {},
+    pixelWidth : 0.0,
+    resolution : 0,
+    callback : {},
+    parameters : {}
+};
+
+var computeAmbientOcclusionAtScratch = {
+    position : worldPositionScratch,
```
pjcozzi: Same comment here and throughout if needed.

pjcozzi:
> I thought 20% of the minimum AABB dimension might be better for our use case since...

I am skeptical here given the diversity of models used in Cesium (this will be used beyond buildings) and many buildings have vastly different footprints, e.g., Philadelphia convention center vs. row home.

We can keep this and see how it works in practice, but I'm afraid many models will come back "without much AO" because the distance ends up being too conservative.

lib/bakeAmbientOcclusion.js
likangning93: Should I switch these two (`computeAmbientOcclusion` and `trianglePixelMarch`) back to just using arguments? From the video it seems like inline caching would speed up property access but since that would require a constructor function we would lose the self documentation.

pjcozzi: You could add the constructor function, construct these once, and then just set the properties on each function call.

likangning93: Oh, I see. Thanks!

likangning93:
Updated!

lib/bakeAmbientOcclusion.js
```diff
@@ -902,6 +877,25 @@ function processPrimitive(primitive, meshPrimitiveID, parameters, node) {
 
 ////////// Rendering //////////
 
+function TrianglePixelMarchOptions() {
+    this.uv0 = {};
```
pjcozzi: Why set these to an empty object?  Why not set them to their actual type, e.g., `new Cartesian3()`?

likangning93:
Updated.

pjcozzi:
Exactly, looks good.

@lilleyse merge when ready!

lib/GeometryMath.js
```diff
@@ -0,0 +1,104 @@
...
+    //>>includeStart('debug', pragmas.debug);
+    if (!defined(barycentric)) {
+        throw new DeveloperError('in GeometryMath: barycentric is required');
```
mramato: Get rid of `in GeometryMath: ` throughout your error messages, that's what callstacks are for.

likangning93:
Thanks @mramato!

lilleyse:
In the outputted glTF the uri type for the aoBuffer is `data:[object Object]` while buffers are normally `data:application/octet-stream`

lilleyse:
Just a thought, unless specified should `toTexture` always be set to true if the primitive has a diffuse texture?

lib/GeometryMath.js
```diff
@@ -0,0 +1,104 @@
...
+    result.x = 0.0;
+    result.y = 0.0;
+    result.z = 0.0;
+    Cartesian3.multiplyByScalar(vector0, barycentric.x, sumBarycentricScratch);
+    Cartesian3.add(result, sumBarycentricScratch, result);
+    Cartesian3.multiplyByScalar(vector1, barycentric.y, sumBarycentricScratch);
+    Cartesian3.add(result, sumBarycentricScratch, result);
+    Cartesian3.multiplyByScalar(vector2, barycentric.z, sumBarycentricScratch);
+    Cartesian3.add(result, sumBarycentricScratch, result);
+    return result;
```
lilleyse: I think these lines might actually be clearer by not using the `Cartesian` functions, and probably faster too.

lilleyse: But it doesn't matter that much to me, I'm also fine as is.

lilleyse:
The blend AO setting has a couple edge cases that may be worth addressing. Most of the glTF shaders premultiply the alpha: `color = vec4(color.rgb * diffuse.a, diffuse.a);`. This will cause the final ao blending `gl_FragColor.rgb = mix(gl_FragColor.rgb, v_ambientOcclusion * u_diffuse.rgb, 0.5);` to become incorrect, and get results like this:

air

If it's somehow possible to blend the AO with the diffuse color before gl_FragColor, that would be ideal.

lilleyse:
I noticed some artifacts when rendering to a texture. In general the vertex shading looks a lot smoother to me.
airbottom
airbottom2

lilleyse:
I used:
`node ./bin/gltf-pipeline.js ./test/Cesium_Air.glb --ao.quality=medium --ao.groundPlane --ao.toTexture -o ./test/hello_AO_air.gltf`
and
`node ./bin/gltf-pipeline.js ./test/Cesium_Air.glb --ao.quality=medium --ao.groundPlane -o ./test/hello_AO_air_verts.gltf`

lilleyse:
I'll probably do more testing on the other PR just because it's faster to generate, but the code here looks good.

lib/bakeAmbientOcclusion.js
```diff
@@ -11,53 +11,495 @@ var Matrix3 = Cesium.Matrix3;
...
+            for (var uniformName in uniforms) {
+                if (uniforms.hasOwnProperty(uniformName)) {
+                    if (uniforms[uniformName] === 'diffuse') {
+                        glslDiffuseColor = uniformName;
+                        break;
+                    }
+                }
+            }
+            if (diffuseParameter.type === WebGLConstants.SAMPLER_2D) {
+                glslNewVaryings += 'varying vec2 v_ambientOcclusionUV; \n';
```
lilleyse: As we talked about offline, it would be better to reuse the shader's texcoord varying if we can get its name.

likangning93:
As discussed offline, I added some code to locate the existing texcoord varying and changed the shader modification code to premultiply alpha onto the AO color contribution. Observation: alpha blending sometimes behaves differently between machines?

lib/bakeAmbientOcclusion.js
```diff
@@ -332,7 +332,8 @@ function addVertexData(options) {
         },
         extras: {
             _pipeline: {
-                source: new Buffer(new Float32Array(allAOData).buffer)
+                source: new Buffer(new Float32Array(allAOData).buffer),
+                extension: 'application/octet-stream'
```
lilleyse: This can be `.bin`. Should result in the correct uri type later in the pipeline.

lilleyse:
`varying vec2 v_ambientOcclusionUV;` is still in the final shaders (not used).

lilleyse:
Also success with the plane:

air2

lilleyse:
> Observation: alpha blending sometimes behaves differently between machines?

I definitely need to look into this more. Especially since the examples in https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/128 works on my machine and not yours...

lilleyse:
> I think the line artifacting with rendering to texture is coming from the rasterization I was using for the triangles... I put conservative rasterization on the roadmap, but I'm not sure if we'll get to it before SIGGRAPH.

Ah, that's fine.

likangning93:
Okay. Updated!

lilleyse:
Ok this looks good to merge, if I have any issues I will bring them up in the newer PRs.

lib/GeometryMath.js
pjcozzi: I am hesitant to change this unless it could same some operations.

pjcozzi:
Nice review, thanks @lilleyse!

lib/GeometryMath.js
lilleyse: Sounds fine to me, @likangning93 can you revert this back in your other PR?

likangning93: ok

PR #115, opened 2016-06-28
Ported Tomas Akenine-Moller's triangle-AABB implementation to javascript

likangning93:
Tomas Akenine-Moller's Reference C implementation can be found [here](http://fileadmin.cs.lth.se/cs/Personal/Tomas_Akenine-Moller/code/tribox3.txt).
Theory page [here](http://fileadmin.cs.lth.se/cs/Personal/Tomas_Akenine-Moller/code/).

In preparation for uniform grid acceleration in Ambient Occlusion as well as future spatial datastructures.
Should this actually make its way into Cesium at some point?


pjcozzi:
> Should this actually make its way into Cesium at some point?

Yes, something like a `CesiumCore` or `CesiumGeometry` module as part of the Cesium 2.0 effort, https://github.com/AnalyticalGraphicsInc/cesium/issues/2524.

lib/triBoxOverlap.js
```diff
@@ -0,0 +1,212 @@
+'use strict';
+
+/**
+ * Function and helpers for computing the intersection between a static triangle and a static AABB.
+ * Adapted from Tomas Akenine-Möller's public domain implementation: http://www.cs.lth.se/home/Tomas_Akenine_Moller/code/
```
pjcozzi: Still add this to LICENSE.md as public domain so all third-party code is referenced from there.

lib/triBoxOverlap.js
```diff
@@ -0,0 +1,212 @@
...
+ */
+
+var Cesium = require('cesium');
+var Cartesian3 = Cesium.Cartesian3;
+
+module.exports = triBoxOverlap;
```
pjcozzi: We don't use abbreviations in public APIs so make this `triangleBoxOverlap`.

lib/triBoxOverlap.js
```diff
@@ -0,0 +1,212 @@
...
+        return true;
+    }
+    return false;
+}
+
+////// X-tests //////
+
+function axisTestX01(a, b, fa, fb, boxHalfSize, v0, v2) {
+    var p0 = a * v0.y - b * v0.z;
+	var p2 = a * v2.y - b * v2.z;
```
pjcozzi: Whitespace.

lib/triBoxOverlap.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+    var rad = fa * boxHalfSize.x + fb * boxHalfSize.z;
+    return (min > rad || max < -rad);
+}
+
+////// Z-tests //////
+
+function axisTestZ12(a, b, fa, fb, boxHalfSize, v1, v2) {
+    var p1 = a * v1.x - b * v1.y;
+	var p2 = a * v2.x - b * v2.y;
```
pjcozzi: Same comment.

lib/triBoxOverlap.js
```diff
@@ -0,0 +1,212 @@
...
+        min = p2;
+        max = p1;
+    }
+	var rad = fa * boxHalfSize.x + fb * boxHalfSize.y;
+    return (min > rad || max < -rad);
+}
+
+function axisTestZ0(a, b, fa, fb, boxHalfSize, v0, v1) {
+    var p0 = a * v0.x - b * v0.y;
+	var p1 = a * v1.x - b * v1.y;
```
pjcozzi: Same.

lib/triBoxOverlap.js
```diff
@@ -0,0 +1,212 @@
...
+    var max = p0;
+    if (p0 < p1) {
+        min = p0;
+        max = p1;
+    }
+	var rad = fa * boxHalfSize.x + fb * boxHalfSize.y;
+    return (min > rad || max < -rad);
+}
+
+var triangleNormal = new Cartesian3();
```
pjcozzi: Here and below, use `Scratch` suffix so it is clear these are scratch variables.

lib/triBoxOverlap.js
```diff
@@ -0,0 +1,212 @@
...
+var edges = [];
+edges.push(new Cartesian3());
+edges.push(new Cartesian3());
+edges.push(new Cartesian3());
+
+var v0 = new Cartesian3();
+var v1 = new Cartesian3();
+var v2 = new Cartesian3();
+
+function triBoxOverlap(boxCenter, boxHalfSize, triangle) {
```
pjcozzi: Put this and the scratch variables at the top of the file and the helper function below.  This will read better.  It relies on "hoisting" the helper functions, but that is OK.  Never rely on hoisting variables though.

lib/triBoxOverlap.js
pjcozzi: Actually `triangleAxisAlignedBoundingBoxOverlap` is a more precise name.

lib/triBoxOverlap.js
pjcozzi: Can we make this efficiently use Cesium's [AxisAlignedBoundingBox](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/AxisAlignedBoundingBox.js)?

specs/lib/triBoxOverlapSpec.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+var Cesium = require('cesium');
+var Cartesian3 = Cesium.Cartesian3;
+var triBoxOverlap = require('../../lib/triBoxOverlap');
+
+describe('triBoxOverlap', function() {
```
pjcozzi: Are these my original tests?  How is test coverage?  It is important that we test all the case and combinations since we are going to build on this.

lib/triBoxOverlap.js
likangning93: Looks like the third party code section is pretty out of date, should I open an issue for that or just update it here?

specs/lib/triBoxOverlapSpec.js
likangning93: These tests are new, and they don't cover every possible return path although they cover all the helper functions. Do you have comprehensive tests? Otherwise I could write them, but figuring out the geometry for each case could take some time.

lib/triBoxOverlap.js
pjcozzi: Updating it as part of this PR would be fine and appreciated.

specs/lib/triBoxOverlapSpec.js
pjcozzi: Make sure that code coverage is 100% even if the cyclomatic complexity is not.  Also test representative, but perhaps not exhaustive cases, e.g.,
* Completely inside box
* Completely outside box in each of the possible 26 voxels (or less)
* Intersecting parallel with the x plane, y plane, and z plane
* Diangle intersects
* Etc

We want to be very confident in this code so when we build on it, we don't have to chase down bugs in a more complex system.

likangning93:
Updated: addressed the style issues and added Cesium's AxisAlignedBoundingBox. I will add tests in another commit.

pjcozzi:
Looks good.  Thanks for updating LICENSE.md.

pjcozzi:
Coverage is only 93% for this file.  Can we get it to 100%?

https://coveralls.io/builds/6802561/source?filename=lib%2FtriangleAxisAlignedBoundingBoxOverlap.js

likangning93:
I got coverage to go up a little by fiddling with some of the specs, but as discussed offline getting coverage to 100% will be difficult as the algorithm is heavily optimized for early return.

I think we have enough specs to be pretty confident in this port of the algorithm, though.

likangning93:
Another thought: could we get in touch with Dr. Akenine-Möller about late return edge case triangles?

pjcozzi:
I'll merge this now, and will try to get you in touch offline.

Issue #116, opened 2016-06-28
-o option should recognize extension


pjcozzi:
`-o example.glb` should output Binary glTF, not regular glTF with a `.glb` extension.

Throw an exception if the `-b` option is provided and the extension doesn't match.

lasalvavida:
I thought we fixed this in #109?

lasalvavida:
Closed by #117.

PR #117, opened 2016-06-28
Output correct glTF type based on extension

JoshuaStorm:
See #116 

Also added a `DeveloperError` to be thrown if output path given has neither `.glb` or `.gltf` file extensions.

### DIFF
--- a/bin/gltf-pipeline.js
+++ b/bin/gltf-pipeline.js
@@ -27,6 +27,7 @@ var fileName = path.basename(gltfPath, fileExtension);
 var filePath = path.dirname(gltfPath);
 
 var outputPath = defaultValue(argv._[1], defaultValue(argv.o, argv.output));
+var outputExtension = path.extname(outputPath);
 var binary = defaultValue(defaultValue(argv.b, argv.binary), false);
 var separate = defaultValue(defaultValue(argv.s, argv.separate), false);
 var separateImage = defaultValue(defaultValue(argv.t, argv.separateImage), false);
@@ -38,6 +39,10 @@ if (!defined(gltfPath)) {
 
 if (fileExtension !== '.glb' && fileExtension !== '.gltf') {
     throw new DeveloperError('Invalid glTF file.');
+}
+
+if (outputExtension !== '.gltf' && outputExtension !== '.glb' || binary && outputExtension !== '.glb') {
+    throw new DeveloperError('Invalid output path extension.');
 }
 
 if (!defined(outputPath)) {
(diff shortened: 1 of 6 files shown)


lasalvavida:
@JoshuaStorm Can you add a test making sure the exception gets thrown?

JoshuaStorm:
@lasalvavida I am actually unsure how we would go about testing `gltf-pipeline` since it is just the command line interface. We don't actually have a spec associated with `gltf-pipeline` yet.

lasalvavida:
@JoshuaStorm You should move this check into `lib/gltfPipeline.js` `processFileToDisk` and `processJsonToDisk`. Add the test to `specs/lib/gltfPipelineSpec.js`.

lib/gltfPipeline.js
```diff
@@ -96,11 +97,12 @@ function processFile(inputPath, options, callback) {
 }
 
 function writeFile(gltf, outputPath, options, callback) {
+    var fileExtension = path.extname(outputPath);
```
pjcozzi: I don't know how all of this is wired together, but can't we just make sure that `binary` is set to `true` before this is called; otherwise, this is forcing some of the command-line logic deeper in the engine.

bin/gltf-pipeline.js
```diff
@@ -40,6 +41,10 @@ if (fileExtension !== '.glb' && fileExtension !== '.gltf') {
     throw new DeveloperError('Invalid glTF file.');
 }
 
+if (outputExtension !== '.gltf' && outputExtension !== '.glb' || binary && outputExtension !== '.glb') {
```
pjcozzi: Use more parenthesis.  See https://github.com/AnalyticalGraphicsInc/cesium/tree/master/Documentation/Contributors/CodingGuide#formatting

pjcozzi:
> We don't actually have a spec associated with gltf-pipeline yet.

I think this is fine for now.

lasalvavida:
@pjcozzi I think this check and exception should be thrown in `lib/gltfPipeline.js` so that this also happens when gltf-pipeline is used as a node module, not just from command line.

JoshuaStorm:
@lasalvavida That makes sense, I imagine we should do the same for the other two exceptions in `gltf-pipeline` as well?

lasalvavida:
@JoshuaStorm That sounds good to me. Maybe add a `validateOptions` function and wrap all of that logic up in there.

JoshuaStorm:
@lasalvavida On second thought, wouldn't it make most sense to check for these exceptions within the `readGltf` and `writeGltf` functions? We already do check most of them in these functions anyway.

lasalvavida:
> @lasalvavida On second thought, wouldn't it make most sense to check for these exceptions within the readGltf and writeGltf functions? We already do check most of them in these functions anyway.

That could work too; I'm fine with either.

pjcozzi:
@lasalvavida please merge when you guys are happy.

lasalvavida:
Tested; this resolves #116 correctly. Good job @JoshuaStorm!

PR #118, opened 2016-06-29
Added a getUniqueId helper

likangning93:
This takes some prefix and some JSON and makes sure the id will be unique if it is added to the JSON.

See https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/113

### DIFF
--- a/lib/getUniqueId.js
+++ b/lib/getUniqueId.js
@@ -1,47 +1,24 @@
 'use strict';
-var Cesium = require('cesium');
-var defined = Cesium.defined;
 
 module.exports = getUniqueId;
 
 // Given a prefix for a new ID, checks the json for matching prefixes throughout and generates a unique ID.
-function getUniqueId(json, prefix) {
+function getUniqueId(gltf, prefix) {
     var prefixMatches = [];
-    // Perform stack based depth-first traversal of the json
-    var rootIDs = Object.keys(json);
-    if (!defined(rootIDs)) {
-        return prefix;
-    }
-    var rootIdCount = rootIDs.length;
-    var idString = '';
 
-    var stack = [];
-    var currentObject;
-    for (var i = 0; i < rootIdCount; i++) {
-        // Check ID against the prefix
-        idString = rootIDs[i];
-        if (idString.startsWith(prefix)) {
-            prefixMatches.push(idString);
-        }
-        currentObject = json[idString];
-        if (currentObject.constructor === Object) {
-            stack.push(json[idString]);
-        }
-    }
-
-    while (stack.length > 0) {
-        currentObject = stack.pop();
-        var ids = Object.keys(currentObject);
-        var idCount = ids.length;
-        for (i = 0; i < idCount; i++) {
-            // Check ID against the prefix
-            idString = ids[i];
-            if (idString.startsWith(prefix)) {
-                prefixMatches.push(idString);
-            }
-
-            if (currentObject.constructor === Object) {
-                stack.push(currentObject[idString]);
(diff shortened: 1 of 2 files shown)


lib/getUniqueId.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+
+module.exports = getUniqueId;
+
+// Given a prefix for a new ID, checks the json for matching prefixes throughout and generates a unique ID.
+function getUniqueId(json, prefix) {
```
pjcozzi: `json` should be `gltf` or whatever we call it throughout this project.  `json` is too generic.

pjcozzi:
This doesn't need to search the entire glTF JSON.  It only needs to search top-level objects with ids.

For example, see https://github.com/KhronosGroup/glTF/blob/master/specification/README.md#concepts

> IDs for top-level glTF dictionary objects (accessors, animations, buffers, bufferViews, cameras, images, materials, meshes, nodes, programs, samplers, scenes, shaders, skins, techniques, and textures) are in the same namespace and are unique.

lib/getUniqueId.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+module.exports = getUniqueId;
+
+// Given a prefix for a new ID, checks the json for matching prefixes throughout and generates a unique ID.
+function getUniqueId(gltf, prefix) {
+    var prefixMatches = [];
+
+    // Check each top level object in the gltf for instances of the prefix
+    var topLevelObjectIds = Object.keys(gltf);
```
pjcozzi: Not all top-level objects will have properties that are ids, e.g., `asset`.  Refer to the list I provided in https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/118#issuecomment-229408937 and also used here: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/removeUnused.js

likangning93:
Updated.

specs/lib/getUniqueIdSpec.js
```diff
@@ -0,0 +1,84 @@
...
+        },
+        "materials": {
+            "blinn-1": {},
+            "lambert-1": {}
+        }
+    };
+
+    var allIdentifiersGltf = {
+        "nodes": {
+            "name" : ''
```
pjcozzi: To make this slightly more true to the spec, replace `''` here and below with `{}`.

These glTF properties are objects, not strings.

Also, change `"name*"` to `"id*"` since these are ids, not names.  Some glTF properties also have a name (e.g., used for display, not internal references) so naming an id `"name"` can be confusing.

likangning93:
Updated.

lib/getUniqueId.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+var Cesium = require('cesium');
+var defined = Cesium.defined;
+
+module.exports = getUniqueId;
+
+// Given a prefix for a new ID, checks the gltf for matching prefixes in top-level objects with IDs and returns a unique ID.
+function getUniqueId(gltf, prefix) {
+    var prefixMatches = [];
```
lasalvavida: Use a dictionary here instead of an array. Then you can just do lookups below instead of searching the array every time.

pjcozzi: Good eye, @lasalvavida.  Please merge when ready.

likangning93:
Updated. Thanks @lasalvavida!

lasalvavida:
This looks great, thanks @likangning93!

PR #119, opened 2016-06-29
Ceil instead of rounding

lasalvavida:
When decoding something like a joint, it is used as an index in the shader, so always round up the encoded value.


pjcozzi:
Good catch, @lasalvavida!

PR #123, opened 2016-07-01
packArray to mirror readAccessor, resolve attributes issue

### LINKED ISSUE #122: Attributes that aren't `VEC2`, `VEC3`, or `VEC4` break the pipeline
Passing gltf with custom attributes that aren't `VEC2`, `VEC3`, or `VEC4` right now breaks the pipeline. [Example model.](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/343743/hello_AO_vertices_tess.zip) 

Part of the problem is `packArray` in `gltfPrimitiveToCesiumGeometry` but it might also be elsewhere. Since`packArray` uses `readAccessor` it should handle every type that `readAccessor` can process.

JoshuaStorm:
Closes #122 

Also wondering if I should split `packArray` into its own file. It currently isn't used anywhere else, but I could see it being useful in some cases.

### DIFF
--- a/lib/gltfPrimitiveToCesiumGeometry.js
+++ b/lib/gltfPrimitiveToCesiumGeometry.js
@@ -1,71 +1,28 @@
 'use strict';
 var readAccessor = require('./readAccessor');
 var numberOfComponentsForType = require('./numberOfComponentsForType');
+var packArray = require('./packArray');
 var Cesium = require('cesium');
 var Geometry = Cesium.Geometry;
 var GeometryAttribute = Cesium.GeometryAttribute;
 var ComponentDatatype = Cesium.ComponentDatatype;
-var Cartesian2 = Cesium.Cartesian2;
-var Cartesian3 = Cesium.Cartesian3;
-var Cartesian4 = Cesium.Cartesian4;
-var Matrix2 = Cesium.Matrix2;
-var Matrix3 = Cesium.Matrix3;
-var Matrix4 = Cesium.Matrix4;
 var defined = Cesium.defined;
 
 module.exports = gltfPrimitiveToCesiumGeometry;
-
-// Helper function for packing Cartesians to arrays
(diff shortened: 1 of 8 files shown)


pjcozzi:
@likangning93 can you please review and merge?

> Also wondering if I should split packArray into its own file.

This would make it easier to test, and this PR needs unit tests.

likangning93:
I think this looks good. Splitting it out into its own file should also make writing the unit tests pretty straightforward.

One thing that does bother me a little though (which I should have brought up when `readAccessor` was happening in the first place) is the format of passing the accessor data from readAccessor as an object with "Cesium-ready data" and "type as a string."

Not sure if it should be in this PR, but we should probably discuss if there's a cleaner way of doing this  before this format becomes disliked but endemic. At the very least I think we should consider emulating enums or something, like with the WebGL constants.

pjcozzi:
> Not sure if it should be in this PR, but we should probably discuss if there's a cleaner way of doing this before this format becomes disliked but endemic. At the very least I think we should consider emulating enums or something, like with the WebGL constants.

An enum would be fine here and is consistent with how things are handled elsewhere in glTF and Cesium.

Do you have other suggested changes?

likangning93:
A thought: would it be more consistent to pass a `results` array into `readAccessor` for the actual data and have `readAccessor` return an enum with the data type in terms of WebGL constants?

JoshuaStorm:
>A thought: would it be more consistent to pass a results array into readAccessor for the actual data and have readAccessor return an enum with the data type in terms of WebGL constants?

This sounds best to me personally. I can do that in this PR if @pjcozzi agrees.

pjcozzi:
OK

JoshuaStorm:
@likangning93 This is good for another look!

likangning93:
Looks good to me.

PR #124, opened 2016-07-01
Uniform grid for AO

likangning93:
This is ready for a first look! We should try to merge [ao to vertices #113](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/113) before anyone takes a long look at changes in `bakeAmbientOcclusion`, but the grid itself (`StaticUniformGrid` and `buildTriangleStaticUniformGrid`) is ready for review.


likangning93:
Some preliminary benchmarks: running
`time node ./bin/gltf-pipeline.js ./CesiumMan.gltf --ao.resolution=256 --ao.samples=36 --ao.rayDepth=0.2 --ao.diffuse -o cesiumMan_AO.gltf` on `ao-master` takes about 7 minutes 18 seconds while running it with the uniform grid takes about 2 minutes 30 seconds, with the size of the model and the ray distance leading to a 2x8x6 grid (CesiumMan's pretty small).

Here are the models for comparison. They look pretty awful because of how low res the AO computation is and because triangles aren't conservatively texelized when computing AO, but at least they look similar and they both do look like they have AO. Things look a lot better when baking to vertices, which is more important at this point. Most importantly, the grid definitely improves performance! More comparisons with different size grids to come.

[CesiumMan_AO_tex.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/344024/CesiumMan_AO_tex.zip)

lib/bakeAmbientOcclusion.js
```diff
@@ -434,26 +437,34 @@ function computeAmbientOcclusionAt(position, normal, numberSamples, sqrtNumberSa
...
-    for (var triangleSoupIndex = 0; triangleSoupIndex < triangleCount; triangleSoupIndex++) {
-        var positions = triangleSoup[triangleSoupIndex].positions;
-        var distance = rayTriangle(ray, positions[0], positions[1], positions[2], false);
-        if (defined(distance) && distance > nearCull) {
-            minIntersect = Math.min(distance, minIntersect);
-        }
+function raytraceNeighborFunction(positions, parameters) {
+    var distance = rayTriangle(parameters.ray, positions[0], positions[1], positions[2], false);
+    if (defined(distance) && distance > parameters.nearCull) {
+        parameters.minIntersect = Math.min(distance, parameters.minIntersect);
```
likangning93: I just realized: when looking in a grid cell, we don't even need to look at every single triangle. Anything this does after detecting a `distance` less than the max ray distance is wasted!

likangning93: I tried tweaking this on another [branch](https://github.com/likangning93/gltf-pipeline/tree/uniformGridDev). Some comparison numbers for the [2 cylinder engine model](https://github.com/KhronosGroup/glTF/tree/master/sampleModels/2CylinderEngine), which is pretty hefty:

`time node ./bin/gltf-pipeline.js ./2CylinderEngine.gltf --ao.quality=low --ao.ambientShadowContribition=1.0 --ao.rayDistance=5.0 --ao.triangleCentersOnly -o engine_AO_centers_5.gltf`
- 121496 triangles in a 149 x 55 x 54 grid
- 8 minutes and 50 seconds without the early return optimization
- 7 minutes 34 seconds with

`time node ./bin/gltf-pipeline.js ./2CylinderEngine.gltf --ao.quality=low --ao.ambientShadowContribition=1.0 --ao.rayDistance=10.0 --ao.groundPlane --ao.triangleCentersOnly -o engine_AO_centers_10_GP.gltf`
- 79 x 28 x 31 grid
- 17 minutes 13 seconds with early return
- 19 minutes 58 seconds without

likangning93: The version I did that benchmark with had some logic bugs and also wasn't actually doing a full early return, it was still checking each voxel. Updated numbers:

`time node ./bin/gltf-pipeline.js ./2CylinderEngine.gltf --ao.quality=low --ao.ambientShadowContribition=1.0 --ao.rayDistance=5.0 --ao.triangleCentersOnly -o engine_AO_centers_5.gltf`
- 121496 triangles in a 149 x 55 x 54 grid
- 5 minutes 17 seconds with early return
- compare to 8 minutes 50 seconds without

`time node ./bin/gltf-pipeline.js ./2CylinderEngine.gltf --ao.quality=low --ao.ambientShadowContribition=1.0 --ao.rayDistance=10.0 --ao.groundPlane --ao.triangleCentersOnly -o engine_AO_centers_10_GP.gltf`
- 79 x 28 x 31 grid
- 11 minutes 4 seconds with early return
- compare to 19 minutes 58 seconds without

Here are the models. I went ahead and set the shaders to show all AO.
[2CylinderEngine.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/344326/2CylinderEngine.zip)

likangning93:
Updated. CesiumMan only improves to 2 minutes 4 seconds.

pjcozzi:
Great news that the early return is a solid performance win.

In general, are you only seeing 4x performance over no grid?  I thought early results showed 10-100x?

Have you tried graphing the performance of different grid sizes to see what is optimal?  The size of triangles will also impact it, e.g., a wall for a building is very different than an aircraft's engine.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+ *                   This function should expect to take an object and the bounding cell's `min` and cell width. It should
+ *                   return `true` if the object intersects with the cell at all and `false` otherwise.
+ * @param {Number} cellWidth, The cell width of the uniform grid.
+ */
+
+var minScratch = new Cartesian3();
+var maxScratch = new Cartesian3();
+
+function StaticUniformGrid(data, cellWidth, objectAABBFunction, cellCheckFunction) {
+
```
pjcozzi: Minor, remove whitespace here and in general for the first line of a function.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+     * The width of a single cell.
+     * @type {Number}
+     */
+    this.cellWidth = cellWidth;
+
+    /**
+     * The axis aligned bounding box of the grid
+     * @type {AxisAlignedBoundingBox}
+     */
+    this.AABB = new AxisAlignedBoundingBox();
```
pjcozzi: Spell it out; in general, do not use abbreviations in the "public" API.

See https://github.com/AnalyticalGraphicsInc/cesium/tree/master/Documentation/Contributors/CodingGuide#naming

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+     * Array of objects
+     * @type {Array}
+     */
+    this.data = [];
+
+    ////// Populate //////
+    var min = this.AABB.minimum;
+    var max = this.AABB.maximum;
+    var center = this.AABB.center;
+    min.x = Number.POSITIVE_INFINITY;
```
pjcozzi: Why not set this when constructing `AABB` above using the constructor function?

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+    max.x = center.x + (resolution.x / 2.0) * cellWidth;
+    max.y = center.y + (resolution.y / 2.0) * cellWidth;
+    max.z = center.z + (resolution.z / 2.0) * cellWidth;
+
+    var cellCount = resolution.x * resolution.y * resolution.z;
+
+    // Bin the objects
+    var allCellData = [];
+    for (i = 0; i < cellCount; i++) {
+        allCellData.push({
```
pjcozzi: The objects in this array are used a lot, right?  Instead of an object literal, make a constructor function for a new type for better performance.

See https://www.youtube.com/watch?v=XAqIpGU8ZZk

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+        maxScratch.y = Number.NEGATIVE_INFINITY;
+        maxScratch.z = Number.NEGATIVE_INFINITY;
+        objectAABBFunction(item, minScratch, maxScratch);
+
+        // Step over the cells in the AABB, checking for each cell if the object intersects this cell
+        parameters.item = item;
+        StaticUniformGrid.boundingBoxMarch(this, minScratch, maxScratch, binObject, parameters);
+    }
+
+    // Store all the object copies in one contiguous array for better spatial locality
```
pjcozzi: I'm not sure that this is true; it depends on access patterns.  You would want a space fitting curve here.  Could be interesting future work.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+ * Function for executing a function on each cell touched by an axis aligned bounding box within the uniform grid.
+ *
+ * @param {StaticUniformGrid} [grid] A StaticUniformGrid to be marched over
+ * @param {Cartesian3} [min] The minimum coordinate of the AABB.
+ * @param {Cartesian3} [max] The maximum coordinate of the AABB.
+ * @param {Function} [marchFunction] A function to apply at each cell. The function should expect as inputs its parameters, the cell's minimum coordinate, and the cell width.
+ * @param {Object} [parameters] Parameters for the marchFunction
+ */
+
+var cellMarchScratch = new Cartesian3();
```
pjcozzi: Move this above the doc.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+StaticUniformGrid.boundingBoxMarch = function(grid, min, max, marchFunction, parameters) {
+    var stepWidth = grid.cellWidth;
+    var gridMin = grid.AABB.minimum;
+
+    // Compute the minimum coordinate of the first cell
+    var xStart = Math.floor((min.x - gridMin.x) / stepWidth) * stepWidth + gridMin.x;
+    var yStart = Math.floor((min.y - gridMin.y) / stepWidth) * stepWidth + gridMin.y;
+    cellMarchScratch.z = Math.floor((min.z - gridMin.z) / stepWidth) * stepWidth + gridMin.z;
+
+
```
pjcozzi: Remove extra whitespace.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+    // Compute the number of cells that min and max cover in each dimension.
+    var xCount = Math.floor((max.x - min.x) / stepWidth) + 1;
+    var yCount = Math.floor((max.y - min.y) / stepWidth) + 1;
+    var zCount = Math.floor((max.z - min.z) / stepWidth) + 1;
+
+    // March over the cells that the grid covers.
+    for (var z = 0; z < zCount; z++) {
```
pjcozzi: In practice, I don't know if it matters anymore, but I always write loops with, for example, `++z` so it saves a potential copy.

lib/buildTriangleStaticUniformGrid.js
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Function for generating a uniform grid from a list of triangles of form [Cartesian3, Cartesian3, Cartesian3]
+ *
+ * @param {List[Cartesian3, Cartesian3, Cartesian3]} [data] Triangle data to bin.
+ * @param {Number} [cellWidth] The width of each cell in the uniform grid.
+ *
+ * @returns a StaticUniformGrid storing triangles.
+ */
+
+function buildTriangleStaticUniformGrid(data, cellWidth) {
```
pjcozzi: Replace this with a function like `StaticUniformGrid.fromTriangleSoup`.

See https://github.com/AnalyticalGraphicsInc/cesium/tree/master/Documentation/Contributors/CodingGuide#from-constructors

lib/buildTriangleStaticUniformGrid.js
```diff
@@ -0,0 +1,71 @@
...
+    max.x = Math.max(triangle[2].x, max.x);
+    max.y = Math.max(triangle[2].y, max.y);
+    max.z = Math.max(triangle[2].z, max.z);
+}
+
+var triangleCheckScratch = new Cartesian3();
+var halfDimensionsScratch = new Cartesian3();
+var aabbScratch = new AxisAlignedBoundingBox();
+function triangleCellCheck(triangle, cellMin, cellWidth) {
+    var halfWidth = cellWidth / 2.0;
```
pjcozzi: It is also an old school optimization to do `* 0.5` instead of `/ 2.0` since divide units can't be pipelined as efficiently as multipliers.  In practice, it might not matter today.

lib/buildTriangleStaticUniformGrid.js
```diff
@@ -0,0 +1,71 @@
...
+    var center = aabbScratch.center;
+    var maximum = aabbScratch.maximum;
+
+    halfDimensionsScratch.x = halfWidth;
+    halfDimensionsScratch.y = halfWidth;
+    halfDimensionsScratch.z = halfWidth;
+
+    triangleCheckScratch.x = cellMin.x + halfWidth;
```
pjcozzi: Here and below can use `Cartesian3` `add` functions, right?

pjcozzi:
Nice test coverage!  100%!

specs/lib/StaticUniformGridSpec.js
```diff
@@ -0,0 +1,189 @@
...
+    function pointDataAABBFunction(point, min, max) {
+        min.x = Math.min(min.x, point[0]);
+        min.y = Math.min(min.y, point[1]);
+        min.z = Math.min(min.z, point[2]);
+        max.x = Math.max(max.x, point[0]);
+        max.y = Math.max(max.y, point[1]);
+        max.z = Math.max(max.z, point[2]);
+    }
+
+    function pointDataCellCheckFunction(point, position, stepWidth) {
```
pjcozzi: In this context, is `stepWidth` the right name?  Why not just `width`?

pjcozzi: Also drop the `Function` suffix here and throughout.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+
+module.exports = StaticUniformGrid;
+
+/**
+ * A Uniform Grid for unmoving objects.
+ * @alias StaticUniformGrid
+ * @constructor
+ *
+ * @param {Object[]} objects Objects that the grid should store.
+ * @param {Function} aabbFunction A function for checking the axis aligned bounding box of an object in space.
```
pjcozzi: For an example of documenting callbacks, see https://github.com/AnalyticalGraphicsInc/cesium/blob/03191f3c7558d8ab6b3fe6fb002261501a2d9eb6/Source/DataSources/CallbackProperty.js#L106-L113

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+var minScratch = new Cartesian3();
+var maxScratch = new Cartesian3();
+
+function StaticUniformGrid(data, cellWidth, objectAABBFunction, cellCheckFunction) {
+
+    /**
+     * The width of a single cell.
+     * @type {Number}
+     */
+    this.cellWidth = cellWidth;
```
pjcozzi: Most of these properties are immutable, right?  They are set once and then can't be changed later.

Use property getters, https://github.com/AnalyticalGraphicsInc/cesium/tree/master/Documentation/Contributors/CodingGuide#property-gettersetters

Inside this class, it is, of course, OK to access the variable directly.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    // Figure out what the grid's resolution should be. Pad min and max out to match
+    resolution.x = Math.floor((max.x - min.x) / cellWidth) + 1;
+    resolution.y = Math.floor((max.y - min.y) / cellWidth) + 1;
+    resolution.z = Math.floor((max.z - min.z) / cellWidth) + 1;
+
+    center = Cartesian3.add(min, max, center);
+    center = Cartesian3.divideByScalar(center, 2.0, center);
+    min.x = center.x - (resolution.x / 2.0) * cellWidth;
```
pjcozzi: This could use `Cartesian3` `add` functions, right?

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,277 @@
...
+};
+
+function index3DTo1D(x, y, z, resolution){
+    return x + y * resolution.x + z * resolution.x * resolution.y;
+}
+
+StaticUniformGrid.indexOfPosition = function(grid, position) {
+    var min = grid.AABB.minimum;
+    var cellWidth = grid.cellWidth;
+    var x = Math.floor((position.x - min.x) / cellWidth);
```
pjcozzi: Store `1.0 / cellWidth` as a private member then multiply here.

pjcozzi:
@likangning93 great start here; very impressive how fast you put this together!

likangning93:
Okay. I think this is ready for another look!

pjcozzi:
@likangning93 did you see this question: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/124#issuecomment-230122083

likangning93:
Ah. I will try running some benchmarks now.

I think CesiumMan's performance improvement was relatively low because the grid computed for it was tiny, just 2 x 8 x 6. I think performance generally improves when the grid is higher resolution, which runs straight into the problem where the ray depth is coupled to cell size.

I suspect this won't be a huge problem for buildings if we want to compute AO at building-like scale, since extreme ray dropoff distance would make them look like toys. But it is something that's pretty high priority on the roadmap.

likangning93:
Some preliminary benchmarks, again with the building model and statue of liberty model:

Building:
building charts

| model name | parameters | AABB x | AABB y | AABB z | total triangles |  |  |  |  |  |  |  |  |  |  |  |  |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| building | (default) | 175.3351974 | 100.4550018 | 128.6880951 | 5521 |  |  |  |  |  |  |  |  |  |  |  |  |
|  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| ray distance | grid x | grid y | grid z | total cells | max/cell | min/cell | avg/cell | triangle clone count | % increase | t to build grid 1 | t build grid 2 | t build grid 3 | avg time for grid | total time 1 | total time 2 | total time 3 | average total |
| 2 | 92 | 51 | 69 | 323748 | 26 | 0 | 0.20548 | 66523 | 1104.908531 | 1.259 | 1.187 | 1.187 | 1.211 | 7.474 | 7.458 | 7.54 | 7.491 |
| 5 | 40 | 21 | 30 | 25200 | 60 | 0 | 0.82563 | 20806 | 276.8520196 | 0.434 | 0.375 | 0.391 | 0.400 | 11.368 | 11.011 | 11.136 | 11.172 |
| 10 | 22 | 11 | 17 | 4114 | 103 | 0 | 2.84541 | 11706 | 112.0268067 | 0.219 | 0.234 | 0.234 | 0.229 | 24.739 | 24.819 | 24.063 | 24.540 |
| 15 | 16 | 7 | 13 | 1456 | 181 | 0 | 6.50343 | 9469 | 71.50878464 | 0.187 | 0.188 | 0.187 | 0.187 | 46.391 | 46.77 | 44.93 | 46.030 |
| 20 | 13 | 6 | 11 | 858 | 291 | 0 | 9.20163 | 7895 | 42.99945662 | 0.166 | 0.156 | 0.156 | 0.159 | 67.375 | 66.348 | 67.587 | 67.103 |
| 25 | 12 | 5 | 10 | 600 | 371 | 0 | 12.14666 | 7288 | 32.00507155 | 0.156 | 0.156 | 0.156 | 0.156 | 92.201 | 95.623 | 95.494 | 94.439 |
| 30 | 10 | 4 | 9 | 360 | 367 | 0 | 19.31666 | 6954 | 25.95544285 | 0.174 | 0.14 | 0.157 | 0.157 | 123.192 | 118.6 | 124.079 | 121.957 |

Statue:
statue charts

| model name | parameters | AABB x | AABB y | AABB z | total triangles |  |  |  |  |  |  |  |  |  |  |  |  |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| statue | (default) | 16.98472977 | 46.65739822 | 14.47532988 |  |  |  | 8846 |  |  |  |  |  |  |  |  |  |
|  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| ray distance | grid x | grid y | grid z | total cells | max/cell | min/cell | avg/cell | triangle clone count | % increase | t to build grid 1 | t build grid 2 | t to build grid 3 | avg time for grid | total time 1 | total time 2 | total time 3 | average total |
| 2 | 13 | 24 | 12 | 3744 | 384 | 0 | 2.624 | 9825 | 11.06714899 | 0.203 | 0.188 | 0.187 | 0.193 | 8.182 | 8.363 | 7.908 | 8.151 |
| 5 | 8 | 10 | 7 | 560 | 1668 | 0 | 15.975 | 8946 | 1.130454443 | 0.172 | 0.156 | 0.156 | 0.161 | 18.454 | 17.986 | 18.338 | 18.259 |
| 10 | 6 | 5 | 6 | 180 | 1904 | 0 | 49.34 | 8882 | 0.406963599 | 0.156 | 0.172 | 0.157 | 0.162 | 32.299 | 31.298 | 31.023 | 31.540 |
| 15 | 6 | 4 | 5 | 120 | 2172 | 0 | 74.05 | 8886 | 0.452181777 | 0.156 | 0.157 | 0.18 | 0.164 | 41.713 | 38.388 | 40.054 | 40.052 |
| 20 | 5 | 3 | 5 | 75 | 3949 | 0 | 118.4 | 8880 | 0.384354511 | 0.188 | 0.172 | 0.172 | 0.177 | 47.961 | 46.998 | 48.156 | 47.705 |
| 25 | 5 | 2 | 5 | 50 | 5556 | 0 | 177.6 | 8880 | 0.384354511 | 0.158 | 0.157 | 0.157 | 0.157 | 54.864 | 52.942 | 55.257 | 54.354 |
| 30 | 5 | 2 | 5 | 50 | 5556 | 0 | 177.6 | 8880 | 0.384354511 | 0.156 | 0.159 | 0.191 | 0.169 | 55.556 | 63.934 | 56.193 | 58.561 |

likangning93:
Benchmarking Notes:

- Everything was run with "default" settings except the ray distance. Since this meant density defaulted to 1 sample per meter, this means the Statue of Liberty had considerably fewer samples since it has very few triangles that are larger than 1 meter in any dimension. That's part of why it was so much faster.
- Generally, a higher resolution grid seems to lead to less processing time but increased grid construction time, but the construction time is swallowed by the raytracing time for all ray depths/cell widths tested.
- Generally, more cells -> more cloned triangles. I actually ran into Node's memory limit with small cells on the [2 cylinder engine](https://github.com/KhronosGroup/glTF/tree/master/sampleModels/2CylinderEngine), which is pretty heavily tesselated, but the small cell size would have led to poor AO results anyway and using a bigger cell size got around the problem.
- The statue of liberty's grid cell size hits an upper limit. This is probably since it was being rendered with a ground plane, which is a set of "virtual triangles" inserted into the uniform grid. I think increasing the ray distance further would eventually lead to something like a 3 x 1 x 3 grid where all the statue's triangles are in the center cell.

likangning93:
Here's the building at various levels of occlusion:
building_02
building_05
building_10
building_15
building_20
building_25
building_30

likangning93:
The statue starts to look strange at any depth greater than 2 meters. I suspect part of this is the normals problem. I'll go ahead and add that to the roadmap.

statue_02
statue_05

likangning93:
I had another thought: ray-intersection tests for AO isn't true raycasting, but it also isn't just neighbor search. We don't have to check the ray against 27 voxels, just against the 8 in the octant that the ray points out in. I'll give this a try and post some numbers.

likangning93:
This is ready for another look!

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,477 @@
...
+
+    center = Cartesian3.add(min, max, center);
+    center = Cartesian3.divideByScalar(center, 2.0, center);
+
+    halfDimensionsScratch.x = (resolution.x * 0.5) * cellWidth;
+    halfDimensionsScratch.y = (resolution.y * 0.5) * cellWidth;
+    halfDimensionsScratch.z = (resolution.z * 0.5) * cellWidth;
+
+    min = Cartesian3.subtract(center, halfDimensionsScratch, min);
+    max = Cartesian3.add(center, halfDimensionsScratch, max);
```
lilleyse: These vars aren't used later, so do you need these two lines?

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,477 @@
...
+    var center = axisAlignedBoundingBoxScratch.center;
+    var maximum = axisAlignedBoundingBoxScratch.maximum;
+
+    halfDimensionsScratch.x = halfWidth;
+    halfDimensionsScratch.y = halfWidth;
+    halfDimensionsScratch.z = halfWidth;
+
+    axisAlignedBoundingBoxScratch.minimum = cellMin;
+    center = Cartesian3.add(cellMin, halfDimensionsScratch, center);
+    maximum = Cartesian3.add(center, halfDimensionsScratch, maximum);
```
lilleyse: Similar comment for the `center` and `maxmimum` lines which aren't used later.

lilleyse:
Is the `--ao.groundPlane` option working for you right now?

lib/StaticUniformGrid.js
likangning93: This was for updating the grid AABB's internal minimum and maximum, but it'll be more clear as
` this._axisAlignedBoundingBox.minimum = ...`

lib/StaticUniformGrid.js
lilleyse: Ah same reason as above I'm guessing.

likangning93:
`--ao.groundPlane` is working for me, but depending on the model it might not show much occlusion. What model are you using?

lilleyse:
I'm using the plane, but the result doesn't look the same as the screenshot I posted in #113.

likangning93:
It looks like a [yargs](https://www.npmjs.com/package/yargs)-ism, if you just pass the flag with `flat.otherOptions` it will lose any options attached. I'll update the documentation to indicate this.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,477 @@
+'use strict';
+var Cesium = require('cesium');
+var AxisAlignedBoundingBox = Cesium.AxisAlignedBoundingBox;
+var Cartesian3 = Cesium.Cartesian3;
+var defineProperties = Cesium.defineProperties;
+var clone = require('clone');
+var triangleAxisAlignedBoundingBoxOverlap = require('./triangleAxisAlignedBoundingBoxOverlap');
```
lilleyse: When we eventually merge `ao-master` and `master` we should clean up the code to follow the Node guide, like cleaning up the imports and converting applicable callbacks and third-party functions to promises.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,477 @@
...
+
+module.exports = StaticUniformGrid;
+
+/**
+ * A Uniform Grid for unmoving objects.
+ * @alias StaticUniformGrid
+ * @constructor
+ *
+ * @param {Object[]} objects Objects that the grid should store.
+ * @param {Number} cellWidth, The cell width of the uniform grid.
```
lilleyse: Remove the comma.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,477 @@
...
+ * @param {StaticUniformGrid~CompareObjectBoundingBox} compareObjectBoundingBox
+ * @param {StaticUniformGrid~CellCheckFunction} cellCheck
+ */
+
+var minScratch = new Cartesian3();
+var maxScratch = new Cartesian3();
+var halfDimensionsScratch = new Cartesian3();
+
+function StaticUniformGrid(data, cellWidth, compareObjectBoundingBox, cellCheck) {
```
lilleyse: Throughout this file `data`, `objects`, and `items` are used interchangeably. I would pick one name to use.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,477 @@
...
+
+var cellMarchScratch = new Cartesian3();
+/**
+ * Function for executing a function on each cell touched by an axis aligned bounding box within the uniform grid.
+ *
+ * @param {StaticUniformGrid} [grid] A StaticUniformGrid to be marched over
+ * @param {Cartesian3} [min] The minimum coordinate of the bounding box to be "marched" over.
+ * @param {Cartesian3} [max] The maximum coordinate of the bounding box to be "marched" over.
+ * @param {Function} [marchFunction] A function to apply at each cell. The function should expect as inputs its parameters, the cell's minimum coordinate, and the cell width.
+ * @param {Object} [parameters] Parameters for the marchFunction
```
lilleyse: All these options are required, so remove the square brackets. (Also in other areas in this file)

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,477 @@
...
+    }
+};
+
+/**
+ * Function for generating a uniform grid from a list of triangles of form [Cartesian3, Cartesian3, Cartesian3]
+ *
+ * @param {Object[]} [data] Triangle data to bin. Each triangle should be of form [Cartesian3, Cartesian3, Cartesian3]
+ * @param {Number} [cellWidth] The width of each cell in the uniform grid.
+ *
+ * @returns a new StaticUniformGrid storing triangles.
```
lilleyse: Add the type of the return object to the doc.

lib/StaticUniformGrid.js
```diff
@@ -0,0 +1,477 @@
...
+ */
+
+/**
+ * A function that checks whether or not some type of geometry or other spatial data intersects a uniform grid cell.
+ * @callback StaticUniformGrid~CellCheckFunction
+ *
+ * @param {Object} item An object that should be checked against.
+ * @param {Cartesian3} corner The minimum corner of the cell.
+ * @param {Number} stepWidth The cell width. Uniform in all 3 directions.
+ * @returns {Boolean} True if the item intersects the cell, false otherwise.
```
lilleyse: Add the param name here.

lib/bakeAmbientOcclusion.js
```diff
@@ -1236,26 +1242,37 @@ function computeAmbientOcclusionAt(options) {
...
-        var positions = triangleSoup[triangleSoupIndex];
-        var distance = IntersectionTests.rayTriangleParametric(ray, positions[0], positions[1], positions[2], false);
-        if (defined(distance) && distance > nearCull) {
-            minIntersect = Math.min(distance, minIntersect);
-        }
-    }
-    return minIntersect;
+function raytraceNeighborFunction(positions, parameters) {
+    var distance = IntersectionTests.rayTriangleParametric(parameters.ray, positions[0], positions[1], positions[2], false);
+    parameters.hasIntersect = defined(distance) && distance > parameters.nearCull && distance < parameters.farCull;
```
lilleyse: Add some parentheses around each part so its cleaner to read.

lilleyse:
Heh most of my suggestions are just doc related. The good looks good, and I especially like how generic it is. I'm going to run through some models quickly but it seems good to me.

lilleyse:
I noticed an issue when running a random model through: 
[shovel.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/360315/shovel.gltf.txt) (Ignore the texture/geometry issues)

` throw new DeveloperError('Could not find attribute a_texcoord0 or corresponding varying in technique ' + techniqueIds[i]);`

I think hardcoding `texcoord0` is unsafe, and instead you should find the name of the attribute associated with the `TEXCOORD` or `TEXCOORD_0` semantic. In this case the attribute is `a_texcoord_0`, which is why it failed.

pjcozzi:
> When we eventually merge ao-master and master we should clean up the code to follow the Node guide, like cleaning up the imports and converting applicable callbacks and third-party functions to promises.

I'd like to get `ao-master` merged into `master` (and deployed) ASAP.  This can be part of the cleanup roadmap.

likangning93:
Handling that wasn't too bad, although I guess there's no guarantee that the first `TEXCOORD` parameter is actually the diffuse texture one. Updated!

likangning93:
Actually, it would be cleaner and more robust to just extract the `texture2D` call with the diffuse uniform. I'll try this in the morning, the function for this should probably get its own unit test too.

likangning93:
Okay: I updated this so that the shader modification code now locates the instruction containing the appropriate `texture2D` call, finds the diffuse color assignee, and saves a copy of that for the AO blending later.

```
precision highp float;
varying vec3 v_normal;
uniform vec4 u_ambient;
varying vec2 v_texcoord0;
uniform sampler2D u_diffuse;
vec3 diffuseForAmbientOcclusion; // added by AO stage

//...

void mainBeforeAO() {
//...
    diffuse = texture2D(u_diffuse, v_texcoord0);
    diffuseForAmbientOcclusion = vec3(diffuse); // added by AO stage
//...
}
varying float v_ambientOcclusion; 

void main() 
{ 
    mainBeforeAO(); 
    gl_FragColor.rgb = mix(gl_FragColor.rgb, gl_FragColor.a * v_ambientOcclusion * diffuseForAmbientOcclusion.rgb, 0.5); 
}
```

This still isn't 100% robust since someone could write a shader that does something weird to the `texture2D` result and packs it into a `vec2` or something, but it should work fine for if `diffuse` here is a vec3 or a vec4, which should be almost all cases. This should also get around problems with compressed texture coordinates. Final bonus, now having AO doesn't needlessly increase texture lookups.

likangning93:
Updated. I don't have a whole lot of experience with regexes, so I added some specs to be sure that they're doing what I think they're doing.

lilleyse:
At this point now that we can identify the line that gets the diffuse color, could we just blend with ao right after and not even need `mainBeforeAO`?

likangning93:
I think it's more robust to use the AO to modify the existing lighting after that computation is completely done, since then it can raise overly dark areas and darken well-lit areas that should be occluded. I actually didn't look very deeply into how to incorporate the AO, but replace/blend/multiply seem like they would work with virtually any shader.

lilleyse:
Okay that makes sense, we can keep it how it is.

lilleyse:
Ok the new code looks good and I tested a few different cases.

lilleyse:
To finish AO up, open a PR of `ao-master` into `master`. Reference the node guide to convert callbacks to promises, clean up imports, etc. We can then do some final testing once that's all done.

pjcozzi:
Congrats on this merge, @likangning93!  Great job @lilleyse @likangning93!

PR #127, opened 2016-07-05
Add --cesium flag for exporting a model with the Cesium sun as a light source

### LINKED ISSUE #114: Undefined variable "czm_sunDirectionEC"
I tried to convert a simple OBJ model to glTF, using https://github.com/AnalyticalGraphicsInc/OBJ2GLTF

The resulting fragment shader code contained a line 

    vec3 l = normalize(czm_sunDirectionEC);

but the variable `czm_sunDirectionEC` has not been declared at this point. After a bit of tracing and backward search, it appears that this variable is inserted in the shader by the gltf-pipeline, here:

https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/modelMaterialsCommon.js#L417

(At least this variable name appears nowhere else, and I can't imagine how the shader that is constructed there should be valid when this line is inserted)

lilleyse:
Fixes #114 

Added a command line flag `--cesium` for optimizing a model for use in Cesium. Right now this just means using `czm_sunDirectionEC` in the shader, but there could be other use cases for a having a Cesium-only pipeline. By default if a model does not have lights/shaders and goes through `modelMaterialCommon`, it will use a light source that is always pointing in the direction of the camera, which is the same behavior as COLLADA2GLTF. The images below show with the flag set vs. not set.

capture3
capture4

### DIFF
--- a/bin/gltf-pipeline.js
+++ b/bin/gltf-pipeline.js
@@ -17,7 +17,8 @@ if (process.argv.length < 3 || defined(argv.h) || defined(argv.help)) {
         '  -t --separateImage, write out separate textures, but embed geometry/animation data files, and shader files.\n' +
         '  -q --quantize, quantize the attributes of this model.\n' +
         '  -n --encodeNormals, oct-encode the normals of this model.\n' +
-        '  -c --compressTextureCoordinates, compress the texture coordinates of this model.\n';
+        '  -c --compressTextureCoordinates, compress the texture coordinates of this model.\n' +
+        '     --cesium, optimize the glTF for Cesium by using the sun as a default light source. \n';
     process.stdout.write(help);
     return;
 }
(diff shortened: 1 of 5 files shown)


lilleyse:
I need to fix some failing tests quickly.

lilleyse:
@pjcozzi Do you think this should be a checkbox in the online converter?

lilleyse:
Updated.

pjcozzi:
> @pjcozzi Do you think this should be a checkbox in the online converter?

No, at least not yet.  I don't want to expose too many options and complicate the interface until there is strong demand for a feature.

pjcozzi:
Are you adding more unit tests for this?

lilleyse:
> Are you adding more unit tests for this?

Ah yes I should.

lilleyse:
Added a test.

PR #128, opened 2016-07-06
Apply alpha blending to transparent models

### LINKED ISSUE #78: Handle transparent images
The pipeline should be able to analyze an image to check if it has transparency, and if so edit the techniques for alpha blending. In the obj2gltf project I did a simple approach for this here: https://github.com/AnalyticalGraphicsInc/OBJ2GLTF/blob/master/lib/image.js. However there are also many pngs that have alpha channels but don't actually use them, so this approach wouldn't work.

lilleyse:
Fixes #78 

If a material has a diffuse color with alpha < 1.0 or a diffuse texture with any translucent pixels, then alpha blending is automatically applied to the technique. This works for the `modelMaterialCommon` path, as well as regular materials.

Other changes:
* I removed the `imageProcess` flag and just always decode images, which makes it easy to check later if an image is transparent or not. Otherwise the image processing pipeline works the same as before.
* I updated `modelMaterialsCommon` to not save `transparent` and `doubleSided` in the generated material because these are never used.
* Tweaked a lot of tests to work with the new setup, but mostly in minor ways.

I tested the changes with all the different configurations:
* opaque color box
* opaque textured box
* opaque color box (no technqiues/shaders)
* opaque textured box (no techniques/shaders)
* transparent color box
* transparent textured box
* transparent color box (no technqiues/shaders)
* transparent textured box (no techniques/shaders)

[box_transparent.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/350706/box_transparent.zip)


lilleyse:
The coverage went way down, I'll look into that.

lilleyse:
Ok better now.

pjcozzi:
@lasalvavida do you want to do a first review of this, and then I'll take a quick look?

pjcozzi:
@lasalvavida do you have time to look at this?

lasalvavida:
Yes, I'll take a look today.

lasalvavida:
This all looks good to me.

Tested by making the milktruck transparent.

milktruck-transparent

Looks awesome, thanks @lilleyse!

pjcozzi:
Thanks for the review, @lasalvavida!

PR #129, opened 2016-07-08
Cleaning up to get ready for publishing to npm

lilleyse:
@mramato Would you like to take a quick look here?


TypeScriptDefinitions/async.d.ts
```diff
@@ -0,0 +1,165 @@
+// Type definitions for Async 1.4.2
```
pjcozzi: What is this for?  Part of the `async` package?  Does it need to be in this location?

pjcozzi: Was this directory supposed to be submitted?

pjcozzi:
@lilleyse tangentially related: did you create a roadmap issue for the other cleanup for gltf-pipeline?

TypeScriptDefinitions/async.d.ts
lilleyse: I saw that it was included in some of our other repos, @mramato should it be included?

mramato: Yes, for now this directory should be submitted.  These are TypeScript definition files that aid Webstorm in intellisense and error checking and get updated when you update dependencies in package.json.  They are discussed in my upcoming node guide.  I'm hoping to improve things in the long term so that these will be handled slightly differently (and possibly not submitted) but for now this is the same thing we do in all of our other modules.

mramato:
@lilleyse I updated my guide to flesh out the publishing section some more: https://github.com/AnalyticalGraphicsInc/agi-cesium-people/tree/node-coding-guide/NodeCodingGuide#publishing-and-versioning-your-module  It now has a section on the "first time" you publish.

lilleyse:
> @lilleyse tangentially related: did you create a roadmap issue for the other cleanup for gltf-pipeline?

Not yet, I do have a running list of issues that I can create an issue for soon.

lilleyse:
> @lilleyse I updated my guide to flesh out the publishing section some more:

Cool thanks, I'll check it out.

lilleyse:
Are there any other suggestions for this?

specs/lib/addPipelineExtrasSpec.js
```diff
@@ -8,6 +8,9 @@ describe('addPipelineExtras', function() {
 
     beforeAll(function(done) {
         fs.readFile(gltfExtrasPath, function(err, data) {
+            if (err) {
```
mramato: Use `defined` here (though I know this will eventually go away when everything is promisified.

mramato:
Are we planning on publishing this with callbacks and the breaking backwards compatibility when we refactor for promises?  Refactoring should be pretty easy and will clean up a lot of the boilerplate code, so I would recommend we do it sooner rather than later.

mramato:
That's all the comments I have.

lilleyse:
> Are we planning on publishing this with callbacks and the breaking backwards compatibility when we refactor for promises? Refactoring should be pretty easy and will clean up a lot of the boilerplate code, so I would recommend we do it sooner rather than later.

That's a good question. We still have a lot of cleanup to do #130, but at the same time we want to update the online model converter soon... @pjcozzi thoughts?

mramato:
@lilleyse I believe as long as you keep the version to `x.0.0`, you can break compatibility whenever you want.  Another option would be to publish `0.1.0-alpha1` and then just increment alpha2/alpha3/etc... this is a valid node version and is standard for pre-release libraries (but still allows us to publish to npm and use it as a dependency.

lilleyse:
Okay, `0.1.0-alpha1` seems fitting then.

specs/lib/addPipelineExtrasSpec.js
lilleyse: There are too many areas that we do this right now, I think I'm just going to save this for the cleanup.

pjcozzi:
Update the README.md since the npm scripts changed, e.g., `npm run jasmine` -> `npm run test`

lilleyse:
Updated

pjcozzi:
Nice!

Issue #130, opened 2016-07-11
Cleanup


lilleyse:
Here is a rough list of tasks for bringing this repo up to production. Feel free to edit or add new items to the list:
- [ ] We should have a common set of models that we use for testing new features, especially models that have created problems in the past. However I'm not sure if they should be hosted in this repo.
- [ ] Refactor gltf-pipeline as a series of stages with dependencies
  - https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/99
  - https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/75
  - Use promises instead of callbacks
- [x] General cleanup, api standardizing, file renaming, folder organization
  - [x] Many functions expect an options object. Instead of sending in `{}` in a lot of places, just do it the Cesium way and create the options object if it's not defined with `options = defaultValue(options, defaultValue.EMPTY_OBJECT);`
- [x] Use recommended libraries, #178
- [x] Use JSDoc everywhere
- [ ] Progress events: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/92
- [x] Something like jQuery for glTF to eliminate redundant glTF data structure traversal
- Note: If there are any changes to the api we will need to update [OBJ2GLTF](https://github.com/AnalyticalGraphicsInc/OBJ2GLTF) and the website model converter

lasalvavida:
Added a list of functions that are considered "stages" and should be converted to use promises in #99.

JoshuaStorm:
We use `mkdirp` in a handful of files, which the Node guide encourages be replaced with `fs-extra`'s `ensureDir` and `ensureFile`. Should this be added to the cleanup?

lasalvavida:
> We use mkdirp in a handful of files, which the Node guide encourages be replaced with fs-extra's ensureDir and ensureFile. Should this be added to the cleanup?

Yes

JoshuaStorm:
Oh, I guess that is included in the

>Use libraries recommended in the node guide

my bad

pjcozzi:
Added "Something like jQuery for glTF to eliminate redundant glTF data structure traversal"

lasalvavida:
> We should have a common set of models that we use for testing new features, especially models that have created problems in the past. However I'm not sure if they should be hosted in this repo.

@lilleyse What do you think about include the glTF sampleModels as a git submodule?

lilleyse:
[...]

PR #131, opened 2016-07-11
Replace mkdirp with fs-extra usage

JoshuaStorm:
As a part of #130

### DIFF
--- a/lib/writeBinaryGltf.js
+++ b/lib/writeBinaryGltf.js
@@ -1,8 +1,7 @@
 'use strict';
-var fs = require('fs');
+var fse = require('fs-extra');
 var path = require('path');
 var async = require('async');
-var mkdirp = require('mkdirp');
 var getBinaryGltf = require('./getBinaryGltf');
 var writeSource = require('./writeSource');
 var Cesium = require('cesium');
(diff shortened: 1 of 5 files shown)


lib/writeGltf.js
```diff
@@ -24,7 +23,7 @@ function writeGltf(gltf, outputPath, embed, embedImage, createDirectory, callbac
     if (createDirectory) {
         outputPath = path.join(path.dirname(outputPath), 'output', path.basename(outputPath));
     }
-    mkdirp.sync(path.dirname(outputPath));
+    fse.ensureDirSync(path.dirname(outputPath));
```
lilleyse: You don't need this because `outputFile` will create a new directory if it doesn't exist.

lib/writeBinaryGltf.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
-var fs = require('fs');
+var fse = require('fs-extra');
```
lilleyse: These should be named `fsExtra`.

lib/writeGltf.js
```diff
@@ -38,7 +37,7 @@ function writeGltf(gltf, outputPath, embed, embedImage, createDirectory, callbac
             }
         } else {
             removePipelineExtras(gltf);
-            fs.writeFile(outputPath, JSON.stringify(gltf, undefined, 2), function (err) {
+            fse.outputFile(outputPath, JSON.stringify(gltf, undefined, 2), function (err) {
```
lilleyse: `outputJson` may be a good choice here.

lilleyse:
Looks good!

PR #132, opened 2016-07-11
Added initial JSDocs for pipeline stages

JoshuaStorm:
As a part of #130 

Still need to add docs for 'intermediate' steps of the stages. 

Unsure if I should be using hyphens following the parameter names, they're in the JSDocs documentation, but not in the `quantizeAttributes` that was already documented.

### DIFF
--- a/lib/addDefaults.js
+++ b/lib/addDefaults.js
@@ -10,6 +10,53 @@ var findAccessorMinMax = require('./findAccessorMinMax');
 
 module.exports = addDefaults;
 
+/**
+ * Adds default values to any missing glTF values of the hierarchy.
+ *
+ * @param {Object} gltf - An object holding a glTF hierarchy
+ * @param {Object} [options=undefined] - Defines more specific defaults behavior
+ * @param {String} [options.diffuseTechnique='LAMBERT'] - Specifies the diffuse lighting technique
+ * @param {String} [options.specularTechnique='BLINN'] - Specifies the specular lighting technique
+ *
+ * @returns glTF hierarchy with added default values
+ */
+function addDefaults(gltf, options) {
+    options = defaultValue(options, Cesium.defaultValue.EMPTY_OBJECT);
+
+    if (!defined(gltf)) {
+        return undefined;
+    }
+
+    if (defined(gltf.allExtensions)) {
+        gltf.extensionsUsed = gltf.allExtensions;
+        gltf.allExtensions = undefined;
+    }
+    gltf.extensionsUsed = defaultValue(gltf.extensionsUsed, []);
+
+    accessorDefaults(gltf);
+    animationDefaults(gltf);
+    assetDefaults(gltf);
+    bufferDefaults(gltf);
+    bufferViewDefaults(gltf);
+    cameraDefaults(gltf);
+    imageDefaults(gltf);
+    lightDefaults(gltf);
+    techniqueDefaults(gltf);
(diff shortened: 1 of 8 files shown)


lib/addDefaults.js
```diff
@@ -10,6 +10,53 @@ var findAccessorMinMax = require('./findAccessorMinMax');
 
 module.exports = addDefaults;
 
+/**
+ * Adds default values to any missing glTF values of the hierarchy.
+ *
+ * @param {Object} gltf - An object holding a glTF hierarchy
```
lilleyse: Even though Cesium doesn't use the hyphens I kind of like them here. @pjcozzi do you have a preference?

lib/combineMeshes.js
```diff
@@ -3,8 +3,16 @@ var Cesium = require('cesium');
 var defined = Cesium.defined;
 module.exports = combineMeshes;
 
-//Combines meshes in a node. 
-//Should be run before the combinePrimitives stage to merge all primitives for a node into one mesh.
+
+
```
lilleyse: Remove extra space.

lib/encodeImages.js
```diff
@@ -4,6 +4,12 @@ var Jimp = require('jimp');
 
 module.exports = encodeImages;
 
+/**
+ * Encode images to a glTF buffer
+ *
+ * @param {Object} gltf - An object holding a glTF hierarchy
+ * @param {Function} callback
```
lilleyse: The believe callbacks are documented like this: http://usejsdoc.org/tags-callback.html. At the same time, we will eventually change the callbacks to promises anyways.

lib/removeUnused.js
```diff
@@ -16,8 +16,16 @@ var removeUnusedShaders = require('./removeUnusedShaders');
 
 module.exports = removeUnused;
 
+/**
+ * Removes unused values in a a glTF hierarchy.
+ * 
+ * @param {Object} gltf - An object holding a glTF hierarchy
+ * @param {Object} stats - An OptimizationStatistics Object
```
lilleyse: Instead of {Object} make it {OptimizationStatistics}

lib/removeUnused.js
```diff
@@ -16,8 +16,16 @@ var removeUnusedShaders = require('./removeUnusedShaders');
...
+ * 
+ * @returns glTF without redundant objects
```
lilleyse: This description is a bit confusing.

lib/removeUnusedVertices.js
```diff
@@ -12,7 +12,13 @@ var writeBufferComponentType = require('./writeBufferComponentType');
 
 module.exports = removeUnusedVertices;
 
-//Removes attributes from indexed primitives that aren't used
+/**
+ * Removes attributes from indexed primitives that aren't used
+ *
+ * @param {Object} gltf - An object holding a glTF hierarchy
+ * 
+ * @returns glTF without unnecessary vertices
```
lilleyse: Change unnecessary to unused just to match the stage name.

lib/addDefaults.js
lilleyse: For the sake of consistency with other areas in this project, let's just remove the hyphens.

JoshuaStorm:
I'm going to hold off on this until we have the majority of these converted to promise patterns

lilleyse:
Closed in favor of #157

PR #135, opened 2016-07-13
Async to promises. writeGltf, writeBinaryGltf, writeFile -- callbacks to Promises

JoshuaStorm:
~~Not~~ ready for review.

I am currently switching some files to be `Promise`s instead of using callbacks, but I've hit a few snags which have resulted in me doing some anti-patterns that I wanted to ask a few questions about to resolve.

I'll be adding inline comments so it is easier to see what I am talking about.


specs/lib/writeBuffersSpec.js
```diff
@@ -39,27 +41,39 @@ describe('writeBuffers', function() {
...
-                expect(bufferEqual(outputData, bufferData)).toBe(true);
-                done();
-            });
+        writeGltf(gltf, options)
+            .then(function() {
+                expect(gltf.buffers.CesiumTexturedBoxTest.extras).not.toBeDefined();
+                expect(gltf.buffers.CesiumTexturedBoxTest.uri).toEqual('CesiumTexturedBoxTest.bin');
+                readFile(outputBufferPath)
+                    .then(function(outputData) {
+                        expect(bufferEqual(outputData, bufferData)).toBe(true);
```
JoshuaStorm: I am fairly certain I shouldn't have to be nesting `.then` like this, but I am unsure how I could write this in order to have `outputData` accessible by the second `.then` function


I have attempted the code below, but that results in `outputData` being undefined.
```
writeGltf(gltf, options)
            .then(function() {
                expect(gltf.buffers.CesiumTexturedBoxTest.extras).not.toBeDefined();
                expect(gltf.buffers.CesiumTexturedBoxTest.uri).toEqual('CesiumTexturedBoxTest.bin');
            })
            .then(function() {
                readFile(outputBufferPath);
            })
            .then(function(outputData) {
                expect(bufferEqual(outputData, bufferData)).toBe(true);
                done();
            });
```

lib/gltfPipeline.js
```diff
@@ -117,41 +116,52 @@ function processFile(inputPath, options, callback) {
...
-            outputPath : outputPath,
-            embed : embed,
-            embedImage : embedImage,
-            createDirectory : createDirectory
-        }, callback);
-    } else {
-        writeGltf(gltf, outputPath, embed, embedImage, createDirectory, callback);
-    }
+
+    return new Promise(function() {
```
JoshuaStorm: I don't quite understand when `Promise`s should have a `resolve` parameter. 

I previously had this function take a `resolve` and called `resolve()` at the end of the function, but this seemed to caused `writeFile(...).then` to be triggered preemptively. 

Do promises that ultimately return other promises not need `resolve` (since the last promise on the chain will trigger the resolve?)

specs/lib/writeBuffersSpec.js
lasalvavida: @JoshuaStorm 
You need to return a promise to do chaining like that.

```javascript
writeGltf(gltf, options)
    .then(function() {
        expect(gltf.buffers.CesiumTexturedBoxTest.extras).not.toBeDefined();
        expect(gltf.buffers.CesiumTexturedBoxTest.uri).toEqual('CesiumTexturedBoxTest.bin');
        return readFile(outputBufferPath);
    })
    .then(function(outputData) {
        expect(bufferEqual(outputData, bufferData)).toBe(true);
        done();
    });

File: lib/getBinaryGltf.js [hunk 3]
```diff
@@ -57,57 +58,59 @@ function updateBinaryObject(gltf, pipelineExtras, name, state) {
...
+        header.writeUInt32LE(sceneLength, 12);
+        header.writeUInt32LE(0, 16);
+
+        // Create scene buffer and overall buffer
+        var scene = new Buffer(sceneString);
+        var glb = Buffer.concat([header, scene, body], glbLength);
+
+        if (callback) {
+            process.nextTick(function () {
+                callback(header, scene, body);
```
JoshuaStorm: I am unsure how I can get this same functionality out of a Promise.

lib/gltfPipeline.js
lasalvavida: If you are making a new promise, your function should at least take a `resolve` parameter, which you will call when the function completes. If an exception can be thrown, you should take a `reject` parameter as well and `reject(err)`. That being said, you shouldn't need to write a new promise very often. For example, in this function instead of making a new promise, you should just return `writeBinaryGltf` or `writeGltf` since they are already promises.

lib/getBinaryGltf.js
lasalvavida: You get rid of the callback and wrap up the parameters. This whole statement can be replaced by:

```javascript
resolve({
    header : header,
    scene : scene,
    body : body
});

Comment (lasalvavida) [inline lib/getBinaryGltf.js]: Actually, are the `header` `scene` and `body` parameters necessary? We may be able to take them out entirely.

Comment (lilleyse) [inline lib/getBinaryGltf.js]: As far as I can tell they are used just for testing purposes.

File: specs/lib/writeBuffersSpec.js (hunk 1 above)
Comment (JoshuaStorm) [inline specs/lib/writeBuffersSpec.js]: @lasalvavida 
I appreciate the quick responses to these questions. That makes a lot of sense but I seemed to have just been gleaning over this fact while reading the documentation.

File: lib/getBinaryGltf.js (hunk 3 above)
Comment (lasalvavida) [inline lib/getBinaryGltf.js]: My suggestion would be to have getBinaryGltf return the `header` `scene` and `body`. Make a `getBinaryGltfSpec` and migrate tests that need those parameters there. Then have `writeBinaryGltf` just concat them together and write the output, removing the callback.

File: specs/lib/gltfPipelineSpec.js [hunk 4]
```diff
@@ -50,92 +51,111 @@ describe('gltfPipeline', function() {
...
+            .then(function(gltf) {
+                return processJSONToDisk(gltf, outputGltfPath, processOptions);
+            })
+            .then(function() {
                 expect(path.normalize(spy.calls.first().args[0])).toEqual(path.normalize(outputGltfPath));
                 done();
+            })
+            .catch(function(err) {
+                done();
+                throw err;
```
JoshuaStorm: The three above tests are failing

The spy is never being triggered and I am not exactly sure why. I do a similar spy setup for `writeGltfSpec` and `writeBinaryGltfSpec` that work fine. 

@lasalvavida or @lilleyse, if you have a minute could you briefly look over these tests and see if there is anything I'm obviously missing

lilleyse: I think it's because you remove the args to the fake callback, so args[0] is undefined.

JoshuaStorm: I am fairly certain `spy.calls.first()` accesses the args passed to the function spied on.

See `writeGltfSpec` for example

JoshuaStorm: Doing a few `console.log` traces, it seems that for some reason the the `processFileToDisk` and `processJSONToDisk` are resolving before the `outputJson`  in `writeGltf`

I'm guessing there is something wrong with my promises

lasalvavida: > I'm guessing there is something wrong with my promises

That is true. Hang tight for a little bit; I'm almost done fixing this

lasalvavida:
@JoshuaStorm This branch should be clean again

lasalvavida:
I'm seeing errors in Travis, but this is clean locally for me

JoshuaStorm:
Ignore the most recent commit, it was off-base

I'll look more into this

JoshuaStorm:
I have resolved the Travis failure that occurred due to attempting to write to a directory without the directory first being made.

However, this solution is a bit hacky:

I want to use both synchronous and asynchronous `fs-extra` functions. As such, I cannot just `promisifyAll` of `fs-extra` since that results in an error being thrown when a synchronous function is called. 

A simple solution would be just to `require('fs-extra)` and promisify the individual functions I want to use as promises, but this causes an issue where I cannot spy on the individual function (there may be a simple resolution to this that I am unaware of)


For example, if I do
```
var fsExtra = require('fs-extra');
var outputJson = Promise.promisify(fsExtra.outputJson);
```
and attempt to do `spyOn(fsExtra, 'outputJson')`, the spy does not seem to be triggered by calling the promisified `outputJson`


My current solution is a bit clunky, and inefficient: I use both `var fsExtra = require('fs-extra')` and `var fsPromise = Promise.promisifyAll(require('fs-extra'))` so I can use the synchronous and asynchronous functions, while also being able to `spyOn(fsPromise, 'outputJson')`

lasalvavida:
@JoshuaStorm do the following:

```javascript
fsExtra.outputJsonAsync = Promise.promisify(fsExtra.outputJson);
...
spyOn(fsExtra, 'outputJsonAsync')
```

JoshuaStorm:
I knew I had to be forgetting something straight forward like that.
Thanks!

JoshuaStorm:
This should be good for review now!

lib/writeGltf.js
```diff
@@ -19,31 +28,24 @@ function writeGltf(gltf, outputPath, embed, embedImage, createDirectory, callbac
     if (outputExtension !== '.gltf') {
         throw new DeveloperError('Invalid output path extension.');
     }
+
     // Create the output directory if specified
     if (createDirectory) {
         outputPath = path.join(path.dirname(outputPath), 'output', path.basename(outputPath));
+        fsExtra.ensureDirSync(path.dirname(outputPath));
```
lilleyse: You don't need this because `outputJson` will create the dir if it doesn't exist.

lib/writeGltf.js
```diff
@@ -19,31 +28,24 @@ function writeGltf(gltf, outputPath, embed, embedImage, createDirectory, callbac
...
-                throw err;
-            }
-        } else {
+    var writeSources = [
+        writeSource(gltf, basePath, 'buffers', embed, embedImage),
+        writeSource(gltf, basePath, 'images', embed, embedImage),
+        writeSource(gltf, basePath, 'shaders', embed, embedImage)
+    ];
+    
+
```
lilleyse: Remove extra line.

lib/writeSource.js
```diff
@@ -32,30 +32,14 @@ function writeSource(gltf, basePath, name, embed, embedImage, callback) {
...
                 }
-            } else if (callback) {
-                callback();
             }
-        });
-    } else if (callback) {
-        callback();
+        }
     }
+    return Promise.all(promises);
```
lilleyse: This file is so much cleaner now!

specs/lib/cesiumGeometryToGltfPrimitiveSpec.js
```diff
@@ -1,13 +1,16 @@
...
 var GeometryAttribute = Cesium.GeometryAttribute;
 var Geometry = Cesium.Geometry;
 
+var addDefaults = require('../../lib/addDefaults');
+var cesiumGeometryToGltfPrimitive = require('../../lib/cesiumGeometryToGltfPrimitive');
+var packArray = require('../../lib/packArray');
+var readAccessor = require('../../lib/readAccessor');
+var readGltf = require('../../lib/readGltf');
+
+
```
lilleyse: Remove extra line.

specs/lib/getBinaryGltfSpec.js
```diff
@@ -0,0 +1,127 @@
...
+
+var gltfPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest_BinaryInput.gltf';
+var scenePath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest_BinaryCheck.gltf';
+var bufferPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.bin';
+var imagePath = './specs/data/boxTexturedUnoptimized/Cesium_Logo_Flat_Binary.png';
+var fragmentShaderPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest0FS_Binary.glsl';
+var vertexShaderPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest0VS_Binary.glsl';
+
+describe('getBinaryGltf', function() {
+
```
lilleyse: Remove line.

specs/lib/getBinaryGltfSpec.js
```diff
@@ -0,0 +1,127 @@
...
+        vertexShader: vertexShaderPath
+    };
+
+    beforeAll(function (done) {
+        var options = {};
+        readGltf(gltfPath, options)
+            .then(function (gltf) {
+                testData.gltf = gltf;
+
+                fs.readFile(scenePath, function (err, data) {
```
lilleyse: Promisify this.

specs/lib/writeBuffersSpec.js
```diff
@@ -1,13 +1,17 @@
...
 var outputPath = './specs/data/boxTexturedUnoptimized/output.gltf';
-var outputBufferPath = './specs/data/boxTexturedUnoptimized/output/CesiumTexturedBoxTest.bin';
+var outputBufferPath = './specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.bin';
```
lilleyse: Why did this directory change if `createDirectory` is still true?

lib/writeGltf.js
JoshuaStorm: I had to do this due to `writeSource` using `fs.writeFile` (not fsExtras), but a better solution is probably just to use fsExtras in `writeSource`, so I'll go ahead and do that.

lilleyse: Ah ok, that solution seems good.

lasalvavida:
Async tests should use

```javascript
expect(promise, done).toResolve();
```

https://github.com/AnalyticalGraphicsInc/agi-cesium-people/tree/node-coding-guide/NodeCodingGuide#testing-differences-from-cesium

JoshuaStorm:
@lasalvavida Good catch, I forgot about that.

Just to make sure I do these properly: 

This 
```
        writeGltf(gltf, options)
            .then(function() {
                expect(gltf.shaders.CesiumTexturedBoxTest0FS.extras).not.toBeDefined();
                expect(gltf.shaders.CesiumTexturedBoxTest0FS.uri).toEqual('CesiumTexturedBoxTest0FS.glsl');
                return fsReadFile(outputFragmentShaderPath);
            })
            .then(function(outputData) {
                expect(bufferEqual(outputData, fragmentShaderData)).toBe(true);
                done();
            });
```



should become

```
        var promise = writeGltf(gltf, options)
            .then(function() {
                expect(gltf.shaders.CesiumTexturedBoxTest0FS.extras).not.toBeDefined();
                expect(gltf.shaders.CesiumTexturedBoxTest0FS.uri).toEqual('CesiumTexturedBoxTest0FS.glsl');
                return fsReadFile(outputFragmentShaderPath);
            })
            .then(function(outputData) {
                expect(bufferEqual(outputData, fragmentShaderData)).toBe(true);
                done();
            });
        expect(promise, done).toResolve();

```


Right?


Thanks!

lasalvavida:
Make sure you don't call `done` like in your example.

I would also write it a little more compactly, like this:

```javascript
expect(writeGltf(gltf, options)
    .then(function() {
        expect(gltf.shaders.CesiumTexturedBoxTest0FS.extras).not.toBeDefined();
        expect(gltf.shaders.CesiumTexturedBoxTest0FS.uri).toEqual('CesiumTexturedBoxTest0FS.glsl');
        return fsReadFile(outputFragmentShaderPath);
    })
    .then(function(outputData) {
        expect(bufferEqual(outputData, fragmentShaderData)).toBe(true);
    }), done).toResolve();
```

lasalvavida:
I think `bin/gltf-pipeline.js` may need updating since `processFileToDisk` is a promise now.

lasalvavida:
If node already waits for all promises to terminate, it may be as simple as putting a comment that says that so there's no confusion.

lib/readGltf.js
```diff
@@ -1,16 +1,20 @@
 'use strict';
-var path = require('path');
+var Cesium = require('cesium');
 var fs = require('fs');
+var path = require('path');
+var Promise = require('bluebird');
+var readFile = Promise.promisify(fs.readFile);
```
lasalvavida: Make this fsReadFile for consistency with the node guide.

lib/writeGltf.js
```diff
@@ -1,16 +1,25 @@
...
-var Cesium = require('cesium');
+var Promise = require('bluebird');
+
 var defined = Cesium.defined;
 var DeveloperError = Cesium.DeveloperError;
 
+var removePipelineExtras = require('./removePipelineExtras');
+var writeSource = require('./writeSource');
+
+fsExtra.outputJson = Promise.promisify(fsExtra.outputJson);
```
lasalvavida: Make this something like outputJsonAsync or outputJsonPromise; don't overwrite the original function.

specs/lib/bakeAmbientOcclusionSpec.js
```diff
@@ -2,14 +2,15 @@
 var bakeAmbientOcclusion = require('../../lib/bakeAmbientOcclusion');
```
lasalvavida: This should be down with the other local requires

lasalvavida:
Those are my last few comments; this looks awesome @JoshuaStorm!

lasalvavida:
@lilleyse This does change the public API; do we want to merge this now or wait until after the first alpha release?

lilleyse:
> @lilleyse This does change the public API; do we want to merge this now or wait until after the first alpha release?

I was thinking about this too. It's probably best to wait.

I have another question though. Should our public API return promises at all, or should those functions take callbacks? Internally the callback would just wrap around the promises. It seems like callbacks are still the standard for most other node libraries. @mramato do you have thoughts on this?

mramato:
Our public API should absolutely return promises.  There are two ways to go here:

1. Return promises and not provide callbacks at all. (The ideal solution)
2. Support both, which is what I've seen other libraries do.  Basically, any public function that normally returns a promise would take an optional callback function as it's final argument.  Passing in the callback will cause it to act like a traditional callback function and not return a promise.  We could easily write a single `promiseToCallback` utility function that handles this for us and we would have to call it wherever we return a promise from the public API.

While 2 wouldn't be a ton of code, it does add some additional overhead and clutter the API a bit.  We would also have to add an extra test for every public API function to cover it.  My gut says to go with 1 until enough people ask for 2.  It would be a non-breaking change to add later.

@pjcozzi What do you think?

pjcozzi:
> My gut says to go with 1 until enough people ask for 2

Big +1

lilleyse:
Thanks @lasalvavida and @JoshuaStorm

Issue #139, opened 2016-07-18
In WebStorm, if --ao is put last, it is given the value of the working directory


lasalvavida:
How to reproduce:

1. Make a new Node.js run configuration.
2. run `./bin/gltf-pipeline.js ... --ao`

This actually happens if you put any flag without a value last. But it fails on ao because the typeof check in`bin/gltf-pipeline.js` doesn't work so it just gets assigned to be the path.

lasalvavida:
@likangning93 This has been updated to more accurately reflect the issue

likangning93:
I think this is currently true for any yargs arguments. When I tack something else like `-t` or `--fakeArg` onto the invocation it ends up putting the working directory value with those too.

lasalvavida:
You are correct; I did put that in the last line of the issue. However, I don't think it fails on any of them because we just check if they are defined.

I think just changing:
```javascript
if (typeof(aoOptions) === 'boolean') {
    aoOptions = {};
}
```

to

```javascript
if (typeof(aoOptions) === 'boolean' || typeof(aoOptions) === 'string') {
    aoOptions = {};
}
```

should resolve this.

likangning93:
Ah I see. Thanks!

lasalvavida:
Fixed by #139

PR #140, opened 2016-07-18
AO fixes: command line flag fix, removed unnecessary gltf output

likangning93:
fixes https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/139.
Also, I found and removed some random non-spec attributes that were ending up in gltf passing through the AO stage. I think it must have been left over from a copy-paste at some point...


lasalvavida:
Looks good to me, thanks @likangning93 !

Issue #141, opened 2016-07-19
Images are bigger after encoding with Jimp


lasalvavida:
For example, the CesiumMilkTruck.png from the [glTF sample models](https://github.com/KhronosGroup/glTF/tree/master/sampleModels/CesiumMilkTruck) is 902 KB.

If you run it through Jimp and re-encode it, it ends up as 1,441 KB.

This is obviously not ideal, and it would also be useful to be able to optimize texture images for the web in general.

I did take a brief look at [imagemin](https://www.npmjs.com/package/imagemin) this morning, but I was getting a failure that I think is probably due to AGI's firewall. The [imagemin-optipng](https://www.npmjs.com/package/imagemin-optipng) module has a vendor folder on git with native executables in it that is missing when I get it as an npm module.

Imagemin seems to be the most widely used solution for this, but if we can avoid it, I'd really rather not depend on node modules that have native dependencies.

pjcozzi:
Is this because the resolution changed?  Or because the compression is different?

We should address this soon since this is kinda the opposite of an optimization pipeline.

Longer-term, @lasalvavida is right that we'll want to be able to resize images and more aggressively compress them, especially when the source textures are so big that many texels will map to one pixel at the closest zoom level.

lasalvavida:
> Is this because the resolution changed? Or because the compression is different?

The latter. The resolution is the same.

#142 Compresses without resizing and gets CesiumMilkTruck from 902 KB to 265 KB using lossless png compression.

lilleyse:
> We should address this soon since this is kinda the opposite of an optimization pipeline

@pjcozzi by default the pipeline does not re-encode images, so the pipeline is not making things worse. However these stats show there is definitely room for us to optimize.

pjcozzi:
>>We should address this soon since this is kinda the opposite of an optimization pipeline
>@pjcozzi by default the pipeline does not re-encode images, so the pipeline is not making things worse. However these stats show there is definitely room for us to optimize.

So when is this being used?

lilleyse:
> So when is this being used?

Only when the image is marked as changed, which is only for writing ao into textures.

likangning93:
[...]

PR #142, opened 2016-07-19
Images can be minified using '-m'

lasalvavida:
Addresses #141.

As much as I don't like including native dependencies, the results are hard to argue with:

CesiumMilkTruck run through the pipeline goes from 1,366 KB to 517 KB with the `-m` flag.

You may need to disable TLS for packages with external binaries if you're behind a firewall.
`NODE_TLS_REJECT_UNAUTHORIZED=0 npm install imagemin-pngquant`

Probably worth considering.


pjcozzi:
Do not merge until #141 is fully discussed.  There will need to be a very strong argument to include native code.

lasalvavida:
Closing in favor of texture compression

Issue #143, opened 2016-07-20
Some models don't load in Cesium after quantization


lilleyse:
I've been noticing a pattern where glTFs that don't have shaders/technqiues that are converted with quantization enabled do not load in Cesium. There must be some interaction going on between these two steps. @lasalvavida can you take a look into this?

[dragon-no-shaders.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/374015/dragon-no-shaders.gltf.txt)

`test/dragon-no-shaders.gltf -q`

![capture](https://cloud.githubusercontent.com/assets/915398/16991634/3114c766-4e6a-11e6-9ad6-f9219e9cb59e.JPG)

lasalvavida:
These models are not valid glTF. They have a TEXCOORD_0 semantic in the mesh, but it isn't declared as an attribute in the technique.

lasalvavida:
The crash happens because quantization tries to decode TEXCOORD, but can't map it to a variable in the shader since it isn't declared in the technique.

lilleyse:
It's true that the models aren't valid initially, but after the `modelMaterialCommon` step they should have techniques, but maybe those too are wrong.

lilleyse:
Another issue I'm seeing is with quantization and ao together:

[Shadow_Tester_Point.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/374139/Shadow_Tester_Point.gltf.txt)

![capture2](https://cloud.githubusercontent.com/assets/915398/16993607/1607b62e-4e72-11e6-8bd5-cb4419b3c5ea.JPG)

lasalvavida:
The _OCCLUSION semantic being prefixed by an underscore is the issue here. When the shader gets modified by quantization, there is a double underscore which causes the shader to fail compilation

lasalvavida:
Closed by #144 and #145.

PR #144, opened 2016-07-20
Remove unused primitive attributes

lasalvavida:
Resolves part of #143. @lilleyse Dragon no shaders should convert to valid glTF after this change.


lilleyse:
Nice, thanks!

PR #146, opened 2016-07-20
Fix write source

lilleyse:
This is fixed in https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/135, but since we plan on releasing before then this change should be merged in.

When `-s` is supplied the `writeFile` fails when the output directory doesn't exist. I changed it to `outputFile`.


lasalvavida:
Looks good to me, thanks @lilleyse!

PR #147, opened 2016-07-20
Change ceil back to round

lasalvavida:
@lilleyse, we changed this before to make sure that rounding for integers went up for models with JOINT semantics. That is no longer an issue.

Currently if you run 2CylinderEngine through the pipeline with quantization, it tries to write 65536 to a uint16 buffer and crashes. This is because of a very minor floating point error that is solved by rounding instead.


mramato:
Can this be easily unit tested?  Since we are publishing libraries now and this is going on the website, it would be nice to add tests for problems as we find them.

lasalvavida:
Added a test. 

In the future, we will want some sort of automated testing system that converts sample models to catch stuff like this.

lilleyse:
Looks good.

PR #148, opened 2016-07-20
Only compress accessors that use JOINT semantic

lasalvavida:
ReciprocatingSaw currently fails to render in Cesium after converting.

This is because it has a normal accessor with only integer values, but all unsigned.

`compressIntegerAccessors` puts them into an `UNSIGNED_BYTE` which crashes WebGL for some reason. (Note: if they are in `BYTE`, it is fine).

This change allows `compressIntegerAccessors` to distinguish based on semantics and the main pipeline only runs it on JOINT, which is really what it was for anyway.

### DIFF
--- a/lib/compressIntegerAccessors.js
+++ b/lib/compressIntegerAccessors.js
@@ -62,7 +62,9 @@ function compressIntegerAccessors(gltf, options) {
 }
 
 function compressIntegerAccessor(gltf, primitive, accessorId, matchedSemantic) {
+    /* These callback parameters are unused */
     void(primitive, matchedSemantic);
+
     var accessors = gltf.accessors;
     var accessor = accessors[accessorId];
     var accessorReader = new AccessorReader(gltf, accessor);
--- a/lib/compressIntegerAccessors.js
+++ b/lib/compressIntegerAccessors.js
@@ -1,10 +1,14 @@
 'use strict';
 var Cesium = require('cesium');
+var Promise = require('bluebird');
+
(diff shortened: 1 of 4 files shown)


lib/compressIntegerAccessors.js
```diff
@@ -41,29 +45,40 @@ var unsignedComponentTypes = [
...
+        var semantic = semantics[i];
+        promises.push(getAccessorsForSemantic(gltf, semantic, compressIntegerAccessor));
+    }
+    return Promise.all(promises).then(function() {
+        uninterLeaveAndPackBuffers(gltf);
+    });
+}
+
+function compressIntegerAccessor(gltf, primitive, accessorId, matchedSemantic) {
+    void(primitive, matchedSemantic);
```
lilleyse: What is this line doing?

lasalvavida: It keeps jsHint from failing. I'm explicitly stating that those arguments from the callback aren't used here.

lilleyse: Ah.. ok. We could also set "unused" to false in jshintrc, which I ended up doing in obj2gltf.

lasalvavida: I'm fine with either, do you have a strong preference?

lilleyse: Nope, though if we stay with void maybe a small comment would help.

lasalvavida:
Updated

PR #150, opened 2016-07-21
Review jsonpath for object traversal

lasalvavida:
@pjcozzi, @lilleyse, as discussed offline.

Here's an example with `generateNormals`.

The package can be found on [npm](https://www.npmjs.com/package/jsonpath) under the MIT license.

There's also an [online sandbox](http://jsonpath.com/) for testing queries.


pjcozzi:
Part of #130

lib/generateNormals.js
```diff
@@ -14,10 +17,7 @@ module.exports = generateNormals;
  * @param gltf
  */
 function generateNormals(gltf) {
-    var primitives = getPrimitives(gltf);
-    var primitivesLength = primitives.length;
-    for (var i = 0; i < primitivesLength; i++) {
-        var primitive = primitives[i];
+    jp.apply(gltf, '$.meshes[*].primitives[*]', function (primitive) {
```
pjcozzi: Have you looked at microbenchmarks?  How fast is this compared to doing it by hand and how much GC pressure does it create?  Given how general this library is and how specific our needs are, I suspect we can do something much better ourselves with very little work.  Even if we use this library, as our needs grow, we'll end up writing custom helper functions.

lasalvavida: I did a quick look around and didn't find much for existing performance benchmarks. I'll throw something together later this week. Might even make an interesting blog post.

pjcozzi: What was the verdict here?

pjcozzi:
@lasalvavida I'm going to close this since it will likely be replaced by #191 and custom functions (for fewer dependencies in Cesium) as discussed offline.  Thanks for exploring this.

PR #151, opened 2016-07-29
Combine nodes

lasalvavida:
Adds stage to flatten node hierarchies where possible. Reciprocating saw screenshot attached for comparison.

saw-comparison

@lilleyse Could you look at this when you get a chance?

### DIFF
--- a/lib/combineNodes.js
+++ b/lib/combineNodes.js
@@ -1,0 +1,367 @@
+'use strict';
+var Cesium = require('cesium');
+var jp = require('jsonpath');
+var _ = require('underscore');
+
+var Cartesian3 = Cesium.Cartesian3;
+var Matrix4 = Cesium.Matrix4;
+var defined = Cesium.defined;
+
+var AccessorReader = require('./AccessorReader');
+var NodeHelpers = require('./NodeHelpers');
+var getUniqueId = require('./getUniqueId');
+var readAccessor = require('./readAccessor');
+var removeUnusedNodes = require('./removeUnusedNodes');
+var writeAccessor = require('./writeAccessor');
+
+module.exports = combineNodes;
+
+var rootNodePrefix = 'rootNode';
+var scratchTransform = new Matrix4();
+function combineNodes(gltf) {
+    var nodes = gltf.nodes;
+    var exclude = {};
+    getAnimationTargets(gltf, exclude);
+    getTechniqueTargets(gltf, exclude);
+    jp.apply(gltf, '$.scenes[*]', function(scene) {
+        var sceneNodes = scene.nodes;
+        if (defined(sceneNodes)) {
+            var rootNodeId = getUniqueId(gltf, rootNodePrefix);
+            var rootNode = {
+                children : sceneNodes,
+                meshes : [],
+                matrix : [
+                    1.0, 0.0, 0.0, 0.0,
+                    0.0, 1.0, 0.0, 0.0,
+                    0.0, 0.0, 1.0, 0.0,
+                    0.0, 0.0, 0.0, 1.0
+                ]
+            };
+            nodes[rootNodeId] = rootNode;
+            mergeChildrenIntoNode(gltf, rootNodeId, NodeHelpers.mapMeshesToNodes(gltf), mapMeshesToMeshes(gltf), exclude);
+            removeVestigialNodes(gltf, rootNodeId, exclude);
+            if (rootNode.children.length > 0 && rootNode.meshes.length === 0) {
+                scene.nodes = rootNode.children;
+            } else {
+                scene.nodes = [rootNodeId];
+            }
+        }
+        return scene;
+    });
+
+    removeUnusedNodes(gltf);
+}
+
+function getAnimationTargets(gltf, targets) {
+    jp.apply(gltf, '$.animations[*].channels[*].target.id', function(nodeId) {
+        targets[nodeId] = true;
+        return nodeId;
+    });
+    return targets;
+}
+
+function getTechniqueTargets(gltf, targets) {
+    jp.apply(gltf, '$.techniques[*].parameters[?(@.node)]', function(techniqueParameters) {
+        targets[techniqueParameters.node] = true;
+        return techniqueParameters;
+    });
+    return targets;
+}
+
+function mergeChildrenIntoNode(gltf, nodeId, meshesToNodes, meshesToMeshes, exclude) {
+    var nodes = gltf.nodes;
+    var meshes = gltf.meshes;
+    var node = nodes[nodeId];
+    var children = node.children;
+    var transform = scratchTransform;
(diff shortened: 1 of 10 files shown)


pjcozzi:
@lasalvavida does this replace #51?  Is any of this code based on #51?

Have you tested on a variety of models, including ones where some sub-trees can be combined and some can't?

lasalvavida:
No, none of this is based on #51. I'm pretty sure I've handled all of the edge cases. I used reciprocating saw as an example here because CAD models typically have repeated meshes/overlapping mesh data which are the two cases where primitives can't be transformed and passed up the node tree. 

I have tested on:
Cesium Balloon,
Cesium Milk Truck,
2 Cylinder Engine,
Buggy,
BrainStem,
Reciprocating Saw

package.json
```diff
@@ -36,10 +36,12 @@
     "fs-extra": "0.30.0",
     "image-size": "0.5.0",
     "jimp": "0.2.24",
+    "jsonpath": "^0.2.6",
     "mime": "1.3.4",
     "mkdirp": "0.5.1",
     "object-values": "1.0.0",
     "promise": "7.1.1",
+    "underscore": "^1.8.3",
```
lilleyse: Remove the carets from these.

specs/lib/combinePrimitivesSpec.js
```diff
@@ -120,5 +120,5 @@ describe('combinePrimitives', function() {
                     combinePrimitives(componentTypeError);
                 }).toThrowDeveloperError();
             }), done).toResolve();
-    });
+    });*/
```
lilleyse: What is the plan for these specs?

lasalvavida: My mistake, this shouldn't have been committed. Will revert.

lilleyse:
Testing looks good from my end too.

lasalvavida:
Updated

lib/combineNodes.js
```diff
@@ -0,0 +1,367 @@
+'use strict';
+var Cesium = require('cesium');
+var jp = require('jsonpath');
+var _ = require('underscore');
```
pjcozzi: @mramato is it standard practice to use the name `_` here?

package.json
```diff
@@ -36,10 +36,12 @@
     "fs-extra": "0.30.0",
     "image-size": "0.5.0",
     "jimp": "0.2.24",
+    "jsonpath": "0.2.6",
```
pjcozzi: Update CHANGES.md for this and `underscore`.  I also think we should evaluate this library a bit more as I mentioned in  #150.  It actually could be find if it is slower than custom functions given that the bottlenecks are elsewhere, but I would want to know is it just 2x slower?  50x slower?  How big is it?  etc.

pjcozzi:
@tfili can you please review the test cases in [combineNodesSpec.js](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/151/files#diff-75ba3b52be15c9aec44fa7aecc80d83e) and see if we missed any from your experience with COLLADA?

pjcozzi:
Consider adding one more unit tests that is more like a "system test" that tests several scenarios within one node hierarchy where some sub-trees can be combined and some can't.  Verify that it is as aggressive as possible above and below the roots of the sub-trees that can't be merged into their parent.

Along these lines, can siblings be merged, e.g.,?
```
n0
|
+- n1
|
+- n2
|
+- n3
```
If `n3` can't be merged into `n0`, for example, can `n1` and `n2` be combined (assuming compatibility) so that `n0` only has two children?

lasalvavida:
If `n1` and `n2` can be combined, their meshes will be transformed and passed to `n0`. Then `n1` and `n2` will have no children and no meshes, so they will be removed and `n0` will have one child: `n3`.

lib/combineNodes.js
mramato: Unfortunately, yes.  It's the same idea as using `$` for jquery, which I also hate.  That being said, I would strongly suggest not depending on underscore, there is absolutely no reason that I know of to use it over native JavaScript functions.

lib/combineNodes.js
```diff
@@ -0,0 +1,367 @@
+'use strict';
+var Cesium = require('cesium');
+var jp = require('jsonpath');
```
mramato: variable names should always match module names.

pjcozzi:
Sounds good, please make sure this is covered by tests.

lib/combineNodes.js
pjcozzi: I am fine with removing it; Rob, what do you think?

lasalvavida: It does have a deep comparison `_.equals` which is really what I'm using it for. We have rewritten that function so many times in this library, I figured it might be best to try to consolidate it to a third party utility. 

@mramato I'm fine with using a different node module for deep equals, rolling our own or just keeping underscore, whatever you think is best practice here.

lasalvavida: Also `_.range` is useful for generating indices.

mramato: Underscore is an old, traditionally browser-based library meant to make up for missing features of the browser API and incompatibility from platform to platform.  For Node.js, there is only a single platform and 99% of functionality is either directly supported by the native API or is otherwise only a couple of lines of code.  Here's an article that tries to lay out some examples:  https://www.reindex.io/blog/you-might-not-need-underscore/ That article does use some ES2015 examples, which technically we can't use yet, but it's really just shorthand for stuff that can be done in ES5 already.  We should never use a library if there is an equivalent native API already available, it just makes the code slower and less uniform.

As for equals, I don't see it actually being called anywhere in this PR, but if we really need a deep equality function, something like https://www.npmjs.com/package/deep-equal is probably a better choice.  Also, if we are only using deep-equal in tests, then there's already jasimine.toEqual and assert.equals

Normally, we avoid generic deep comparisons in our code and prefer implementing custom equals functions directly (such as `Matrix4.equals` for example) because they are way faster.  If we need to compare completely unknown blobs, then of course we may need to have generic comparisons, but since gltf is a known schema, I'm not sure of why that would be the case.

lasalvavida: I'm using `_.isEqual` and `_.find` to search for duplicate primitives in an array. That was something else -> optimizing array searching.

I'll take a look around though and see if there's some better choices for that as well.

mramato: Other than deep-equals, I'm not sure why we would need any other libraries.  .find is just a for loop with a break statement.  It provides no benefit.  We already have binarySearch in Cesium if your arrays are sorted.  If they aren't sorted, then there is no "optimized" search method; it's always a linear search.

lasalvavida: Yup, you're right. I thought maybe they were doing a randomized search, which is more optimal than linear for non-sorted data, but it is just linear.

PR #152, opened 2016-08-05
Remove duplicate materials, techniques, programs and shaders

lasalvavida:
From the roadmap #1.

@lilleyse, could you look at this when you get a chance?


lib/mergeDuplicateMaterials.js
```diff
@@ -0,0 +1,55 @@
...
+    var uniqueMaterials = [];
+    for (var materialId in materials) {
+        if (materials.hasOwnProperty(materialId)) {
+            var material = materials[materialId];
+            var uniqueMaterialsLength = uniqueMaterials.length;
+            var unique = true;
+            for (var i = 0; i < uniqueMaterialsLength; i++) {
+                var uniqueMaterialId = uniqueMaterials[i];
+                var uniqueMaterial = materials[uniqueMaterialId];
+                if (deepEqual(material, uniqueMaterial)) {
```
lilleyse: If two materials are the same but have different names, should they also get merged?

lib/mergeDuplicateMaterials.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+var Cesium = require('cesium');
+var deepEqual = require('deep-equal');
+
+var defined = Cesium.defined;
+
+var removeUnusedMaterials = require('./removeUnusedMaterials');
```
lilleyse: Based on the node guide, the include order should be like:

```javascript
var restifyErrors = require('restify-errors');
var aFunction = require('./lib/aFunction');
var bFunction = require('./lib/bFunction');

var Ellipsoid = Cesium.Ellipsoid;
var ResourceNotFoundError = restifyErrors.ResourceNotFoundError;
```

lilleyse:
That's all my comments. Looks good @lasalvavida.

lib/mergeDuplicateMaterials.js
pjcozzi: Yes, since optimizations do not always guarantee preserving names, e.g., like when combing nodes.

pjcozzi:
Are we sure this needs to be as verbosely coded as it is?  For example, see how [removeUnused](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/lib/removeUnused.js) is implemented.

lib/gltfPipeline.js
```diff
@@ -67,6 +71,10 @@ function processJSONWithExtras(gltfWithExtras, options) {
     var stats = new OptimizationStatistics();
     addDefaults(gltfWithExtras, options);
     removeUnused(gltfWithExtras, stats);
+    mergeDuplicateShaders(gltfWithExtras);
```
pjcozzi: _As of now_, is there any reason to also run these stages at the end of the pipeline, e.g., because other stages blindly add new shaders, programs, etc.?

lib/mergeDuplicateMaterials.js
```diff
@@ -0,0 +1,55 @@
...
+                    break;
+                }
+            }
+            if (unique) {
+                uniqueMaterials.push(materialId);
+            }
+        }
+    }
+    remapMaterials(gltf, materialIdMapping);
+    removeUnusedMaterials(gltf);
```
pjcozzi: Although it is not the best name, perhaps rename `mergeDuplicateMaterials` to `removeUnusedAndDuplicateMaterials`, then make `removeUnusedMaterials` private and do not call it from the main pipeline so it is not executed twice.  It kinda helps with performance, but the main reason is to keep the pipeline conceptually simple.

Same comment for all stages in this PR.

specs/lib/mergeDuplicateMaterialsSpec.js
```diff
@@ -0,0 +1,68 @@
...
+                        {
+                            material : 'materialTwo'
+                        },
+                        {
+                            material : 'materialThree'
+                        }
+                    ]
+                }
+            },
+            materials : {
```
pjcozzi: Throughout, also add a material/etc that is not used and verify it is removed.

lib/gltfPipeline.js
lasalvavida: I do not think so.

lib/mergeDuplicateMaterials.js
lasalvavida: Yup, good catch.

specs/lib/mergeDuplicateMaterialsSpec.js
lasalvavida: I'm not sure that's necessary. This stage is designed to merge duplicates, we have other stages for `removeUnusedXXX` that we already have tests for.

pjcozzi: Yes, but see https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/152/files/1d35880f23bafa7c6e685e592077c11b5c41ca50#r73891779.

We would still keep the unit tests for removing; this would be more of an integration test since removing is still part of the function's contract.

lasalvavida:
Updated. This is now merged with the new API and includes doc and is scoped more in line with the way `RemoveUnusedProperties` is done. `name` and `extras` are ignored on property comparison. I've elected to keep the name schema the same, but to make removing after merging not part of this function's contract. `removeAllUnused` is run after this stage in the main pipeline instead.

lib/MergeDuplicateProperties.js
```diff
@@ -0,0 +1,383 @@
...
+function MergeDuplicateProperties() {}
+
+/**
+ * Merges all duplicate elements in the glTF asset in top-down order so merged references resolve down the hierarchy.
+ *
+ * @param {Object} gltf A javascript object containing a glTF asset.
+ * @returns {Object} The glTF asset with merged duplicate elements.
+ */
+MergeDuplicateProperties.mergeAllDuplicates = function(gltf) {
+    MergeDuplicateProperties.mergeDuplicateAccessors(gltf);
```
pjcozzi: The names are a bit redundant (as I imagine they are for removing unused).  It's a bit awkward because the class is a verb, not a noun.  Maybe:
```
MergeDuplicateProperties.mergeDuplicateAccessors -> MergeDuplicateProperties.mergeAccessors
MergeDuplicateProperties.mergeDuplicateAccessors -> DuplicateEliminator.mergeAccessors
MergeDuplicateProperties.mergeDuplicateAccessors -> DuplicateOptimizer.mergeAccessors
```
Consider a similar rename for removing unused; breaking the API is no problem at this point.

lasalvavida:
Updated. Unless anybody has anything else, this should be ready to merge

lasalvavida:
@lilleyse could you look at this when you get a chance?

lilleyse:
Looks good to me. Are there any glTF models that I can test with?

lasalvavida:
You can test with the VirtualCity glTF sample model. For vertex shaders, VC2VS is a duplicate of VC0VC, VC5VS is a duplicate of VC3VS, and VC6VS is a duplicate of VC0VS. For materials, Vmtl06 is a duplicate of Vmtl105, and Vmtl08 is a duplicate of Vmtl03.

lilleyse:
Looks good!

PR #155, opened 2016-08-10
Dont optimize without indices

### LINKED ISSUE #154: gltfPrimitiveToCesiumGeometry assumes that indices is defined
Running a gltf model without indices through the pipeline crashes currently at the `cacheOptimization` stage.

lasalvavida:
Fixes #154, @lilleyse, could you look at this when you get a chance?

### DIFF
--- a/specs/lib/cacheOptimizationSpec.js
+++ b/specs/lib/cacheOptimizationSpec.js
@@ -75,8 +75,6 @@ describe('cacheOptimization', function() {
                 }
             }
         };
-        expect(function() {
-            cacheOptimization(gltf);
-        }).not.toThrowRuntimeError();
+        cacheOptimization(gltf);
     });
 });
--- a/specs/lib/cacheOptimizationSpec.js
+++ b/specs/lib/cacheOptimizationSpec.js
@@ -1,4 +1,7 @@
 'use strict';
+var Cesium = require('cesium');
+var GeometryPipeline = Cesium.GeometryPipeline;
+
 var addDefaults = require('../../lib/addDefaults');
 var cacheOptimization = require('../../lib/cacheOptimization');
 var readAccessor = require('../../lib/readAccessor');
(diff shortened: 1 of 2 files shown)


specs/lib/cacheOptimizationSpec.js
```diff
@@ -59,4 +59,24 @@ describe('cacheOptimization', function() {
...
+                                NORMAL : 'notImportantEither'
+                            }
+                        }
+                    ]
+                }
+            }
+        };
+        expect(function() {
+            cacheOptimization(gltf);
+        }).not.toThrowRuntimeError();
```
lasalvavida: I just checked and this doesn't fail without the changes, so you're right, we need a better test here.

lasalvavida:
Updated

specs/lib/cacheOptimizationSpec.js
```diff
@@ -59,4 +59,22 @@ describe('cacheOptimization', function() {
...
+                            attributes : {
+                                POSITION : 'notImportant',
+                                NORMAL : 'notImportantEither'
+                            }
+                        }
+                    ]
+                }
+            }
+        };
+        cacheOptimization(gltf);
```
lilleyse: You should have at least on expect in here. Maybe just create a spy for `GeometryPipeline.reorderForPostVertexCache` and check that it's not called.

lilleyse: Oh well I guess that would pass for both...

lasalvavida: No, that should work. gltfPrimitiveToCesiumGeometry would work too.

lasalvavida:
Updated

lilleyse:
Sorry just had a last minute thought, can you update CHANGES.md with the change? We should start getting in the habit of this now.

lasalvavida:
Yup

Issue #153, opened 2016-08-08
Exposing helpers


likangning93:
As discussed offline, a lot of potentially useful gltf-pipeline helpers aren't being exposed right now. We can keep a running list here. Some ideas for exposing them:
- ~~export a `utils.js` file to gltf-pipeline that exposes useful utils. this to avoid clutter in `index.js`~~
- ~~expose requested utilities as "_private"~~
- ~~just expose everything?~~ Expose a selection

Wishlist:
* [ ] readAccessor
* [ ] getUniqueId
* [ ] packArray
* [ ] byteLengthForComponentType

pjcozzi:
It is perfectly fine to treat the gltf-pipeline public API the same way we treat the Cesium API meaning that these functions could be exposed as individual functions, e.g., like [Cesium.clone](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/clone.js); there's no need to group them into one util class.

lasalvavida:
I will try to get to this today; it's been on my to-do list for a while.

lasalvavida:
Fixed in #157, closing this issue.

Issue #108, opened 2016-06-24
Combine Primitives needs to be fixed


lasalvavida:
This is a parent issue for #85 and #103.

Currently `combinePrimitives` produces unpredictable results and shouldn't be in master as per #107.

The TODO from #107 should be removed once this is rectified.

lasalvavida:
Closed by #162.

PR #157, opened 2016-08-10
Export public API via index.js

lasalvavida:
~~All lib/**/*.js functions are exposed via index.js. Also adds a gulp task for regenerating this file.~~

A collection of functions are exposed as the public API. They are all documented.


lasalvavida:
Resolves #153

index.js
```diff
@@ -1,6 +1,73 @@
 module.exports = {
-    gltfPipeline : require('./lib/gltfPipeline'),
-    readGltf : require('./lib/readGltf'),
-    writeBinaryGltf : require('./lib/writeBinaryGltf'),
-    writeGltf : require('./lib/writeGltf')
-};
+    AccessorReader : require('./lib/AccessorReader.js'),
```
pjcozzi: Are you sure you want to expose all of this?  That is _a lot_ of API surface area.  We learned a very hard lesson in Cesium to only expose what is needed to keep the maintain cost way down.

Please only expose what is currently needed and make sure the public API has full reference doc that can be generated like we do for Cesium.  Don't worry about the generated doc's styling or hosting right now.

lasalvavida:
I've gone through and trimmed down what we are putting into the public API. Everything exposed now has documentation. It probably isn't perfect, but it is a good starting point.

Things that are exposed: "stages", and helper functions that are fairly glTF specific i.e. `byteLengthForComponentType`, but not `triangleAxisAlignedBoundingBoxOverlap` since it is more geometry focused.

lasalvavida:
I also moved the addLightDefaults function to modelMaterialsCommon to resolve #158. There was a TODO for this.

lasalvavida:
@likangning93 I have exposed everything listed in #153, let me know if there's anything else you want exposed that I missed.

lasalvavida:
Also making a note here that combinePrimitives is not currently exposed. It should be exposed once #108 is resolved.

index.js
```diff
@@ -1,6 +1,59 @@
...
+    removeUnusedSkins : require('./lib/removeUnusedSkins.js'),
+    removeUnusedTechniques : require('./lib/removeUnusedTechniques.js'),
+    removeUnusedTextures : require('./lib/removeUnusedTextures.js'),
+    removeUnusedVertices : require('./lib/removeUnusedVertices.js'),
+    uninterleaveAndPackBuffers : require('./lib/uninterleaveAndPackBuffers.js'),
+    writeAccessor : require('./lib/writeAccessor.js'),
+    writeBinaryGltf : require('./lib/writeBinaryGltf.js'),
+    writeBufferComponentType : require('./lib/writeBufferComponentType.js'),
+    writeGltf : require('./lib/writeGltf.js')
+};
```
lilleyse: The main problem with exposing a lot of these functions is that they may rely on previous steps, like `addPipelineExtras`. Either we need to handle that in the functions themselves, or expect users to call things like `addPipelineExtras`, `loadGltfUris`, etc on their own.

lasalvavida: Would it be enough to just put the dependencies in the doc until we work out custom pipelines?

lilleyse: Yeah I think that's ok for now.

lilleyse:
I just made minor edits. Thanks for adding all this documentation and fixing a bunch of problems you noticed.

I organized some of the requires to follow the node guide, where Cesium functions are set after all the requires:

```
var restifyErrors = require('restify-errors');
var aFunction = require('./lib/aFunction');
var bFunction = require('./lib/bFunction');

var Ellipsoid = Cesium.Ellipsoid;
var ResourceNotFoundError = restifyErrors.ResourceNotFoundError;
```

lib/gltfPipeline.js
```diff
@@ -52,6 +53,13 @@ function writeSources(gltf) {
         });
 }
 
+/**
+ * Add pipeline extras and load uris, then process the gltf.
+ *
+ * @param {Object} gltf A javascript object holding a glTF hierarchy.
+ * @param {Object} options Options to apply to stages during optimization.
```
lilleyse: In some areas like this the options aren't documented.

lasalvavida: We pass options to stages and they should be documented for all of those. I think it would be a bit verbose to document all of the different stage options here.

lilleyse: Well it may be redundant in some cases. Maybe just make a note to refer to the `processJSONWithExtras` doc.

lilleyse: Ah you beat me to it.

lasalvavida:
@lilleyse updated

lilleyse:
The new changes look good. @pjcozzi feel free to do a final check.

index.js
```diff
@@ -1,6 +1,59 @@
...
+    convertDagToTree : require('./lib/convertDagToTree.js'),
+    encodeImages : require('./lib/encodeImages.js'),
+    findAccessorMinMax : require('./lib/findAccessorMinMax.js'),
+    generateNormals : require('./lib/generateNormals.js'),
+    getAccessorByteStride : require('./lib/getAccessorByteStride.js'),
+    getBinaryGltf : require('./lib/getBinaryGltf.js'),
+    getUniqueId : require('./lib/getUniqueId.js'),
+    gltfPipeline : require('./lib/gltfPipeline.js'),
+    loadGltfUris : require('./lib/loadGltfUris.js'),
+    mergeBuffers : require('./lib/mergeBuffers.js'),
```
lilleyse: I just noticed that there isn't doc for this one.

index.js
```diff
@@ -1,6 +1,59 @@
...
+    NodeHelpers : require('./lib/NodeHelpers.js'),
+    numberOfComponentsForType : require('./lib/numberOfComponentsForType.js'),
+    octEncodeNormals : require('./lib/octEncodeNormals.js'),
+    packArray : require('./lib/packArray.js'),
+    parseBinaryGltf : require('./lib/parseBinaryGltf.js'),
+    quantizeAttributes : require('./lib/quantizeAttributes.js'),
+    readAccessor : require('./lib/readAccessor.js'),
+    readBufferComponentType : require('./lib/readBufferComponentType.js'),
+    readGltf : require('./lib/readGltf.js'),
+    removeDuplicatePrimitives : require('./lib/removeDuplicatePrimitives.js'),
```
lilleyse: And doc for this.

index.js
lilleyse: Never mind...

index.js
lilleyse: Never mind...

pjcozzi:
Add instructions for building doc to the README.md.

lasalvavida:
As discussed with @lilleyse, I have moved some items out of the global scope. This included some other clean up that was convenient to do during this process. See CHANGES.md for details.

I've pruned a few more things out of the public API. This should be ready for another look.

CHANGES.md
```diff
@@ -4,6 +4,12 @@ Change Log
 ### Next Release
 
 * cacheOptimization no longer crashes on primitives without indices. [#154](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/154)
+* Public API is exposed via index.js [#153](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/153)
+  * Documentation has been added for all exposed functions.
+  * OptimizationStats is removed from removeUnusedStages.
+  * gltfPipeline.js is now named Pipeline.js.
+  * bakeAmbientOcclusion.js is now named AmbientOcclusion.js, bakeAmbientOcclusion is accessible via this object.
+  * All removeUnusedElement type stages have been consolidated to RemoveUnusedElements to clean up the global scope.
```
pjcozzi: In the context of glTF, what is an "element?"  The spec does not define this.  Why not just "RemoveUnusedProperties?"

CHANGES.md
```diff
@@ -4,6 +4,12 @@ Change Log
 ### Next Release
 
 * cacheOptimization no longer crashes on primitives without indices. [#154](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/154)
+* Public API is exposed via index.js [#153](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/153)
+  * Documentation has been added for all exposed functions.
+  * OptimizationStats is removed from removeUnusedStages.
```
pjcozzi: Throughout CHANGES.md use, for example, `OptimizationStats` not just OptimizationStats for classes/functions/etc.

pjcozzi:
Throughout the doc, please change "gltf" to "glTF", e.g.,

> "Process a gltf on disk into memory..."

to

> "Process a glTF on disk into memory..."

Also, in this context, it should really be "glTF asset" or "glTF model", whichever the glTF spec uses.

lib/NodeHelpers.js
```diff
@@ -131,17 +160,24 @@ function getLocalMatrix4(node, result) {
...
+/**
+ * Perform an operation on each primitive in a scene.
+ * primitiveFunction has the signature: function(primitive, meshPrimitiveId, parameters, node).
+ *
+ * @param {Object} gltf A javascript object holding a glTF hierarchy.
+ * @param {Object} scene The glTF scene to traverse.
+ * @param {Function} primitiveFunction The function to apply to each primitive.
+ * @param {Object} parameters Parameters to pass to primitiveFunction.
+ */
+NodeHelpers.forEachPrimitiveInScene = function(gltf, scene, primitiveFunction, parameters) {
```
pjcozzi: This is unrelated to this class and could be its own file.

Also, I am not confident that this class in general is the right design compared to separate functions.  Combing loosely related functions into a "helper" class is a typical code "smell."  We can leave it for now (would love to see a name without "helper" in it though), but I expect we will change this as we introduce whatever the jQuery for glTF ends up being.

lib/cacheOptimization.js
```diff
@@ -62,6 +63,18 @@ function getIndependentPrimitive(dictionary, indicesId, attributeAccessors) {
...
+ *
+ * @param {Object} gltf A javascript object holding a glTF hierarchy.
+ * @param {Number} [cacheSize=24] The cacheSize to use for Cesium.GeometryPipeline.reorderForPostVertexCache.
+ * @returns {Object} gltf with cache-optimized primitives.
+ *
+ * gltf must be initialized for the pipeline.
+ * @see addPipelineExtras
+ * @see loadGltfUris
+ */
 function cacheOptimization(gltf, cacheSize) {
```
pjcozzi: Rename this to a verb and get more precise doc from Cesium.

lib/combineMeshes.js
```diff
@@ -1,10 +1,21 @@
...
+
 var defined = Cesium.defined;
+
 module.exports = combineMeshes;
 
-//Combines meshes in a node. 
-//Should be run before the combinePrimitives stage to merge all primitives for a node into one mesh.
+/**
+ * Combines meshes in a node.
+ * Should be run before the combinePrimitives stage to merge all primitives for a node into one mesh.
```
pjcozzi: Why not `{@link combinePrimitives}`?

Same comment throughout when it applies.

lib/findAccessorMinMax.js
```diff
@@ -9,6 +10,17 @@ var readBufferComponentType = require('./readBufferComponentType');
...
+ *
+ * @param {Object} gltf A javascript object holding a glTF hierarchy.
+ * @param {Object} accessor The accessor object from gltf to read.
+ * @returns {{min: Array, max: Array}} min holding the array of minimum values and max holding the array of maximum values.
+ *
+ * gltf must be initialized for the pipeline.
+ * @see addPipelineExtras
+ * @see loadGltfUris
+ */
 function findAccessorMinMax(gltf, accessor) {
```
pjcozzi: Generally, we do not use abbreviations in public APIs, but this is what glTF calls them so it is OK here.

pjcozzi:
Throughout, the doc needs a lot of work.  I am fine with merging this to get the mechanics and then writing more solid doc for a code sprint or something like that, see #163.

@lasalvavida before we merge this, can you please take a pass through our [Documentation Guide](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Documentation/Contributors/DocumentationGuide/README.md) to see if there are other things we should get into this PR?

lib/writeBufferComponentType.js
```diff
@@ -4,6 +4,14 @@ var WebGLConstants = Cesium.WebGLConstants;
...
 
+/**
+ * Writes a value to a buffer as a particular component type.
+ *
+ * @param {Buffer} buffer The buffer to be written.
+ * @param {Number} componentType WebGLConstants value for component type.
+ * @param {Number} value The value to write.
+ * @param {Number} byteOffset The offset into the buffer to be written.
+ */
 function writeBufferComponentType(buffer, componentType, value, byteOffset) {
```
pjcozzi: Why does this function need `Component` in it?  Why not just `writeBufferComponent`?  This writes one component to a buffer, right?  The "type" doesn't need to be explicit in the name.

lib/writeBinaryGltf.js
```diff
@@ -14,6 +14,21 @@ fsExtra.outputFileAsync = Promise.promisify(fsExtra.outputFile);
 
 module.exports = writeBinaryGltf;
 
+/**
+ * Writes gltf to file as binary glTF.
```
pjcozzi: > Writes gltf to **a** file as binary glTF.

lib/writeAccessor.js
```diff
@@ -4,6 +4,17 @@ var numberOfComponentsForType = require('./numberOfComponentsForType');
...
+ *
+ * @param {Object} gltf A javascript object holding a glTF hierarchy.
+ * @param {Object} accessor glTF accessor where data should be written.
+ * @param {Array} dataArray Data to write into the accessor.
+ *
+ * gltf must be initialized for the pipeline.
+ * @see addPipelineExtras
+ * @see loadGltfUris
+ */
 function writeAccessor(gltf, accessor, dataArray) {
```
pjcozzi: `dataArray` is kind of awkward name.  Generally, we use `values` in Cesium.

lib/writeBufferComponentType.js
pjcozzi: Same comment for `readBufferComponentType`.

pjcozzi:
@lilleyse can you please scrub this public API to make sure the naming and general patterns follow Cesium's best practices?  I could a few things with a quick look, but I suspect there are more from old code...not necessarily changes in this PR...but I want to get the initial public API as close to right as possible even if it won't be widely used for a while.

pjcozzi:
@lasalvavida this is a huge step in the right direction, thanks for starting to kick this into shape!

lasalvavida:
Updated, @lilleyse, this is ready for a look.

lib/NodeHelpers.js
lasalvavida: I'm going to leave it for now. It isn't exposed as part of the public API so we can feel free to replace or change it at any time without breaking anything.

pjcozzi: Let's mark all classes that are not exposed in the public API as `@private` like we do in Cesium; otherwise, it is not obvious unless you look in index.js.

lasalvavida: Sorry, I thought I already did that here. Will fix.

lasalvavida:
Updated

lilleyse:
I did some general cleanup which someone else should look over before I merge. The recent changes to index.js seem good to me.

index.js
```diff
@@ -28,16 +26,18 @@ module.exports = {
     optimizeCache : require('./lib/optimizeCache'),
     packArray : require('./lib/packArray'),
     parseBinaryGltf : require('./lib/parseBinaryGltf'),
+    Pipeline : require('./lib/Pipeline'),
```
lasalvavida: Is this correct? Shouldn't the capitalization put this first?

lib/AccessorReader.js
```diff
@@ -5,15 +5,16 @@ var defaultValue = Cesium.defaultValue;
 var byteLengthForComponentType = require('./byteLengthForComponentType');
 var getAccessorByteStride = require('./getAccessorByteStride');
 var numberOfComponentsForType = require('./numberOfComponentsForType');
-var readBufferComponentType = require('./readBufferComponent');
-var writeBufferComponentType = require('./writeBufferComponent');
+var readBufferComponent = require('./readBufferComponent');
```
lasalvavida: Good catch, I used the JetBrains refactor tool and it must have missed those.

index.js
lilleyse: Oh, I didn't know that was a common thing to do. If so I'll correct it.

lasalvavida: @mramato, I can't find a point of reference for this. Which way should this go?

index.js
```diff
@@ -1,6 +1,43 @@
 module.exports = {
-    gltfPipeline : require('./lib/gltfPipeline'),
+    AccessorReader : require('./lib/AccessorReader'),
```
mramato: In there a reason we just don't just `module.exports = require('./lib');` here instead of requiring every object individually?

lilleyse: There are some files we don't export.

index.js
mramato: It doesn't really matter in this case.  Since this won't be auto-sorted, whatever you guys want to do is fine.

index.js
mramato: I think manually maintaining this list could be error prone and easy to forget about.  Some other options would be to have a `private` folder at the same level of `lib` to make the decision more conscious during development; or you could just export everything and mark private objects `@private`.

That being said, I'm okay if you go the manual route until it becomes a pain.

lib/AmbientOcclusion.js
```diff
@@ -99,21 +99,25 @@ function bakeAmbientOcclusion(aoOptions) {
...
             scene : scene,
             aoBufferByPrimitive : raytracerScene.aoBufferByPrimitive,
             shaderMode : options.shaderMode,
             ambientShadowContribution : options.ambientShadowContribution
         });
     }
-}
+};
+
+AmbientOcclusion.generateOptions = generateOptions;
```
mramato: Are these functions used outside of testing?  If not, then get rid of the `AmbientOcclusion` object and just store these directly on `bakeAmbientOcclusion` (and make `bakeAmbientOcclusion` the only thing you export).

lasalvavida:
Updated. `AmbientOcclusion` is `bakeAmbientOcclusion`, but the stage function is directly exported with the test functions exposed as private members.

lasalvavida:
@lilleyse Is there anything else we need to do here or is this good to merge?

lilleyse:
I think this is good to go.

Once we publish to npm we will have to remember to update several projects because of the `gltfPipeline.js` -> `Pipeline.js` name change.

Issue #160, opened 2016-08-11
Remove underscore usage from combineNodes


lasalvavida:
Also remove it as a dependency in package.json.

See the discussion at https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/151#discussion_r73058316 for details.

In short, we've decided not to use underscore in this project.

lasalvavida:
Closed by #162.

Issue #103, opened 2016-06-21
Buggy model gets mangled by pipeline


JoshuaStorm:
While testing my own branch, I have found a model that is mangled in my branch as well as master:

Quickly commenting out each step and seeing the results seems to show that the model is getting mangled by the `combinePrimitives` step.

[buggy.gltf.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/326569/buggy.gltf.zip)

![goodbuggy](https://cloud.githubusercontent.com/assets/12414183/16245842/b88b5530-37d0-11e6-8b01-08f364b12ed5.jpg)
![mangledbuggy](https://cloud.githubusercontent.com/assets/12414183/16245852/bcf89132-37d0-11e6-9b8c-c3e990e5cded.jpg)

lasalvavida:
Closed by #162.

PR #162, opened 2016-08-12
Combine primitives

lasalvavida:
This is worth an initial look and some testing while I add tests for everything.

This pull request contains a few things:
1. Re-adds combine primitives, closing out #108 and by proxy #85 and #103.
2. Splits out some primitive helper functions from combineNodes into PrimitiveHelpers.
3. Removes underscore usage #160.
3. mergeBuffers will no longer fail if the new buffer name is the same as an already existing one.
4. Fixes an error in `removeDuplicatePrimitives.removePrimitivesFromMesh`
5. Adds a second pass of light optimization

Some preliminary testing with the buggy model:

|  | Original | Master | Master w/ combine primitives |
| --- | --- | --- | --- |
| File size | 10,414 KB | 6,091 KB | 6,046 KB |
| Nodes | 208 | 179 | 179 |
| Meshes | 109 | 163 | 68 |
| Draw Calls | 236 | 273 | 236 |


lib/PrimitiveHelpers.js
```diff
@@ -0,0 +1,212 @@
...
+var Cesium = require('cesium');
+var deepEqual = require('deep-equal');
+
+var Cartesian3 = Cesium.Cartesian3;
+var Matrix4 = Cesium.Matrix4;
+var defined = Cesium.defined;
+
+var AccessorReader = require('./AccessorReader');
+var readAccessor = require('./readAccessor');
+var writeAccessor = require('./writeAccessor');
```
lilleyse: Same note about include order from the other PR.

lib/PrimitiveHelpers.js
```diff
@@ -0,0 +1,212 @@
...
+    getAllPrimitives : getAllPrimitives,
+    getPrimitivesByMaterialMode : getPrimitivesByMaterialMode,
+    markPrimitiveConflicts : markPrimitiveConflicts,
+    primitiveEquals : primitiveEquals,
+    primitivesShareAttributeAccessor : primitivesShareAttributeAccessor,
+    primitivesHaveOverlappingIndexAccessors : primitivesHaveOverlappingIndexAccessors,
+    transformPrimitives : transformPrimitives
+};
+
+
```
lilleyse: Spacing

lib/combinePrimitives.js
```diff
@@ -1,255 +1,310 @@
...
 
-var byteLengthForComponentType = require('./byteLengthForComponentType');
-var getAccessorByteStride = require('./getAccessorByteStride');
+var PrimitiveHelpers = require('./PrimitiveHelpers');
+var findAccessorMinMax = require('./findAccessorMinMax');
+var getUniqueId = require('./getUniqueId');
 var numberOfComponentsForType = require('./numberOfComponentsForType');
-var readBufferComponentType = require('./readBufferComponentType');
-var writeBufferComponentType = require('./writeBufferComponentType');
+var readAccessor = require('./readAccessor');
```
lilleyse: Include order

lib/combinePrimitives.js
```diff
@@ -1,255 +1,310 @@
...
-}
+/**
+ * Combines all of the provided primitives if possible.
+ * If primitives have no shared data with the other primitives, their attributes and indices will be concatenated together and
+ * the indices will be incremented by the attribute offset. If the primitives have the same attribute accessors but different index
+ * accessors, the index accessors will be concatenated. In all other cases, the primitives will be left alone.
+ *
+ * @param gltf
+ * @param primitives
+ * @returns {*}
```
lilleyse: This doc is incomplete.

lib/combinePrimitives.js
```diff
@@ -1,255 +1,310 @@
...
-        "type": referenceAccessor.type,
-        "extras": {
-            "_pipeline": {
-                "deleteExtras": true
+    gltf.buffers[bufferId] = {
+        byteLength : attributeBuffer.length,
+        type : 'arraybuffer',
+        extras : {
+            _pipeline : {
+                source : attributeBuffer
```
lilleyse: I think this should have additional properties like `mergeBuffers`
```javascript
source: source,
extension: ".bin",
deleteExtras: true
```

lib/combinePrimitives.js
```diff
@@ -1,255 +1,310 @@
...
-                writeBufferComponentType(accessorSource, accessor.componentType, value, accessorSourceOffset + k * componentByteLength);
-                min[k] = Math.min(min[k], value);
-                max[k] = Math.max(max[k], value);
+    };
+    gltf.buffers[bufferId] = {
+        byteLength : indicesBuffer.length,
+        type : 'arraybuffer',
+        extras : {
+            _pipeline : {
+                source : indicesBuffer
```
lilleyse: Same here.

lib/gltfPipeline.js
```diff
@@ -79,21 +79,31 @@ function processJSONWithExtras(gltfWithExtras, options) {
...
+    removeDuplicatePrimitives(gltfWithExtras);
+    combinePrimitives(gltfWithExtras);
+    combineMeshes(gltfWithExtras);
+    removeUnused(gltfWithExtras, stats);
     generateNormals(gltfWithExtras);
-    // TODO: Combine primitives can be uncommented and added back into the pipeline once it is fixed, but right now there are too many issues with it to allow in the main pipeline.
-    // combinePrimitives(gltfWithExtras);
-    // Merging duplicate vertices again to prevent repeat data in newly combined primitives
-    // mergeDuplicateVertices(gltfWithExtras);
+    mergeDuplicateVertices(gltfWithExtras);
```
lilleyse: As we talked about offline it may be worth testing if we still need this here, or if the final `mergeDuplicateVertices` is good enough.

lib/readAccessor.js
```diff
@@ -20,9 +20,11 @@ module.exports = readAccessor;
  * @param {Object} gltf
  * @param {Object} accessor
  * @param {Array} results
+ * @param {Boolean} [generateAttributes=false]
```
lilleyse: Should this be `[generateAttributes=true]`?

lib/removeDuplicatePrimitives.js
```diff
@@ -1,12 +1,12 @@
 'use strict';
 var Cesium = require('cesium');
-var _ = require('underscore');
+
+var defined = Cesium.defined;
 
 var NodeHelpers = require('./NodeHelpers');
+var PrimitiveHelpers = require('./PrimitiveHelpers');
 var getUniqueId = require('./getUniqueId');
```
lilleyse: Fix require order here too. (And alphabetical).

package.json
```diff
@@ -33,6 +33,7 @@
     "clone": "1.0.2",
     "data-uri-to-buffer": "0.0.4",
     "datauri": "1.0.4",
+    "deep-equal": "^1.0.1",
```
lilleyse: Remove the ^

lilleyse:
Looks great! I did some simple testing and it seems to work well.

pjcozzi:
@lasalvavida can you verify that it makes sense for the buggy model to have 95 fewer meshes, but only 37 fewer draw calls?

lib/PrimitiveHelpers.js
```diff
@@ -0,0 +1,212 @@
...
+var Cartesian3 = Cesium.Cartesian3;
+var Matrix4 = Cesium.Matrix4;
+var defined = Cesium.defined;
+
+var AccessorReader = require('./AccessorReader');
+var readAccessor = require('./readAccessor');
+var writeAccessor = require('./writeAccessor');
+
+module.exports = {
```
pjcozzi: @lilleyse @lasalvavida can you two please review the design of this class?  I don't think it is needed.  Some functions could be private in the files they are used in; others could be standalone functions.

Let me know if you need help.

lasalvavida:
@pjcozzi Master currently bloats the number of meshes in cases where two meshes that are in node endpoints that can't be combined contain identical primitives. That is what drives up the draw call count. The output with this PR removes that bloating by recombining meshes after primitives get combined.

tl;dr It has the same number of draw calls as the version in the sampleModels, the current output of master is adding more unnecessary draw calls.

pjcozzi:
Gotcha.

lilleyse:
@lasalvavida Is there more to do for this? The latest merge commit has a few issues.

lasalvavida:
There was a bad merge; it still needs tests. I'll add those now.

lasalvavida:
@lilleyse This is updated and ready for a look

lasalvavida:
@lilleyse, I've done some cleanup here to get this in line with master again, and I've pulled the `createAccessor` utility from the hard-normals branch to simplify the code a bit. I've tested a few models, and everything looks good on my end. Let me know if there's anything else that needs to be done here before merging.

lilleyse:
I'll finish reviewing this soon.

Not super relevant to this PR, but does the second `mergeDuplicateVertices` call in `Pipeline.processJSONWithExtras` have much effect anymore? It would be great if we could remove that one if possible to help improve performance.

lasalvavida:
I have removed the second `mergeDuplicateVertices` call. With the light optimization, it is no longer needed; the files produced with and without it are the same size.

pjcozzi:
@lilleyse will you be able to review this soon?  We have a lot of open PRs in this branch and I would like to get them merged.

lilleyse:
Yes I will review this today.

lib/Pipeline.js
```diff
@@ -105,8 +107,17 @@ Pipeline.processJSONWithExtras  = function(gltfWithExtras, options) {
...
-    // Run removeAll stage again after all pipeline stages have been run to remove objects that become unused
+    // Run through light optimization for newly created data
+    MergeDuplicateProperties.mergeAccessors(gltfWithExtras);
+    removeDuplicatePrimitives(gltfWithExtras);
+    combineNodes(gltfWithExtras);
+    combineMeshes(gltfWithExtras);
+    MergeDuplicateProperties.mergeAccessors(gltfWithExtras);
+    removeDuplicatePrimitives(gltfWithExtras);
+    combinePrimitives(gltfWithExtras);
+    combineMeshes(gltfWithExtras);
```
lilleyse: I think we talked offline about this a while ago, but I forget the conversation...

If possible we should try to avoid the second cleanup, and ideally our stages shouldn't be producing unoptimized gltf's. Sorry if I forgot the reason behind this.

lib/PrimitiveHelpers.js
```diff
@@ -0,0 +1,213 @@
...
+        if (indices.indexOf(index) >= 0) {
+            return true;
+        }
+        accessorReader.next();
+    }
+    return false;
+}
+
+function transformPrimitives(gltf, primitives, transform) {
+    var inverseTranspose = new Matrix4();
```
lilleyse: Can't you just do a simple loop over the position and normal arrays, transforming each set of 3 values with the transform and then writing back into the buffer. I don't see the need for involving indices.

lib/PrimitiveHelpers.js
```diff
@@ -0,0 +1,213 @@
...
+            primitives = primitives.concat(mesh.primitives);
+        }
+    }
+    return primitives;
+}
+
+function generateIndices(v, k) {
+    void(v); // v is an unused parameter
+    return k;
+}
```
lilleyse: If still needed, just put this code in `transformPrimitives`.

lib/combineNodes.js
```diff
@@ -265,123 +261,12 @@ function meshHasConflict(gltf, mesh, compareMesh) {
         var primitive = primitives[i];
         for (var j = 0; j < comparePrimitivesLength; j++) {
             var comparePrimitive = comparePrimitives[j];
-            if (primitivesShareAttributeAccessor(primitive, comparePrimitive)) {
-                if (primitivesHaveOverlappingIndexAccessors(gltf, primitive, comparePrimitive)) {
+            if (PrimitiveHelpers.primitivesShareAttributeAccessor(primitive, comparePrimitive)) {
+                if (PrimitiveHelpers.primitivesHaveOverlappingIndexAccessors(gltf, primitive, comparePrimitive)) {
```
lilleyse: Create a function for this in `PrimitiveHelpers` that is called here and inside `markPrimitiveConflicts`.

lib/combinePrimitives.js
```diff
@@ -1,255 +1,252 @@
...
-    if (defined(gltf.accessors) && defined(gltf.bufferViews) && defined(gltf.buffers)) {
-        var meshes = gltf.meshes;
-        if (defined(meshes)) {
-            for (var meshId in meshes) {
-                if (meshes.hasOwnProperty(meshId)) {
-                    var mesh = meshes[meshId];
-                    var primitives = mesh.primitives;
-                    if (defined(primitives)) {
-                        mesh.primitives = combineMeshPrimitives(gltf, meshId, primitives);
+    PrimitiveHelpers.markPrimitiveConflicts(gltf, PrimitiveHelpers.getAllPrimitives(gltf));
```
lilleyse: It's a little awkward to save the conflicts in the pipeline object. If possible try to keep it in here.

lilleyse:
Thanks for the update @lasalvavida 
It's been a while since I looked at this so fill me in case my comments don't make sense. I'll do some testing tomorrow.

lilleyse:
I'm having issues converting this model: 
[dragon_low.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/567387/dragon_low.gltf.txt)

lib/PrimitiveHelpers.js
lasalvavida: However, this should use `AccessorReader` instead of `readAccessor`. I will make that change.

lilleyse: Ah right, thanks for the correction.

lasalvavida:
This is ready for review. Referring to my original benchmarks the stats are now:

|  | Original | Master | Master w/ combine primitives |
| --- | --- | --- | --- |
| File size | 10,414 KB | 6,091 KB | 6,023 KB |
| Nodes | 208 | 179 | 179 |
| Meshes | 109 | 163 | 61 |
| Draw Calls | 236 | 273 | 196 |

lilleyse:
This is really close. Every model I've tested works fine. Those numbers are nice!

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -3,81 +3,22 @@ var Cesium = require('cesium');
...
-    var attributes = primitive.attributes;
-    for (var attributeSemantic in attributes) {
-        if (attributes.hasOwnProperty(attributeSemantic)) {
-            if (attributeSemantic.indexOf(semantic) === 0) {
-                return attributeSemantic;
-            }
-        }
+    var semantics = getPrimitiveAttributeSemantics(primitive, semantic);
+    if (semantics.length <= 0) {
+        primitive.attributes[semantic] = createAccessor(gltf, packedLength, 'VEC3', WebGLConstants.FLOAT, WebGLConstants.ARRAY_BUFFER);
```
lasalvavida: Ah yes, you're right

lasalvavida:
Updated

lilleyse:
Nice, I'm glad to finally get this in! Also because of adjustments here the pipeline just runs a lot faster.

PR #165, opened 2016-08-29
Added support for glTF 1.0.1 normalized to quantizeAttributes

lasalvavida:
Adds an optional flag to use the glTF 1.0.1 normalized property for higher precision decode matrices.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### Next Release
+
+* `quantizedAttributes` has an optional `normalized` flag to use the glTF 1.0.1 `accessor.normalized` for a higher precision decode matrix.
+
 ### 0.1.0-alpha4 - 2016-08-25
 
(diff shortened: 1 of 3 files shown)


lasalvavida:
Used to generate the test model for https://github.com/AnalyticalGraphicsInc/cesium/pull/4247

lilleyse:
> @lilleyse, can you look at this when you get a chance?

Yes I'll check this out soon.

pjcozzi:
CC @lexaknyazev - glTF 1.0.1

pjcozzi:
I'm not following why this is needed.

lasalvavida:
This is one of the advantages of having glTF 1.0.1. By adding the normalized flag to a quantized accessor, the decode matrix doesn't need to be stored as 1.0/65535.0. This gives you the option to generate quantized models that decode with higher precision.

lilleyse:
+1 from me

pjcozzi:
OK, does this still strictly conform to the extension or does that need an update?

lasalvavida:
No updates needed, this strictly conforms to the extension.

PR #167, opened 2016-08-31
Added utility for changing accessor componentType

lasalvavida:
Added generic utility for changing accessor component type. This simplifies `compressIntegerAccessors` and will also be useful for adding uint32 indices support to combinePrimitives.

@lilleyse, could you look at this when you get a chance?


lilleyse:
Looks good.

PR #168, opened 2016-09-01
addDefaults doesn't add states.disable

### LINKED ISSUE #159: addDefaults.statesDefaults adds states.disable
This is not defined in the [glTF specification](https://github.com/KhronosGroup/glTF/tree/master/specification#states)

lasalvavida:
Resolves #159. After this change, pipeline models will pass validation: http://github.khronos.org/glTF-Validator/

@lilleyse or @pjcozzi  could you look at this when you get a chance? Should be a pretty easy merge.

### DIFF
--- a/lib/addDefaults.js
+++ b/lib/addDefaults.js
@@ -431,10 +431,6 @@ function statesDefaults(states) {
     if (!defined(states.enable)) {
         states.enable = [];
     }
-
-    if (!defined(states.disable)) {
-        states.disable = [];
-    }
 }
 
 function techniqueDefaults(gltf) {


lilleyse:
Yup that was easy.

emackey:
Hey guys, I think `states.disable` was from an old version of the spec, it's not there anymore.

https://github.com/KhronosGroup/glTF/blob/master/specification/schema/technique.states.schema.json

lasalvavida:
> Hey guys, I think states.disable was from an old version of the spec, it's not there anymore.

Yes, that's why we removed here.

emackey:
Wow, misread that, sorry!

PR #169, opened 2016-09-01
Add KHR_binary_glTF to extensionsUsed when exporting binary glTF

lasalvavida:
Another pretty simple change for passing validation: http://github.khronos.org/glTF-Validator/


lilleyse:
Looks good!

PR #170, opened 2016-09-02
Full yargs cli

lasalvavida:
Uses yargs to fully establish default settings, types, etc. Makes the command line `-h` flag a lot more useful, and easier to maintain.


lasalvavida:
@lilleyse Could you look at this when you get a chance?

bin/gltf-pipeline.js
```diff
@@ -1,55 +1,116 @@
...
-        '     --ao.toTexture: Bake AO to existing diffuse textures instead of to vertices. Does not modify shaders. Default: inactive.\n' +
-        '     --ao.groundPlane: Simulate a ground plane at the lowest point of the model when baking AO. Default: inactive.\n' +
-        '     --ao.ambientShadowContribution: Amount of AO to show when blending between shader computed lighting and AO. 1.0 is full AO, 0.5 is a 50/50 blend. Default: 0.5\n' +
-        '     --ao.quality: Valid settings are high, medium, and low. Default: low\n' +
-        '     --cesium, optimize the glTF for Cesium by using the sun as a default light source. \n';
-    process.stdout.write(help);
-    return;
-}
+var argv = require('yargs')
+    .usage('Usage: $0 -i inputPath -o outputPath')
```
lilleyse: Should the usage and example be like `node $0`?

lilleyse:
Using yargs in this way is it possible to allow processing a file without `-i` and `-o`?

lilleyse:
Looks good, just those comments.

lasalvavida:
This has been updated as well. 'node' has been added to the usage and help output. Input and output can also be passed implicitly as the first and second unmarked parameters without `-i` and `-o`. This will need a few updates if we merge #173 first.

pjcozzi:
Looks nice!

@lilleyse OK to merge?

lilleyse:
Only issue I'm noticing is it throws an error when no arguments are supplied. Should be a quick fix @lasalvavida and then I'll merge.

lasalvavida:
@lilleyse Could you look at this when you get a chance?

lilleyse:
Ah thanks for the reminder, I did review the latest changes but must have gotten sidetracked.

PR #172, opened 2016-09-08
Update b3dm spec

tfili:
b3dm files now have an additional field for binaryBatchTableLength so the header is now 24-bytes. The batchLength was also moved to the end of the header. See spec https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/TileFormats/Batched3DModel/README.md.

Once this is merged old b3dm files won't work. I'm thinking we don't need the backwards compatibility, but we can if anyone feels strongly about it.

### DIFF
--- a/bin/b3dm-pipeline.js
+++ b/bin/b3dm-pipeline.js
@@ -86,6 +86,8 @@ if (defined(directory)) {
 
 function processB3dm(b3dmPath) {
     var batchLength;
+    var batchTableByteLengthJson;
+    var batchTableByteLengthBinary;
     var batchTableByteLength;
     var batchTableBuffer;
     return readB3dm(b3dmPath)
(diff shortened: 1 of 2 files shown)


pjcozzi:
This PR and much of the `b3dm-pipeline` branch do not belong in this project; this is project is for glTF, and should not include 3D Tiles unless strictly required.

@lasalvavida let's discuss when you are in.

lasalvavida:
> This PR and much of the b3dm-pipeline branch do not belong in this project; this is project is for glTF, and should not include 3D Tiles unless strictly required.

Agreed, having a branch here was really only meant as a quick fix. This should probably be in 3D-Tiles-Tools, right?

lilleyse:
Yeah the changes are in https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/6

lasalvavida:
>Yeah the changes are in https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/pull/6

That's not quite the same. This is `b3dm` in -> `b3dm` out and this branch has some changes that should be in the main pipeline and some that should be separated out since they're `b3dm` specific. Will open PRs.

PR #173, opened 2016-10-10
Hard normals

tfili:
@lasalvavida Just opening a pull request here so we can keep track of what needs to be done to get this in.

CC @pjcozzi


lasalvavida:
This is ready for review. Adds support for stripping off existing normals with `-r` and generating face normals with `-f`. I've run 2CylinderEngine through and below are some screenshots to highlight the difference between the existing smooth normals and the added hard normals.

Smooth Normals:

2-cylinder-smooth-normals

Hard Normals:

2-cylinder-hard-normals

lasalvavida:
Note: the material generation in `generateNormals` may be changed or removed entirely when I finish adding model material common generation, but for now I think that it is acceptable.

pjcozzi:
Thanks @lasalvavida.

@tfili once you confirm that this meets your needs, @lilleyse can review and merge.

tfili:
@lasalvavida this looks pretty good. The one issue we saw was you can't run this twice in a row because it either generates an invalid technique or we are missing an if check when processing a technique.

tfili:
@lasalvavida After a little digging the culprit seems to be the remove normals step. It removes the technique from the material so it can be regenerated, but there are stages before that happens that expect the technique to exist.

lasalvavida:
Okay, I should have some time tomorrow to take a look at this.

lasalvavida:
This should be fixed now

lilleyse:
@tfili let me know when this is all good from your end and I'll do a review.

pjcozzi:
> @tfili let me know when this is all good from your end and I'll do a review.

@tfili is this ready for review?

tfili:
@pjcozzi @lilleyse This looks good to me.

pjcozzi:
@lilleyse please merge when you are happy with the code.

lib/generateNormals.js
```diff
@@ -20,24 +40,173 @@ module.exports = generateNormals;
...
+            }
+            generateMaterial(gltf, primitive, generatedMaterials);
         }
     }
+
+    removeAccessors(gltf);
+    removeBufferViews(gltf);
+    removeBuffers(gltf);
+    mergeBuffers(gltf);
+    mergeDuplicateVertices(gltf);
```
lilleyse: Since `mergeDuplicateVertices` is the next step in `Pipeline` we can remove it from here.

lib/techniqueAttributeForSemantic.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
```
lilleyse: Is this file used?

tfili:
@lilleyse I fixed you last 2 suggestions. Let me know if there is anything else.

lilleyse:
Thanks @tfili that should be it.

PR #174, opened 2016-10-11
Added _3DTILESDIFFUSE semantic

lilleyse:
@pjcozzi 
Is it fine that these be added by default?


pjcozzi:
@lilleyse IIRC we discussed offline that this shouldn't be part of `gltf-pipeline` since it is 3D Tiles specific.  I thought we planned to put it a 3D Tiles specific COLLADA2GLTF / obj2gltf workflow...maybe `3d-tiles-tools` is the right place so we don't have to update every glTF converter?

lilleyse:
Updated.
As discussed offline, it only adds the `_3DTILESDIFFUSE` semantic if `optimizeForCesium` is true.

lilleyse:
Also CC: https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/issues/21

pjcozzi:
Is there a unit test for `optimizeForCesium`?  Can we add an expectation for this new case?

Also update CHANGES.md.

lilleyse:
Added unit tests and updated CHANGES.md.

PR #175, opened 2016-10-21
Preserve the existence of non-default shader attributes

lasalvavida:
Edit: The scope of this pull request has changed and the title has been updated to reflect that. I've kept the old text here as a record.

~~Single b3dm files can be processed via the glTF pipeline with these changes. This is based on the hard-normals branch, so that is what it is diffed against currently.~~

~~I am open to suggestions, but I think this is the best way to integrate b3dm support with gltf-pipeline given the current state of the public API, since it requires access to some of the internal machinery of gltf-pipeline.~~

~~This makes it easy to write a script to process tilesets concurrently using gltf-pipeline.~~

``` bash
#!/bin/bash
for f in $(find $1/ -name '*.b3dm'); do
    echo $f
    node ./gltf-pipeline.js -i $f -o $f -f -z &
    NPROC=$(($NPROC+1))
    if [ "$NPROC" -ge 4 ]; then
        wait
        NPROC=0
    fi
done
```


tfili:
@lilleyse Can we get this reviewed? It would be helpful to get this into `hard-normals` then get that into `master` soon.

lilleyse:
I made some small fixes.

Talked with @tfili offline. We aren't sure if these b3dm steps belong in the gltf-pipeline. The main tangle point is when normals are generated a new shader is built, and the `BATCH_ID` needs to be re-inserted in the techniques and shaders. Technically this could be done as a post step in a b3dm specific pipeline.

lilleyse:
@pjcozzi what do you think about keeping this in here vs. 3d-tiles-tools or some other repo?

lasalvavida:
I've been thinking about this a bit, and it may be a better solution to just make sure that BATCH_ID never gets removed in the first place. 

In a more general sense, I think if we generate a new shader from modelMaterialsCommon, it should declare all of the semantics from the primitive, even if it doesn't use them in the shader.

tfili:
@lasalvavida Also one thing I noticed, the way the b3dm handling is currently done I can't call it from another node script. It would be nice if I could call a function in the pipeline like processJSONWithExtras, but all the removing of b3dm headers are in the `bin/gltf-pipeline.js`.

lasalvavida:
I've finished the first part of this. In modelMaterialsCommon, we now preserve the existence of primitive semantics that don't exist by default in the generated shader. I think this is the correct approach and is not a 3d-tiles specific feature, since anyone doing dynamic shader modification using some non-default shader attribute would need this as well.

I'm still looking at the integration of reading and writing b3dms.

lasalvavida:
However, I think I'm going to revert the b3dm changes in this PR for now so we can get it and face normals merged sooner. After this change it would be pretty straightforward to go either path: gltf-pipeline supporting more file types or a 3d-tiles-tool that uses gltf-pipeline.

lasalvavida:
Okay, this should be ready for a look.

lib/modelMaterialsCommon.js
```diff
@@ -727,6 +729,85 @@ function lightDefaults(gltf) {
     }
 }
 
+function getShaderVariable(accessor) {
```
pjcozzi: Can you update the spacing in this file to be consistent with the rest of the codebase?

pjcozzi:
Is this already tested by other unit tests?

lasalvavida:
> Is this already tested by other unit tests?

Doesn't look like it, I'll add a test.

lib/modelMaterialsCommon.js
```diff
@@ -727,6 +729,85 @@ function lightDefaults(gltf) {
...
+  if (defined(technique)) {
+    programId = technique.program;
+    program = programs[programId];
+  }
+  var vertexShaderId;
+  var vertexShader;
+  if (defined(program)) {
+    vertexShaderId = program.vertexShader;
+    vertexShader = shaders[vertexShaderId];
+  }
```
lilleyse: Won't a lot of these checks always be true since it has gone through the modelMaterialsCommon? It should always have a material/technique/program. This also lets you simplify the loop.

lib/modelMaterialsCommon.js
```diff
@@ -727,6 +729,85 @@ function lightDefaults(gltf) {
...
+
+  for (var semantic in attributes) {
+    if (attributes.hasOwnProperty(semantic)) {
+      if (!defined(techniqueParameterForSemantic(technique, semantic))) {
+        if (defined(technique)) {
+          var accessorId = attributes[semantic];
+          var accessor = accessors[accessorId];
+          var lowerCase = semantic.toLowerCase();
+          var attributeName = 'a_' + lowerCase;
```
lilleyse: For the `a_batchId` case this will produce `a_batchid` instead.

lib/modelMaterialsCommon.js
```diff
@@ -727,6 +729,85 @@ function lightDefaults(gltf) {
...
+    if (meshes.hasOwnProperty(meshId)) {
+      var mesh = meshes[meshId];
+      var primitives = mesh.primitives;
+      var primitivesLength = primitives.length;
+      for (var i = 0; i < primitivesLength; i++) {
+        ensureSemanticExistenceForPrimitive(gltf, primitives[i]);
+      }
+    }
+  }
+}
```
lilleyse: If the function above be can be shorted enough, I wouldn't see a problem combining the two functions.

lib/modelMaterialsCommon.js
tfili: So I don't think this matters. We replace the `program.attribute`, `technique.attribute`, `technique.parameter` and in the shader with the same case, so Cesium should handle it correctly because of the semantic.

lilleyse: In Cesium we build the shader with the assumption that it's called `a_batchId`: https://github.com/AnalyticalGraphicsInc/cesium/blob/3d-tiles/Source/Scene/Cesium3DTileBatchTable.js#L498. Maybe this isn't a good assumption though - instead should we use whatever parameter has the `BATCHID` semantic?

lilleyse:
The `a_batchId` issue is addressed in this Cesium PR: https://github.com/AnalyticalGraphicsInc/cesium/pull/4583/files

tfili:
@lilleyse The test has been added and I believe all the PR comments have been addressed.

lib/modelMaterialsCommon.js
```diff
@@ -732,6 +734,69 @@ function lightDefaults(gltf) {
...
+    var attributes = primitive.attributes;
+    var material = materials[primitive.material];
+    var technique = techniques[material.technique];
+    var program = programs[technique.program];
+    var vertexShader = shaders[program.vertexShader];
+
+    for (var semantic in attributes) {
+        if (attributes.hasOwnProperty(semantic)) {
+            if (!defined(techniqueParameterForSemantic(technique, semantic))) {
+                if (defined(technique)) {
```
lilleyse: Check isn't needed.

lib/modelMaterialsCommon.js
```diff
@@ -732,6 +734,69 @@ function lightDefaults(gltf) {
...
+                    var accessorId = attributes[semantic];
+                    var accessor = accessors[accessorId];
+                    var lowerCase = semantic.toLowerCase();
+                    var attributeName = 'a_' + lowerCase;
+                    technique.parameters[lowerCase] = {
+                        semantic: semantic,
+                        type: accessor.componentType
+                    };
+                    technique.attributes[attributeName] = lowerCase;
+                    if (defined(program)) {
```
lilleyse: Same comment.

lib/modelMaterialsCommon.js
```diff
@@ -732,6 +734,69 @@ function lightDefaults(gltf) {
...
+                        program.attributes.push(attributeName);
+                        if (defined(vertexShader)) {
```
lilleyse: Same comment.

specs/lib/addDefaultsSpec.js
```diff
@@ -569,6 +569,73 @@ describe('addDefaults', function() {
...
+                    },
+                    "name": "material1"
+                }
+            }
+        };
+
+        var gltfClone = clone(gltf);
+        addDefaults(gltfClone);
+
+        // Uses the Cesium sun as its default light source
```
lilleyse: Remove this comment.

tfili:
@lilleyse I think that should be everything.

PR #179, opened 2016-10-25
Use typings instead of rolling our own TypeScript definition solution

mramato:
This gets rid of the TypeScriptDefinitions directory and associated gulp task. Instead it uses the typings module which is the official way to manage TS definitions. This now creates a typings folder at npm install type and WebStorm has been configured to simply point to that.

Not only is this cleaner, but the actual intellisense is much better at WebStorm should provide even more help now.

Also removed the request dependency because it's no loner used.

Also added missing entries to npmignore and gitignore.

As part of this change, I also updated all npm modules to there latest version, this is in preparation for turning on greenkeeper.


lilleyse:
Thanks @mramato this does seem better.

PR #183, opened 2016-11-11
Parsing arguments exposed as a separate function

lasalvavida:
Allows API use applications to use the gltf-pipeline argument parsing to create objects for use with the API functions.


lilleyse:
This all looks good.

PR #184, opened 2016-11-18
Transform oct-encoded normals

lasalvavida:
Decode oct-encoded normals when transforming and then re-encode. With these changes, the attached model transforms correctly.

[batched.glb.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/600404/batched.glb.txt)


lilleyse:
Thanks! Do we need a similar fix for quantized attributes?

lasalvavida:
> Thanks! Do we need a similar fix for quantized attributes?

Yup. This was a good short-term fix. I think the better long term one may be to abstract encoding and decoding into `AccessorReader`.

PR #185, opened 2016-11-18
Added command line generation of KHR_materials_common

lasalvavida:
All of the pieces are in place now to open this PR. This is being used to regenerate the 1.0 sample models before the migration to 1.1, and will likely be used for that as well.


lasalvavida:
After experimenting a bit, this still needs to create lights from the lights in the technique. That should be added shortly.

lasalvavida:
This now generates lights from techniques.

lasalvavida:
I found an issue where Box doesn't quite have the right lighting post-generation. I think this is because there is no ambient light generated in the KHR_materials_common sample model, but I'm not sure when it's okay to not have it since it is generated for some of the others.

lilleyse:
Good timing with this PR: https://github.com/AnalyticalGraphicsInc/cesium/issues/4669

I'll review this shortly.

lib/generateModelMaterialsCommon.js
```diff
@@ -0,0 +1,157 @@
...
+            var techniqueParameters = technique.parameters;
+            if (defined(techniqueParameters.ambient) && !defined(lights.defaultAmbient)) {
+                lights.defaultAmbient = {
+                    ambient: {
+                        color: [1, 1, 1]
+                    },
+                    name: 'defaultAmbient',
+                    type: 'ambient'
+                };
+            }
```
lilleyse: > I found an issue where Box doesn't quite have the right lighting post-generation. I think this is because there is no ambient light generated in the KHR_materials_common sample model, but I'm not sure when it's okay to not have it since it is generated for some of the others.

I'm not sure why an ambient light needs to be added. When Cesium or other engines encounter khr_materials_common with an ambient value wouldn't they just modify the shader?

lilleyse:
When going through the API rather than the command line, the technique/doubleSided are undefined if not supplied, but they should be set to the defaults somewhere.

lib/generateModelMaterialsCommon.js
```diff
@@ -0,0 +1,157 @@
...
+    var nodes = gltf.nodes;
+    var techniques = gltf.techniques;
+    var jointCountForMaterialId = getJointCountForMaterials(gltf);
+
+    var gltfExtensions = gltf.extensions;
+    if (!defined(gltfExtensions)) {
+        gltfExtensions = {};
+        gltf.extensions = gltfExtensions;
+    }
+    materialsCommon = {};
```
lilleyse: Does this need `deleteExtras`?

lasalvavida:
> Just having issues with this model:

That's the model. It has ambient set to [1,1,1,1]. If you set it to [0,0,0,1] you get more what you would expect.
dragon_low

lib/generateModelMaterialsCommon.js
lasalvavida: It looks like we have to specify the light for KHR_materials_common, or the ambient part of the shader doesn't get generated. My solution for this is to detect if the technique uses the ambient parameter and create an ambient light in the extension if it does, assuming that since the technique uses the parameter, the shader does as well.

lasalvavida:
@lilleyse, this should be ready to go now.

lilleyse:
> That's the model. It has ambient set to [1,1,1,1]. If you set it to [0,0,0,1] you get more what you would expect.

Ah didn't even look there.

lilleyse:
Looks good, Thanks!

PR #191, opened 2016-12-05
Update version

lasalvavida:
@lilleyse and @pjcozzi could you look at this when you get a chance? ~~This refactors any glTF 0.8 ->1.0 from gltfDefaults out into a global stage called `updateVersion`.~~ (This is now in #223) This stage also includes tools for converting from glTF 1.0 -> 2.0.
~~but `lib/Pipeline.js` still targets 1.0 by default. This will be changed once glTF 1.1 is fully released.~~

See https://github.com/KhronosGroup/glTF/issues/605 for the full list of changes from glTF 1.0 -> 1.1

@lexaknyazev, if you have time I'd appreciate your input here as well.

---

**TODO**

* [x] Test new stage with all glTF sample models


pjcozzi:
@lasalvavida have you tested this with all the sample models?

lasalvavida:
> @lasalvavida have you tested this with all the sample models?

Not yet.

pjcozzi:
>> @lasalvavida have you tested this with all the sample models?
> Not yet.

Added to the tasklist at the top of this PR.  Please add any other items there needed before we should merge this.  Updating the entire pipeline to 1.1 can be a separate PR if you prefer.

pjcozzi:
@lexaknyazev if you are able to look at this, the two files to review are:
* [updateVersion.js](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/updateVersion/lib/updateVersion.js)
* [updateVersionSpec.js](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/updateVersion/specs/lib/updateVersionSpec.js) - tests

specs/lib/updateVersionSpec.js
```diff
@@ -0,0 +1,265 @@
...
+            attribute: 'TEST_ATTRIBUTE'
+        });
+        expect(technique.program).toBe('program');
+        expect(technique.uniforms).toEqual({
+            uniform: 'TEST_UNIFORM'
+        });
+        expect(technique.states).toEqual(['TEST_STATE']);
+    });
+
+    it('updates glTF from 1.0 to 1.1', function() {
```
pjcozzi: Can this also test:
* [ ] `glExtensionsUsed` and `extensionsRequired`
* [ ] `buffer.type` removal
* [ ] `technique.functions.scissor` and `technique.states.enable = 3089` removal
* [ ] Adds all properties that were formerly not required, e.g., `asset`, `buffer.byteLength`, `bufferView.byteLength`, etc.
* [ ] `technique.states.functions.blendColor` / `technique.states.functions.depthRange` parameter changes.

In general, can you look at each bullet in https://github.com/KhronosGroup/glTF/issues/605 and confirm that it is handled?  There are others that I did not mention above.

There are some cases where an exact update won't be possible, and we may have to make a breaking change (to the model itself) and provide a log to the user.  For example, we may have to prepend an underscore to an application-specific parameter semantic or change `camera.perspective.yfov = 0.0` to `camera.perspective.yfov = 1.0`.  In practice, most of these edge cases will not come up.

It is likely cleaner to break this into multiple tests for some of these edge cases.

lib/updateVersion.js
```diff
@@ -0,0 +1,355 @@
...
+        version = gltf.asset.version;
+        if (version === targetVersion) {
+            break;
+        }
+        updateFunction = updateFunctions[version];
+    }
+    return gltf;
+}
+
+function glTF_0_8_to_glTF_1_0(gltf) {
```
pjcozzi: Even for private functions, we don't put underscore in names.

lasalvavida: Do you have a suggested alternative? I was having a hard time making it still readable without underscores.

pjcozzi:
Good start, thanks @lasalvavida!

lib/updateVersion.js
pjcozzi: ha.  I had a few then I deleted them.  I dunno:
* `glTF08To10`
* `glTF08ToglTF10`

specs/lib/updateVersionSpec.js
lasalvavida: > glExtensionsUsed and extensionsRequired

Those properties have no analogue in older versions of glTF; they are just new properties. I can't take a glTF1.0 `extensionsUsed` and make it `extensionsRequired` because we don't know which ones were just used vs. required. As such, I don't think there is any safe updating that can be done here.

> buffer.type removal

Thank you, I did miss that one

> technique.functions.scissor and technique.states.enable = 3089 removal

Agreed, I will add this.

>  Adds all properties that were formerly not required, e.g., asset, buffer.byteLength, bufferView.byteLength, etc.

I'll do `byteLength`, but adding the asset if it doesn't exist is handled by `addDefaults`.

>  technique.states.functions.blendColor / technique.states.functions.depthRange parameter changes.

I think this is in the same realm as camera.perspective properties. If someone has a property that is set to an invalid value, I don't think that it is safe to just change the value to a valid one. Even though it was never explicitly forbidden in 1.0, those restrictions were still imposed by WebGL. I can maybe log a warning or something, but I don't think we can just change the value.

lasalvavida: Actually adding asset if it is not defined is a bit more complex. That being said, I think the existing behavior is the correct approach. If a glTF 0.8 without asset defined is updated, but it has `gltf.version`, the asset property will be added. However, for any other this will throw an exception, because if asset isn't defined, then `asset.version` isn't defined, and there is no safe way to determine the version of the glTF, and what changes need to be made.

pjcozzi: > I can't take a glTF1.0 extensionsUsed and make it extensionsRequired because we don't know which ones were just used vs. required

I think we do know which were required; it is the extensions that are now required, which I suppose is all of the current extensions.

>> technique.states.functions.blendColor / technique.states.functions.depthRange parameter changes.
> I think this is in the same realm as camera.perspective properties.

pjcozzi: >> technique.states.functions.blendColor / technique.states.functions.depthRange parameter changes.

> I think this is in the same realm as camera.perspective properties.

It's a tough call, but the 1.0 -> 1.1 stage should output a valid 1.1 model so I would be fine with clamping the value and providing a warning log.  There's an OK chance that is what the user wants; otherwise, the user would have to edit the model afterward 100% of the time.

pjcozzi: > If it's undefined, we could assume 1.0. I'm just concerned that this could cause some strange behavior down the line somewhere.

Probably OK.  In practice, I suspect virtually 0% of users use 0.8, and post 1.0, `asset` is required.

lasalvavida:
This is updated from the discussion above.

pjcozzi:
Great, will aim to review tomorrow morning.

pjcozzi:
Do we handle this one from https://github.com/KhronosGroup/glTF/issues/605:

> Attribute parameters can't have a value defined in the technique or parameter, #563 (comment).

I think we should strip the `value` and warn in this case.

pjcozzi:
Is this taken into account?

> technique.parameter.count can only be defined when the semantic is JOINTMATRIX or an application-specific semantic is used. It can never be defined for attribute parameters; only uniforms, d2f6945

pjcozzi:
Is this taken into account?

> technique.parameter.semantic is required when the parameter is an attribute, 28e113d

In practice, it is likely to be rare so no need to over invest here.

We could either (1) warning or (2) make a best attempt to map attribute names to semantics and still warn.

pjcozzi:
I reviewed https://github.com/KhronosGroup/glTF/issues/605 and that is all my comments about how this maps to the 1.1 spec.  Please keep an eye on 1.1 issues in the glTF repo for any last-minute updates.

I'll also take a quick look at this code now.

lib/Pipeline.js
```diff
@@ -84,6 +85,9 @@ Pipeline.processJSONWithExtras  = function(gltfWithExtras, options) {
     if (options.removeNormals) {
         removeNormals(gltfWithExtras);
     }
+    updateVersion(gltfWithExtras, {
+        targetVersion: '1.0'
```
pjcozzi: Should this be `1.1`?

lib/updateVersion.js
```diff
@@ -0,0 +1,632 @@
...
+    }
+}
+
+function stripWebGLRevisionNumber(gltf) {
+    var asset = gltf.asset;
+    var profile = asset.profile;
+    if (defined(profile)) {
+        var version = profile.version;
+        if (defined(version)) {
+            profile.version = version[0] + '.' + version[2];
```
pjcozzi: Shouldn't `version[2]` be `version[1]`?

https://github.com/KhronosGroup/glTF/issues/701

And to be safe perhaps the check should be
```javascript
if (defined(version) && (version.length > 2)) {
```

lib/updateVersion.js
```diff
@@ -0,0 +1,632 @@
...
+                    if (defined(attributes)) {
+                        var semantics = Object.keys(attributes);
+                        var semanticsLength = semantics.length;
+                        for (var j = 0; j < semanticsLength; j++) {
+                            var semantic = semantics[j];
+                            if (semantic === 'TEXCOORD') {
+                                attributes.TEXCOORD_0 = attributes[semantic];
+                                delete attributes[semantic];
+                            }
+                            if (semantic === 'COLOR') {
```
pjcozzi: minor, but `} else if (semantic === 'COLOR') {` would be slightly better style.

lib/updateVersion.js
```diff
@@ -0,0 +1,632 @@
...
+}
+
+var knownSemantics = {
+    POSITION: true,
+    NORMAL: true,
+    TEXCOORD: true,
+    COLOR: true,
+    JOINT: true,
+    WEIGHT: true,
+    BATCHID: true
```
pjcozzi: `BATCHID` is part of 3D Tiles, not glTF, so the 3D Tiles spec will need to name this `_BATCHID` (please update the spec if it isn't already), and then we can have backwards compatibility loading in Cesium with a warning (also please open PR if this isn't done).

pjcozzi:
@lasalvavida just those comments.  What is your plan from here?

Perhaps get this fully into shape, but then do not merge until we have the samples converted and Cesium loader so that we know the full stack works together?

lasalvavida:
I'll go through all of these as soon as possible. 

Currently my plan is to integrate a bundled version of gltf-pipeline that contains `parseModelMaterialsCommon`, `addDefaults`, and `updateVersion` into mainline Cesium with `targetVersion` set to `1.0`. This should make testing for `1.1` support pretty simple (just set `targetVersion` to `1.1` and see what breaks).

lib/Pipeline.js
lasalvavida: For now, this is set to `1.0` intentionally since I don't think we want to be generating `1.1` from master by default just yet. Though now that I think about it, command flag for `1.1` may be a better solution.

lib/updateVersion.js
lasalvavida: This is correct as is, but a length check is a good idea.

| 0 | 1 | 2 | 3 | 4 |
|---|---|---|---|---|
| 1 | . | 0 | . | 3|

and we want `1.0`, so `version[0] + '.' + version[2]`.

pjcozzi: Ah, yes!

pjcozzi:
> Currently my plan is to merge this into master here and integrate a bundled version of gltf-pipeline that contains parseModelMaterialsCommon, addDefaults, and updateVersion into mainline Cesium with targetVersion set to 1.0. This should make testing for 1.1 support pretty simple (just set targetVersion to 1.1 and see what breaks).

Sounds good.

lasalvavida:
> technique.parameter.count can only be defined when the semantic is JOINTMATRIX or an application-specific semantic is used. It can never be defined for attribute parameters; only uniforms, d2f6945

>Attribute parameters can't have a value defined in the technique or parameter, #563 (comment).

These are added.

> technique.parameter.semantic is required when the parameter is an attribute, 28e113d

> Properties in technique.parameters must be defined in technique.uniforms or technique.attributes,
#563 (comment).

I don't see a reasonable way to resolve these and it isn't an issue I ran into while upgrading the existing sample models. I think they're best left to the validator.

This is ready for review. I've tested this pretty extensively on the Cesium side and it seems to work well.

lasalvavida:
@pjcozzi

pjcozzi:
>> technique.parameter.semantic is required when the parameter is an attribute, 28e113d
>>
>> Properties in technique.parameters must be defined in technique.uniforms or technique.attributes,
#563 (comment).
>
> I don't see a reasonable way to resolve these and it isn't an issue I ran into while upgrading the existing sample models. I think they're best left to the validator.

Right, gltf-pipeline can't resolve these; however, it should still warn that it knows it is not outputting valid glTF 2.0 because it can't; otherwise, we may get bug reports on how gltf-pipeline output doesn't pass validation and it will be harder to track down.

lib/updateVersion.js
```diff
@@ -10,11 +15,13 @@ var defined = Cesium.defined;
 module.exports = updateVersion;
 
 var updateFunctions = {
-    '0.8': glTF08to10
+    '0.8' : glTF08to10,
+    '1.0' : glTF10to20,
+    '2.0' : null
```
pjcozzi: Can we used `undefined` here?  We generally don't use `null`.

lib/updateVersion.js
```diff
@@ -160,6 +174,540 @@ function glTF08to10(gltf) {
...
+        if (defined(node.scale)) {
+            Cartesian3.unpack(node.scale, scratchScale);
+        }
+        Matrix4.fromTranslationQuaternionRotationScale(scratchTranslation, scratchRotation, scratchScale, scratchMatrix4);
+        return scratchMatrix4;
+    } else {
+        return Matrix4.IDENTITY;
+    }
+}
+
```
pjcozzi: Minor, extra blank line.

pjcozzi:
I might have misread CHANGES.md (and then mis-updated it), but it sounded like all the pipeline stages now run on glTF 2.0.  Is that correct?  I didn't see any changes.

We actually probably want to do this last so since once we go to 2.0, 1.0 will not be supported.

pjcozzi:
Can you expose `updateVersion` in the CLI?  Just the target version needs to be provided and it must be `>=` the input version.

lasalvavida:
> but it sounded like all the pipeline stages now run on glTF 2.0. Is that correct?

That is correct. See changes to `addDefaults`, `removeUnusedProperties`, `quantizeAttributes`, `generatedModelMaterialsCommon`, etc.

pjcozzi:
Also, let's get the full stack updated: here/Cesium and COLLADA2GLTF (obj2gltf can come last, and @lilleyse might do it if available), then there will be some last-minute 2.0 updates we need to add; please review https://github.com/KhronosGroup/glTF/labels/2.0.

pjcozzi:
>> but it sounded like all the pipeline stages now run on glTF 2.0. Is that correct?

> That is correct. See changes to addDefaults, removeUnusedProperties, quantizeAttributes, generatedModelMaterialsCommon, etc.

OK, I am surprised that is all that is required.  So gltf-pipeline no longer works on 1.0 models?  It's OK, I'm just trying to get my head around the current state.

lasalvavida:
> So gltf-pipeline no longer works on 1.0 models?

That is correct, which is why I'm not sure if we want to do this:

> Can you expose updateVersion in the CLI? Just the target version needs to be provided and it must be >= the input version.

Since some stages may not do what you expect on non glTF 2.0 models.

lasalvavida:
I also found some 0.8 -> 1.0 issues in my Cesium testing, and I will add those fixes shortly

pjcozzi:
>> So gltf-pipeline no longer works on 1.0 models?

> That is correct, which is why I'm not sure if we want to do this:

>> Can you expose updateVersion in the CLI? Just the target version needs to be provided and it must be >= the input version.

> Since some stages may not do what you expect on non glTF 2.0 models.

OK, let's document prominently in README.md that only glTF 2.0 is output and glTF 1.0 models are converted to 2.0.  No need for any console output for this.

lasalvavida:
I am closing this here and will re-open as the 2.0 pull request

Issue #189, opened 2016-11-30
Redecleration of `gltf_signNotZero`, `gltf_octDecode` when encoding normals


bkuster:
When encoding normals, but not removing them before hand, the resulting binary gltf cannot be rendered due to `ERROR 0:17 'gltf_signNotZero : function already has a body` (and others in the same fashion, including the `gltf_octDecode` `gltf_oct_dec_a_nromal` `gltf_oct_dec_normal` `gltf_oct_dec_gltf_oct_dec_a_normal` declarations) in Cesium.

Removing the normals before hand or not encoding them works.

The issue has arisen when optimizing a _gltf_ converted with the [COLLADA2GLTF converter v1.0](https://github.com/KhronosGroup/COLLADA2GLTF/) from [this](https://3dwarehouse.sketchup.com/model.html?id=f923282a793a5c71640b2801cae98bff) kmz's collada on an Ubuntu 16.04 machine. The `-b -n -c` flags where set.

pjcozzi:
Thanks for the report, @bkuster!

bkuster:
sure thing, I've been trying to do some debuging myself, so far I've come to the conclusion, that the culprit lies in the `createCombinedVertexShader` of the `Cesium.ShaderSource`. This function adds a ton of newlines and repeats the code block a few times, why I can't figure out yet

bkuster:
I think the issue lies in the generated GLTF. Even though the shaders have a different names, two pairs are exactly the same.

pjcozzi:
@bkuster a PR would be awesome, thanks!  We use the same [Contributor License Agreement (CLA)](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla) as Cesium so please send in one of those as well.  Let me know if you have any questions.

lasalvavida:
Closed by #192

PR #192, opened 2016-12-06
patchedShaders in octEncodedNormals prevents redefinition

bkuster:
#189

### DIFF
--- a/lib/octEncodeNormals.js
+++ b/lib/octEncodeNormals.js
@@ -142,7 +142,7 @@ function patchProgram(gltf, programId, normalVariableName, patchedShaders) {
     var shaders = gltf.shaders;
     var program = programs[programId];
     var vertexShaderId = program.vertexShader;
-    if (patchedShaders[vertexShaderId]) {
+    if (defined(patchedShaders[vertexShaderId])) {
       return;
     }
     var vertexShader = shaders[vertexShaderId];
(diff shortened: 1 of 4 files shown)


bkuster:
hold on, this still fails

bkuster:
this is ready

pjcozzi:
Thanks @bkuster!

@lilleyse could you look at this?

lasalvavida:
@bkuster This looks good to me, thank you. Could you please add a test with two programs that reference the same shader?

Also, it doesn't need to be part of this PR, but it is worth mentioning that `compressTextureCoordinates` will have this issue as well.

lib/octEncodeNormals.js
```diff
@@ -136,14 +137,18 @@ function patchTechnique(gltf, techniqueId) {
     }
 }
 
-function patchProgram(gltf, programId, normalVariableName) {
+function patchProgram(gltf, programId, normalVariableName, patchedShaders) {
     var programs = gltf.programs;
     var shaders = gltf.shaders;
     var program = programs[programId];
     var vertexShaderId = program.vertexShader;
+    if (patchedShaders[vertexShaderId]) {
```
lasalvavida: This should be `defined(patchedShaders[vertexShaderId])`. 

I know this works as-is, but `defined` is used throughout in this way to be more verbose.

bkuster:
updated and wrote a _simple_ spec. I'll see what I can find in the `compressTextureCoordinates`

bkuster:
I went ahead and just applied the same fix to the compressed coordinates. this is ready

lasalvavida:
Yup, looks good to me, thanks @bkuster!

PR #193, opened 2016-12-07
Update npm modules

mramato:
1. Update npm modules to latest version and switch to using `^` for versions, as discussed recently in
https://github.com/AnalyticalGraphicsInc/cesium/issues/4697
2. Remove `datauri`, `mkdirp`, and `promise` dependencies because they were either unused or easily replaced.
3. Update travis to test both Node 4 and 6.

Assuming this goes green, it would be nice to publish a new version to npm.


lasalvavida:
Looks good to me, thanks @mramato.

PR #194, opened 2016-12-08
Write shader fix

lilleyse:
@mramato @lasalvavida 

Will publish 0.1.0-alpha8 right after this is merged.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 Change Log
 ==========
 
+### 0.1.0-alpha8 - 2016-12-08
+* Fixed issue with embedding base64 encoded shader strings inside the glTF.
+
 ### 0.1.0-alpha7 - 2016-12-08
 * `modelMaterialsCommon` renamed to `processModelMaterialsCommon`.
 * Added `generateModelMaterialsCommon` and command line `kmc` flags for generating models with the `KHR_materials_common` extension.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 Change Log
 ==========
 
-### 0.1.0-alpha8 - 2016-12-08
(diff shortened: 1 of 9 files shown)


lasalvavida:
Why is this change necessary? I thought pipelineExtras source was always a Buffer? If it isn't for some reason, we should probably fix whatever made it a string.

lasalvavida:
Taking a quick look, it looks like `compressTextureCoordinates` and `octEncodeNormals` write back the shader as a string. I don't know the use case that generated an error, but I suspect those are probably the culprits. They should write back a buffer instead of the modified string.

mramato:
Thanks @lasalvavida, the library we were using was always creating a buffer so it looks like it was hiding the problem you describe.  @lilleyse, since the published version now has this issue; can we get this fixed and re-published soon?

lilleyse:
Fixed the problem areas.

lasalvavida:
Looks good to me, merge when ready

mramato:
Unit test would be good, since this slipped past the ones we had.  Or at least write up an issue to add a test for this if there's simply no time right now.

lasalvavida:
Good point. Checking for the buffer in the specs for compressTextureCoordinates and octEncodeNormals should be sufficient I think

lilleyse:
Added tests.

lasalvavida:
Looks good, merge and publish when ready

mramato:
@lasalvavida you are free to merge PRs when you are confident they are good.  We want to foster a culture where anyone with write access can merge (they just assume partial responsibility if it's a bad change).  In either case, thanks again for catching the root cause.

mramato:
Thanks @lilleyse!

lilleyse:
Published 0.1.0-alpha8

PR #195, opened 2016-12-14
Add empty data uri if it is missing

lilleyse:
I encountered a model whose shader is defined in the binary-gltf body but doesn't set a data-uri. I believe this is technically invalid according to the spec, but it is worth handling this case anyways.

@lasalvavida


lasalvavida:
Looks good to me, thanks @lilleyse!

PR #198, opened 2016-12-19
Embed all binary textures

bagnell:
Embed all binary textures. Only check if they are supported by Jimp when using AO.

### DIFF
--- a/lib/loadGltfUris.js
+++ b/lib/loadGltfUris.js
@@ -115,10 +115,9 @@ function isTransparent(image) {
 function readImageFromFile(object, name, uriPath) {
     return fsReadFile(uriPath)
         .then(function(data) {
-            var ext = path.extname(uriPath);
             object.extras._pipeline.source = data;
-            object.extras._pipeline.extension = ext;
-            if (name === 'images' && ext !== '.ktx' && ext !== '.gif') {
+            object.extras._pipeline.extension = path.extname(uriPath);
+            if (name === 'images') {
                 return generateJimpImage(object);
             }
         });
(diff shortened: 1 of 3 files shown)


lib/loadGltfUris.js
```diff
@@ -115,9 +115,10 @@ function isTransparent(image) {
 function readImageFromFile(object, name, uriPath) {
     return fsReadFile(uriPath)
         .then(function(data) {
+            var ext = path.extname(uriPath);
             object.extras._pipeline.source = data;
-            object.extras._pipeline.extension = path.extname(uriPath);
-            if (name === 'images') {
+            object.extras._pipeline.extension = ext;
+            if (name === 'images' && ext !== '.ktx' && ext !== '.gif') {
```
lilleyse: Should this check stay in `generateJimpImage`? If the image is embedded it will not reach this code.

bagnell: Generating a binary glTF reaches this code.

lilleyse: I mean when the gltf is loaded it won't get there.

lilleyse: Like this file with an embedded gif - Jimp will try to load it but fail: 
[BoxTextured-optimized.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/662235/BoxTextured-optimized.gltf.txt)

bagnell: Ah, thanks. Updated.

lilleyse:
Looks good - and now gif/ktx can actually go through the pipeline safely without failing (unless using AO).

Issue #199, opened 2016-12-19
GLTF from .dae file


shiva416:
Hi,

I need to convert .dae to .glb files which are more than 10MB files. Inorder to do the conversion I was following below steps,
Step 1: Use Collada2GLTF utility to convert .dae to .gltf files. This works fine.
Step 2: Use GLTF-Pipeline utility thru Node.js to converted exported .gltf files to .glb.

When I perform Step 2 with attached files I end up having below exception.
Appreciate any help.

Command:
node ./bin/gltf-pipeline.js -i ./input/WheelsandDriveUnit.gltf -o WheelsandDriveUnit.glb

Error:
Unhandled rejection TypeError: "value" argument is out of bounds
    at checkInt (buffer.js:1026:11)
    at Buffer.writeUInt16LE (buffer.js:1084:5)
    at writeBufferComponent (C:\gltf-pipeline-master\gltf-pipeline-master\lib\wr
iteBufferComponent.js:24:27)
    at createAccessor (C:\gltf-pipeline-master\gltf-pipeline-master\lib\createAc
cessor.js:49:13)
    at createPrimitive (C:\gltf-pipeline-master\gltf-pipeline-master\lib\combine
Primitives.js:247:29)
    at combinePrimitivesInGroup (C:\gltf-pipeline-master\gltf-pipeline-master\li
b\combinePrimitives.js:222:33)
    at combinePrimitives (C:\gltf-pipeline-master\gltf-pipeline-master\lib\combi
nePrimitives.js:41:42)
    at Function.Pipeline.processJSONWithExtras (C:\gltf-pipeline-master\gltf-pip
eline-master\lib\Pipeline.js:97:5)
    at C:\gltf-pipeline-master\gltf-pipeline-master\lib\Pipeline.js:207:29
    at tryCatcher (C:\gltf-pipeline-master\gltf-pipeline-master\node_modules\blu
ebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (C:\gltf-pipeline-master\gltf-pipeline-
master\node_modules\bluebird\js\release\promise.js:510:31)
    at Promise._settlePromise (C:\gltf-pipeline-master\gltf-pipeline-master\node
_modules\bluebird\js\release\promise.js:567:18)
    at Promise._settlePromise0 (C:\gltf-pipeline-master\gltf-pipeline-master\nod
e_modules\bluebird\js\release\promise.js:612:10)
[...]

PR #200, opened 2016-12-20
Added preserve flag to pipeline

lasalvavida:
@lilleyse, this adds the `-p` flag to preserve the original glTF hierarchy. Optimization stages are not run with this flag.


pjcozzi:
Thanks @lasalvavida.  Will leave #199 open until uint32 is supported properly with glTF 1.1.

Please updated CHANGES.md in master.

PR #203, opened 2017-01-02
Expose functionality to generate tangents and binormals

mlimper:
Adding command line flag and code to generate tangents and binormals for a glTF asset (for example for usage with PBR pipelines). To perform the actual computation,  Cesiums `GeometryPipeline.computeBinormalAndTangent` is used.

Usage:
```
node ./bin/gltf-pipeline.js -i input.gltf --tangentsBinormals -o output.gltf
```


lib/Pipeline.js
```diff
@@ -105,7 +107,12 @@ Pipeline.processJSONWithExtras  = function(gltfWithExtras, options) {
         optimizeForVertexCache(gltfWithExtras);
     }
 
-    // run AO after optimizeForVertexCache since AO adds new attributes.
+    // run generation of tangents / binormals and AO after
+    // optimizeForVertexCache since those steps add new attributes.
+    if (defined(options.tangentsBinormals) && options.tangentsBinormals) {
```
pjcozzi: Why not just
```javascript
if (defined(options.tangentsBinormals)) {
```

lib/Pipeline.js
```diff
@@ -30,6 +31,7 @@ var writeGltf = require('./writeGltf');
 var writeBinaryGltf = require('./writeBinaryGltf');
 var writeSource = require('./writeSource');
 
+
```
pjcozzi: Minor, but extra whitespace.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -1,4 +1,4 @@
-'use strict';
+    'use strict';
```
pjcozzi: Throughout, please preserve the existing formatting.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -18,20 +18,32 @@ function getFirstAttributeSemantic(gltf, primitive, semantic, packedLength) {
     var type;
     if (semantic === 'TEXCOORD') {
         type = 'VEC2';
-    } else if (semantic === 'POSITION' || semantic === 'NORMAL') {
+    } else if (semantic.indexOf('POSITION') === 0 ||
+               semantic.indexOf('NORMAL')   === 0 ||
+               semantic.indexOf('TANGENT')  === 0 ||
```
pjcozzi: The glTF spec does not define `NORMAL` and `TANGENT` attributes.  See

https://github.com/KhronosGroup/glTF/tree/master/specification/1.0#parametersemantic

Perhaps we should add them for 1.1 since it is trivial?

Otherwise, the PBR extension could define them, but they would need to be prefixed with `_` here so the glTF asset is still valid before the extension is added.

pjcozzi: Let's discuss in https://github.com/KhronosGroup/glTF/issues/812.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -18,20 +18,32 @@ function getFirstAttributeSemantic(gltf, primitive, semantic, packedLength) {
...
+               semantic.indexOf('BINORMAL') === 0   ) {
         type = 'VEC3';
     } else {
         throw new DeveloperError('Unsupported attribute semantic: ' + semantic);
     }
     if (semantics.length <= 0) {
-        primitive.attributes[semantic] = createAccessor(gltf, packedLength, type, WebGLConstants.FLOAT, WebGLConstants.ARRAY_BUFFER);
+
+        var id = undefined;
```
pjcozzi: I don't think this will pass JSHint.

Just `var id;`.

lib/cesiumGeometryToGltfPrimitive.js
```diff
@@ -18,20 +18,32 @@ function getFirstAttributeSemantic(gltf, primitive, semantic, packedLength) {
...
+        
+        // all attributes but texcoords may occur only once
+        if (semantic != 'TEXCOORD') {
```
pjcozzi: This won't pass JSHint.

Use `!==` for exact comparison without type coercion.

lib/generateTangentsBinormals.js
```diff
@@ -0,0 +1,73 @@
...
+        												   !defined(tangentAccessorId)    ) {
+            var geometry = gltfPrimitiveToCesiumGeometry(gltf, primitive);
+            GeometryPipeline.computeBinormalAndTangent(geometry);
+            cesiumGeometryToGltfPrimitive(gltf, primitive, geometry);
+        }
+    }
+    return gltf;
+}
+
+//TODO: copied from generateNormals, could be shared
```
pjcozzi: Could you please move this to a separate file (mark it `@private` so it is not part of the public API) add a unit test for it?

pjcozzi:
Thanks @mlimper, this is a nice contribution!  More PBR contributions are very welcome.

lib/Pipeline.js
mlimper: we have a similar function that just checks if the given variable exists (is not `undefined`), guess that's why I got a bit confused here

lib/generateTangentsBinormals.js
mlimper: After a second look, I just saw that there's also a `getPrimitives` function in `PrimitiveHelpers.js`. I used that one now in `getNormals`and `getTangentsBinormals`:
https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/203/commits/ffa215c4f726f8a02edea5275025179fc537335e

The difference is, however, that this function accesses the primitives directly from the meshes, not via the nodes. I believe this makes more sense in this case - otherwise, when going over the nodes, we would potentially iterate over a single mesh multiple times, if it is being shared by several nodes?

I also moved the old function from `generateNormals`, which iterates over the nodes, into the `PrimitiveHelpers` file: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/203/commits/ef141f0a4063d402021819ef4f3cfd93eb812484
However, right now, this function is not being used anywhere.

Does that make sense, should it be kept like this?

README.md
```diff
@@ -41,6 +41,7 @@ node ./bin/gltf-pipeline.js -i ./specs/data/boxTexturedUnoptimized/CesiumTexture
 |`--compressTextureCoordinates`, `-c`|Compress the testure coordinates of this glTF asset.|No, default `false`|
 |`--removeNormals`, `-r`|Strips off existing normals, allowing them to be regenerated.|No, default `false`|
 |`--faceNormals`, `-f`|If normals are missing, they should be generated using the face normal.|No, default `false`|
+|`--tangentsBinormals`|If normals and texture coordinates are given, generate tangents and binormals.|No, default `false`|
```
pjcozzi: Could you also update CHANGES.md please?

Sorry, I forgot to mention last time.

lib/PrimitiveHelpers.js
```diff
@@ -185,6 +185,31 @@ function getPrimitivesByMaterialMode(primitives) {
     return primitivesByMaterialMode;
 }
 
+function getPrimitivesFromNodes(gltf) {
```
pjcozzi: If this isn't used anywhere, I would remove it for now.

pjcozzi:
Can you please add new tests to https://github.com/AnalyticalGraphicsInc/gltf-pipeline/tree/master/specs/lib?

pjcozzi:
Just those comments and this is ready!  Thanks again for contributing and moving forward PBR so quickly!

pjcozzi:
@mlimper also note that binormal should be renamed to bitangent, https://github.com/KhronosGroup/glTF/issues/812#issuecomment-271037920

mlimper:
Just removed the unused function, refactored to "bitangent" (instead of "binormal") throughout*, and added a test case.

(*)Exception: As Cesium is still using "binormal", the respective parts are still using the old term:
https://github.com/mlimper/gltf-pipeline/commit/2985da34f32187c8f1c34d8cd42812fc8b1681b4

I got a conflict in `CHANGES.md`, so I tried to update my fork by rebasing... unfortunately, that somehow produced some [duplicate commits](https://github.com/mlimper/gltf-pipeline/commits/master) right now, which is a bit annoying :-/ I will see if I can fix that somehow....

lib/Pipeline.js
```diff
@@ -88,7 +89,7 @@ Pipeline.processJSONWithExtras  = function(gltfWithExtras, options) {
     }
     addDefaults(gltfWithExtras, options);
     RemoveUnusedProperties.removeAll(gltfWithExtras);
-    generateNormals(gltfWithExtras, options);
+    generateNormals(gltfWithExtras, options);
```
pjcozzi: In future PRs, please be mindful of this extra whitespace added to lines like this.

pjcozzi:
Great, thanks again @mlimper!

lib/Pipeline.js
mlimper: Oh yes, sure - thanks for the pointer, and thanks for the helpful code reviews!

PR #212, opened 2017-01-13
String shaders and browser-friendly parseBinaryGltf

lasalvavida:
@lilleyse, could you take a look at this when you get a chance?

These changes make `parseBinaryGltf` work with Cesium in the browser.

It also changes `extras._pipeline.source` for shaders to be a string. This removes unnecessary conversion back and forth string <-> buffer, and makes more sense in the browser.


lilleyse:
In `writeSource` it still expects a Buffer, so base64 encoding isn't working.

`object.uri = 'data:text/plain;base64,' + source.toString('base64');`

lilleyse:
Otherwise, the code looks good.

lasalvavida:
> In writeSource it still expects a Buffer, so base64 encoding isn't working.

Good catch

lasalvavida:
Updated

PR #213, opened 2017-01-13
Add pipeline extras cleanup

lasalvavida:

This reduces `addPipelineExtras` to not be quite so verbose. It also removes the `deleteExtras` qualification from removing pipeline extras, since there is currently no situation I am aware of where we want to export a gltf with pipeline extras.

### DIFF
--- a/lib/addPipelineExtras.js
+++ b/lib/addPipelineExtras.js
@@ -1,116 +1,65 @@
 'use strict';
 var Cesium = require('cesium');
 
+var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
 
 module.exports = addPipelineExtras;
 
+// Objects with these ids should not have extras added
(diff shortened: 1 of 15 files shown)


lilleyse:
Searching for `deleteExtras` there are still some occurrences left around.

lilleyse:
Very nice, much cleaner. This also fixes an annoyance I was having in a the 3d tiles generator where some extras where sticking around in the extensions.

lasalvavida:
> Searching for deleteExtras there are still some occurrences left around.

Ah yes, I did miss a few.

lasalvavida:
Updated. I think I got them all this time. You may see some in your local doc folder if you have that generated, but built doc isn't in the repo.

lilleyse:
Fixed a minor merge conflict just in the interest of time, looks good.

PR #215, opened 2017-01-13
Make findAccessorMinMax browser friendly

lasalvavida:

This makes `findAccessorMinMax` use `ComponentDatatype` from Cesium so that it works in the browser.

### DIFF
--- a/lib/findAccessorMinMax.js
+++ b/lib/findAccessorMinMax.js
@@ -1,12 +1,11 @@
 'use strict';
 var Cesium = require('cesium');
 
+var ComponentDatatype = Cesium.ComponentDatatype;
 var defined = Cesium.defined;
 
-var byteLengthForComponentType = require('./byteLengthForComponentType');
 var getAccessorByteStride = require('./getAccessorByteStride');
 var numberOfComponentsForType = require('./numberOfComponentsForType');
-var readBufferComponent = require('./readBufferComponent');
 
 module.exports = findAccessorMinMax;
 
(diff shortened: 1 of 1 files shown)


lib/findAccessorMinMax.js
```diff
@@ -35,15 +34,16 @@ function findAccessorMinMax(gltf, accessor) {
...
             var source = buffer.extras._pipeline.source;
+
             var count = accessor.count;
             var byteStride = getAccessorByteStride(accessor);
             var byteOffset = accessor.byteOffset + bufferView.byteOffset;
             var componentType = accessor.componentType;
-            var componentByteLength = byteLengthForComponentType(componentType);
 
             for (var i = 0; i < count; i++) {
+                var typedArray = ComponentDatatype.createArrayBufferView(componentType, source.buffer, byteOffset, numberOfComponents);
```
lilleyse: This should go outside the loop.

lilleyse: Well may require some logic changes, but I think this would be pretty inefficient.

lasalvavida: > This should go outside the loop.

This cannot go outside the loop. We can't guarantee that an accessor is contiguous in the buffer, and we also can't assume that byteStride is a multiple of our component size.

> I think this would be pretty inefficient.

Why? We're just making a view of the existing data, not copying anything.

lilleyse: Ah good points, I forgot about interleaved accessors. I know it's just a view, but it just seems like a heavy operation to be doing on a per element basis. But if there's now way to avoid it then that's fine.

PR #216, opened 2017-01-13
Added script for generating Cesium AMD dependencies

lasalvavida:
@lilleyse or @pjcozzi, this PR adds the script for converting pipeline modules into AMD format for cesium

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ typings
 
 # Generate data
 coverage
+dist
 output
 test
 specs/data/boxTexturedUnoptimized/output
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,5 @@
 /.idea
+/dist
 /coverage
 /doc
 /output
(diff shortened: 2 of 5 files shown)


lilleyse:
Add `dist` to .gitignore and .npmignore

lilleyse:
Also exclude `dist` from `jsHintFiles`, `specFiles`, and the `coverage` task.

pjcozzi:
Thanks @lasalvavida, this is a perfectly reasonable stopgap until we have bandwidth to do proper modules.

For my understanding, there is no browserify or related translation happening here, right?  Is this because these files aren't using any Node API that would need to be transpiled?

lasalvavida:
> For my understanding, there is no browserify or related translation happening here, right?

Nope, no transpiling currently, so anything node-specific will not work. Currently only browser-friendly pipeline code will work with this which I think is probably acceptable for now.

lilleyse:
One small suggestion, add newlines to the ends of newly created dist files.

The code looks solid to me. Once this is running in Cesium successfully this should be good to merge.

pjcozzi:
All sounds good.  @lilleyse please merge when ready.

lasalvavida:
@lilleyse, updated.

pjcozzi:
LGTM

PR #219, opened 2017-01-18
Don't try to add extras to null objects.

kring:
Surprisingly, `typeof null === 'object'`.  I modified `CesiumTexturedBoxTestAddExtras.gltf` to trigger the bug fixed by this PR.  I realize it's kinda dodgy glTF, but I ran into it in a real model.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 Change Log
 ==========
 
+### 0.1.0-alpha11 - ??
+* Fixed a bug in `addPipelineExtras` that made it try to add extras to null objects.
+
 ### 0.1.0-alpha10 - 2017-01-10
 * Added `tangentsBitangents` generation option
 
--- a/lib/addPipelineExtras.js
+++ b/lib/addPipelineExtras.js
@@ -44,7 +44,7 @@ function addPipelineExtras(gltf) {
             var length = object.length;
             for (var i = 0; i < length; i++) {
                 var item = object[i];
-                if (typeof item === 'object') {
+                if (defined(item) && typeof item === 'object') {
                     objectStack.push(item);
                 }
             }
--- a/specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTestAddExtras.gltf
+++ b/specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTestAddExtras.gltf
@@ -6,23 +6,6 @@
             "componentType": 5123,
             "count": 36,
             "type": "SCALAR"
-        },
-        "accessor_25": {
-            "bufferView": "bufferView_29",
-            "byteOffset": 0,
-            "componentType": 5123,
-            "count": 36,
-            "type": "SCALAR",
-            "min": [
-                null,
-                null,
-                null
-            ],
-            "max": [
-                null,
-                null,
-                null
-            ]
         }
     },
     "animations": {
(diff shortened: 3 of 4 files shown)


pjcozzi:
Does the model [pass validation](https://github.com/KhronosGroup/glTF-Validator)?  (Note that is written against glTF 2.0, not 1.0, so it might not pass for other reasons).

If it doesn't pass, we don't want to put workarounds in gltf-pipeline unless it is to workaround a super popular exporter.  It will hurt the overall glTF ecosystem.

If this change is still needed, the only tweak I suggest is:

> I modified CesiumTexturedBoxTestAddExtras.gltf to trigger the bug fixed by this PR...

Instead of changing the file, add a new test that adds the object with `null` properties.

kring:
It was created by `obj2gltf`.  I don't know if there's something weird about my OBJ that caused it.

I think the change is good one even the situation is unlikely.  It's pretty clear the author of the code here just didn't realize that `typeof null === 'object'` (cause that's totally weird).  I'll put it in a separate .gltf, but it's worth mentioning that that gltf is already invalid.  It's missing accessors, for one thing.

pjcozzi:
Can you submit a bug or pull request to obj2gltf?

I still need to see a case where this matters in practice on valid glTF; otherwise, I'm not sure that this is a good change, it is the kind of "fix" that creates unreliable ecosystems where workarounds are required to conform to the unwritten spec.

kring:
> it is the kind of "fix" that creates unreliable ecosystems where workarounds are required to conform to the unwritten spec.

Yeah, maybe.  Though notice that in that function there are three places where a new object is pushed onto `objectStack`.  In two of the three, there's a `defined` check in addition to the `typeof x === 'object'` check.  My PR adds it to the third.  It's not clear to me why the extra check is justified in two cases but not the third.  Also, this model loaded fine prior to https://github.com/AnalyticalGraphicsInc/gltf-pipeline/commit/68d3849b143f3687b7ee53cc05e2b0f7ab2361ef.  Was this "tightening" really intended in a cleanup PR?

Either way, I agree it's worth a minute to figure out why obj2gltf is generating this and I'll do so when I get a chance.

kring:
Let me state it another way.  As currently written, before this PR, a null as a value of property is fine, whether the spec allows nulls there or not.  The model will load just fine.  But if there's a null inside an array, attempting to load the model will crash with a null reference exception.  You're welcome to close this PR if you disagree, but to me that behavior is pretty hard to defend.

kring:
I found the source of the NaNs, in case it matters.  My OBJ has a few normals that are the zero vector.  Maybe they're not used so they don't matter, maybe it's a problem with the original model, I'm not sure.  In any case, obj2gltf normalizes the normals during the conversion process (why? I don't see anything in the glTF spec that requires that normals be normalized), which produces NaN.  That causes the min and max range of the normal attribute to be NaN, and `JSON.stringify` serializes NaN as null.

lilleyse:
This change seems fine to me and just seems like an oversight as all the other areas that check `typeof item === 'object'` in `addPipelineExtras` and `removePipelineExtras` first check that its `defined`.

Also thanks for tracking down the root cause in obj2gltf.

pjcozzi:
Sorry folks, have had bigger fish to fry.  If you guys think this is inline with the spec, then go for it.  If the spec needs work, then please submit an issue to https://github.com/KhronosGroup/glTF/.

As for requiring normals to be normalized, see https://github.com/KhronosGroup/glTF/issues/812.  In practice, most people would want to do this to save runtime processing.

PR #220, opened 2017-01-18
glTF Statistics

shehzan10:
* Merges the [gltf-statistics](https://github.com/AnalyticalGraphicsInc/gltf-statistics) repository.
* Currently, the `--stats` option is used to print the statistics of the input and output onto the console.
* This can be easily expanded to run on any stage of the pipeline by calling the function.
* Another extension, which is a part of gltf-statistics is to write to CSV files. Again, this is very easy to expand on in the future when gltf-pipeline will run for directories.


lib/Pipeline.js
```diff
@@ -88,8 +89,14 @@ Pipeline.processJSONWithExtras  = function(gltfWithExtras, options) {
         removeNormals(gltfWithExtras);
     }
     addDefaults(gltfWithExtras, options);
+
+    // Generate stats for unoptimized input
+    if (options.stats) {
+        gltfStatistics.printStats(gltfWithExtras, options.inputPath);
+    }
```
lilleyse: This should go first, before `removeNormals`.

lib/Pipeline.js
```diff
@@ -149,6 +156,12 @@ Pipeline.processJSONWithExtras  = function(gltfWithExtras, options) {
             if (defined(kmcOptions) && kmcOptions.enable) {
                 generateModelMaterialsCommon(gltfWithExtras, options.kmcOptions);
             }
+
+            // Generate stats for unoptimized input
```
lilleyse: unoptimized -> optimized

lib/Pipeline.js
```diff
@@ -88,8 +89,14 @@ Pipeline.processJSONWithExtras  = function(gltfWithExtras, options) {
...
     addDefaults(gltfWithExtras, options);
+
+    // Generate stats for unoptimized input
+    if (options.stats) {
+        gltfStatistics.printStats(gltfWithExtras, options.inputPath);
+    }
+
     RemoveUnusedProperties.removeAll(gltfWithExtras);
-    generateNormals(gltfWithExtras, options);    
+    generateNormals(gltfWithExtras, options);
```
lilleyse: Thanks for fixing these spacing issues.

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+
+module.exports = {
+    gltfStatistics,
+    getDrawCallStatistics,
+    printStats
+}
+
+function getBuffersSize(buffers) {
+    var size = 0;
+    for (var name in buffers) {
```
lilleyse: Instead of `name`, use `id` which is the proper glTF terminology. Same comment for the other loops.

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+    };
+}
+
+function printStats(gltf, gltfPath) {
+    var stats = gltfStatistics(gltf);
+    var output =
+        '\n' +
+        'Key statistics for ' + gltfPath + '\n' +
+        '--------------\n' +
+        'Total size of all buffers: ' + stats.buffersSizeInBytes.toLocaleString() + ' bytes\n' +
```
lilleyse: I'm not sure if there was an original reason for using `toLocaleString` but even just

`Total size of all buffers: ' + stats.buffersSizeInBytes + ' bytes\n'` should be fine

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+        '\n' +
+        'Draw calls: ' + stats.numberOfDrawCalls.toLocaleString() + '\n' +
+        'Rendered primitives (e.g., triangles): ' + stats.numberOfRenderedPrimitives.toLocaleString() + '\n' +
+        '\n' +
+        'Nodes: ' + stats.numberOfNodes.toLocaleString() + '\n' +
+        'Meshes: ' + stats.numberOfMeshes.toLocaleString() + '\n' +
+        'Materials: ' + stats.numberOfMaterials.toLocaleString() + '\n' +
+        'Animations: ' + stats.numberOfAnimations.toLocaleString() + '\n';
+
+    process.stdout.write(output);
```
lilleyse: Using `console.log` is fine and more consistent with the rest of the code.

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+    return;
```
lilleyse: Unneeded `return;`

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+        return new Error('gltf must be defined');
+    }
+
+    var drawStats = getAllDrawCallStats(gltf);
+
+    return {
+        buffersSizeInBytes : getBuffersSize(gltf.buffers),
+        numberOfImages : getNumberOfProperties(gltf.images),
+        numberOfExternalRequests : getNumberOfExternalRequests(gltf),
+
```
lilleyse: Remove empty line.

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+var Cesium = require('cesium');
+
+var defaultValue = Cesium.defaultValue;
+var defined = Cesium.defined;
+
+module.exports = {
+    gltfStatistics,
+    getDrawCallStatistics,
+    printStats
+}
```
lilleyse: This should instead look like:

```javascript
module.exports = {
    gltfStatistics : gltfStatistics,
    getDrawCallStatistics : getDrawCallStatistics,
    printStats : printStats
};
```

lilleyse: I don't think `getDrawCallStatistics` is necessary to expose.

lilleyse: Rename `gltfStatistics` to `getStatistics` and `printStats` to `printStatistics`.

lilleyse: Make sure to add jsdoc above each of the exposed functions.

README.md
```diff
@@ -42,6 +42,7 @@ node ./bin/gltf-pipeline.js -i ./specs/data/boxTexturedUnoptimized/CesiumTexture
 |`--removeNormals`, `-r`|Strips off existing normals, allowing them to be regenerated.|No, default `false`|
 |`--faceNormals`, `-f`|If normals are missing, they should be generated using the face normal.|No, default `false`|
 |`--tangentsBitangents`|If normals and texture coordinates are given, generate tangents and bitangents.|No, default `false`|
+|`--stats`|Generate statistics for input and output glTF files.|No, default `false`|
```
lilleyse: Mention here and in `parseArguments` that this prints the statistics to the console.

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+    var size = 0;
+    for (var name in buffers) {
+        if (buffers.hasOwnProperty(name)) {
+            size += buffers[name].byteLength;
+        }
+    }
+    return size;
+}
+
+var isDataUriRegex = /^data:/;
```
lilleyse: Use the `isDataUri` helper file instead.

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+
+    var shaders = gltf.shaders;
+    for (name in shaders) {
+        if (shaders.hasOwnProperty(name)) {
+            uri = shaders[name].uri;
+            if (defined(uri) && !isDataUriRegex.test(uri)) {
+                ++count;
+            }
+        }
+    }
```
lilleyse: These can be condensed by factoring out the common code into a helper function.

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+function getNumberOfProperties(object) {
+    var count = 0;
+    for (var name in object) {
+        if (object.hasOwnProperty(name)) {
+            ++count;
+        }
+    }
+
+    return count;
+}
```
lilleyse: Easier off to just use `Object.keys(object).length` rather than this function.

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+        '\n' +
+        'Nodes: ' + stats.numberOfNodes.toLocaleString() + '\n' +
+        'Meshes: ' + stats.numberOfMeshes.toLocaleString() + '\n' +
+        'Materials: ' + stats.numberOfMaterials.toLocaleString() + '\n' +
+        'Animations: ' + stats.numberOfAnimations.toLocaleString() + '\n';
+
+    process.stdout.write(output);
+    return;
+}
+
```
lilleyse: Remove empty line (but keep newline).

lib/gltfStatistics.js
```diff
@@ -0,0 +1,202 @@
...
+    var numberOfRenderedPrimitives = 0;
+
+    var rootNodes = gltf.scenes[gltf.scene].nodes;
+    var rootNodesLength = rootNodes.length;
+
+    for (var i = 0; i < rootNodesLength; ++i) {
+        var stats = getDrawCallStatistics(gltf, rootNodes[i]);
+        numberOfDrawCalls += stats.numberOfDrawCalls;
+        numberOfRenderedPrimitives += stats.numberOfRenderedPrimitives;
+    }
```
lilleyse: I would be fine with removing `getDrawCallStatistics` and just placing that logic in here. Then the code can be heavily condensed heavily by using `PrimitiveHelpers.getAllPrimitives`. Then `getAllDrawCallStats` can take the name `getDrawCallStatistics`.

lib/parseArguments.js
```diff
@@ -90,6 +90,10 @@ function parseArguments(args) {
                 describe: 'Optimize the glTF for Cesium by using the sun as a default light source.',
                 type: 'boolean'
             },
+            'stats': {
+                describe: 'Generate statistics for input and output glTF files.',
+                type: 'boolean'
```
lilleyse: Add a `default`.

lib/Pipeline.js
shehzan10: The reason I put this after `removeNormals` is because the gltfStatistics function requires the defaults to be set (`addDefaults`). Would it be ok to move the `addDefaults` above `removeNormals` too?

lilleyse: As talked about offline, `gltfStatistics` should be able to work on any gltf given to it, so this function should check that `materials` or other things are `undefined` and act accordingly.

shehzan10:
Changes completed

specs/lib/gltfStatisticsSpec.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
```
lilleyse: Use single quotes.

specs/lib/gltfStatisticsSpec.js
```diff
@@ -0,0 +1,48 @@
...
+var assert = require("assert");
```
lilleyse: Use jasmine `expect` functions, not `assert`

specs/lib/gltfStatisticsSpec.js
```diff
@@ -0,0 +1,48 @@
...
+
+                assert.equal(stats.numberOfDrawCalls, 3);
+                assert.equal(stats.numberOfRenderedPrimitives, 5840);
+
+                assert.equal(stats.numberOfNodes, 7);
+                assert.equal(stats.numberOfMeshes, 2);
+                assert.equal(stats.numberOfMaterials, 2);
+                assert.equal(stats.numberOfAnimations, 2);
+            });
+    });
```
lilleyse: I left a similar comment in the `3d-tiles-tools` pr - make sure to check that the promise resolves, like in `writeImagesSpec`.

```javascript
    it('writes an embedded buffer', function(done) {
        var gltf = clone(testGltf);
        var options = {
            outputPath : outputPath,
            embed : true,
            embedImage : true,
            createDirectory : true
        };
        
        expect(writeGltf(gltf, options)
            .then(function() {
                expect(gltf.images.Cesium_Logo_Flat_Low.extras).not.toBeDefined();
                expect(gltf.images.Cesium_Logo_Flat_Low.uri).toEqual(imageUri);
            }), done).toResolve();
    });
```

lib/Pipeline.js
```diff
@@ -82,14 +83,22 @@ Pipeline.processJSON = function(gltf, options) {
  */
 Pipeline.processJSONWithExtras  = function(gltfWithExtras, options) {
     options = defaultValue(options, {});
+
+    addDefaults(gltfWithExtras, options);
+
+    // Generate stats for unoptimized input
+    if (options.stats) {
+        gltfStatistics.printStatistics(gltfWithExtras, options.inputPath);
+    }
```
lilleyse: I looked at this more carefully and noticed that `addDefaults` doesn't add any items that are used by `gltfStatistics`. It does add techniques, shaders, programs and other things when the model materials common extension is used. The order here is ok for now though.

shehzan10: Right, and per our offline discussion, gltfStatistics no longer depends on `addDefaults`. So it would be ok if we decide to move `addDefaults` back to its original location.

lilleyse: `gltfStatistics` would need some tweaks in order to work if some sections of the glTF are missing. Like if `gltf.images` is undefined `getNumberOfProperties` will throw an error.

index.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     getAccessorByteStride : require('./lib/getAccessorByteStride'),
     getBinaryGltf : require('./lib/getBinaryGltf'),
     getUniqueId : require('./lib/getUniqueId'),
+    gltfStatistics : require('./lib/gltfStatistics'),
```
lasalvavida: I think this should probably be named something else. For example when `gltfDefaults` was added to gltf-pipeline, it became `addDefaults`. Maybe something like `computeStatistics`.

lib/parseBinaryGltf.js
```diff
@@ -47,10 +47,11 @@ function parseBinaryGltf(data) {
     var gltf = JSON.parse(scene);
     addPipelineExtras(gltf);
 
-    // Extract binary body
-    var body = data.slice(20 + sceneLength);
+    //Extract binary body
+    byteOffset += sceneLength;
```
lasalvavida: Why did you make this change? Also byteOffset is not declared so this is probably throwing jsHint errors.

lasalvavida:
I see that you added CesiumMan and CesiumAir with this for tests. Can you use some of the existing test models instead? Adding models tends to bloat the repo, so we try to avoid if possible.

lasalvavida:
Also, would it be possible for you to put your whitespace deletion and file permission changes in a separate pull request? It will get merged faster and it would make this one a lot easier to read.

shehzan10:
> Also, would it be possible for you to put your whitespace deletion and file permission changes in a separate pull request? It will get merged faster and it would make this one a lot easier to read.

@lasalvavida I've moved this into #224. I will clean this PR's commits and also have the suggested changes.

shehzan10:
> I see that you added CesiumMan and CesiumAir with this for tests. Can you use some of the existing test models instead? Adding models tends to bloat the repo, so we try to avoid if possible.

These were the tests from [gltf-statistics](https://github.com/AnalyticalGraphicsInc/gltf-statistics) for which there were reference values, hence I copied the files over. I agree it increases the size of the repo. Is there any way to get the statistics of gltf files without actually using this statistics feature? It wouldn't make sense to generate statistics with this and then run the specs for the same numbers.

lasalvavida:
> Is there any way to get the statistics of gltf files without actually using this statistics feature?

Just open them up in a text editor:
* buffersSizeInBytes - add up byteLength for all buffers
* numberOfImages - count them
* numberOfExternalRequests - how many buffers, shaders, images use external URIs?
* numberOfDrawCalls - should be the same as the number of triangles (# of indices / 3)
* numberOfRenderedPrimitives - count them
* numberOfNodes - count
* numberOfMeshes - count
* numberOfMaterials - count
* numberOfAnimations - count

Should be pretty straightforward; if you use `specs/data/riggedSimpleUnoptimized` and `specs/data/boxTexturedUnoptimized`, you should have full coverage for images/animations.

shehzan10:
Cleaned and updated

pjcozzi:
@shehzan10 can you please open a pull request into gltf-statistics to update the README.md to say that project is no longer maintained and to use gltf-pipeline instead (include an example of running it from the command-line)?

There might still be some users of gltf-statistics so I don't want to fully remove the repo or the files yet.

lib/computeStatistics.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    return {
+        numberOfDrawCalls : numberOfDrawCalls,
+        numberOfRenderedPrimitives : numberOfRenderedPrimitives
+    };
+}
+
+/**
+ * @typedef {Object} gltfStatistic
```
pjcozzi: It's not a huge deal here and I was certainty moving fast and loose in gltf-statistics, but we should replace this with an explicit type instead of the object literal.

pjcozzi:
Just those comments; this looks great.  Please verify the reference doc build.

@lasalvavida or @lilleyse can merge when ready.

index.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     getAccessorByteStride : require('./lib/getAccessorByteStride'),
     getBinaryGltf : require('./lib/getBinaryGltf'),
     getUniqueId : require('./lib/getUniqueId'),
+    computeStatistics : require('./lib/computeStatistics'),
```
lasalvavida: This should be in alphabetical order

lib/Pipeline.js
```diff
@@ -18,6 +18,7 @@ var encodeImages = require('./encodeImages');
 var generateModelMaterialsCommon = require('./generateModelMaterialsCommon');
 var generateNormals = require('./generateNormals');
 var generateTangentsBitangents = require('./generateTangentsBitangents');
+var computeStatistics = require('./computeStatistics');
```
lasalvavida: This should be alphabetical as well

lib/computeStatistics.js
shehzan10: Just to clarify, should I change this to be like a class? This would then look like:
```
function ComputeStatistics(...) {
    // properties here
}

ComputeStatistics.getStatistics() {
    // Code
    return ComputeStatistics(...);
}

ComputeStatistics.printStatistics() {
    // Code
}
```

lilleyse: I believe just make the statistics themselves a class like:

```
function Statistcs() {
    this.buffersSizeInBytes = undefined;
    this.numberOfImages = undefined;
    etc...
}
```

shehzan10:
@lilleyse Completed the object type and updated documentation.

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * @param {Object} gltf A javascript object containing a glTF asset with extras and defaults.
+ * @returns {Statistics} Object containing the properties of the assest.
+ *
+ * @see Statistics
+ * @see printStatistics
+ */
+function getStatistics(gltf) {
+    if (!defined(gltf)) {
+        return new Error('gltf must be defined');
```
lilleyse: throw the error instead.

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+        '\nDraw calls: ' + statistics.numberOfDrawCalls +
+        '\nRendered primitives (e.g., triangles): ' + statistics.numberOfRenderedPrimitives +
+        '\n' +
+        '\nNodes: ' + statistics.numberOfNodes +
+        '\nMeshes: ' + statistics.numberOfMeshes +
+        '\nMaterials: ' + statistics.numberOfMaterials +
+        '\nAnimations: ' + statistics.numberOfAnimations;
+
+    console.log(output);
+}
```
lilleyse: I'm starting to think... now that the `Statistics` class exists, this could be a function of the class and `computeStatistics` could be renamed `getStatistics` and just expose the `getStatistics` function. `printStatistics` could instead become `toString` and the caller would be responsible in calling `console.log`. Does this seem like a good idea?

specs/lib/computeStatisticsSpec.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+var fs = require('fs');
+var addDefaults = require('../../lib/addDefaults');
+var addPipelineExtras = require('../../lib/addPipelineExtras');
+var getStatistics = require('../../lib/computeStatistics').getStatistics;
+var readGltf = require('../../lib/readGltf');
+
+describe('computeStatistics', function() {
+
```
lilleyse: Remove whitespace.

specs/lib/computeStatisticsSpec.js
```diff
@@ -0,0 +1,47 @@
...
+                    expect(stats.numberOfDrawCalls).toEqual(1);
+                    expect(stats.numberOfRenderedPrimitives).toEqual(0);
+
+                    expect(stats.numberOfNodes).toEqual(5);
+                    expect(stats.numberOfMeshes).toEqual(1);
+                    expect(stats.numberOfMaterials).toEqual(1);
+                    expect(stats.numberOfAnimations).toEqual(2);
+                }), done).toResolve();
+    });
+});
```
lilleyse: Too much indentation in some areas of this file.

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+function Statistics() {
+    this.bufferSizeInBytes = undefined;
+    this.numberOfImages = undefined;
+    this.numberOfExternalRequests = undefined;
+    this.numberOfDrawCalls = undefined;
+    this.numberOfRenderedPrimitives = undefined;
+    this.numberOfNodes = undefined;
+    this.numberOfMeshes = undefined;
+    this.numberOfMaterials = undefined;
+    this.numberOfAnimations = undefined;
```
lilleyse: My fault, these should be set to `0` to avoid jsHint warnings.

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+
+var defined = Cesium.defined;
+
+module.exports = {
+    getStatistics : getStatistics,
+    printStatistics : printStatistics
+};
+
+/**
+ * Creates an object type for glTF Statistics.
```
lilleyse: This description could be better. Maybe `Contains statistics for a glTF asset`.

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * @property {Number} buffersSizeInBytes The total size of all the buffers.
+ * @property {Number} numberOfImages The number of images part of the assest.
```
lilleyse: `The number of images in the asset` is better. Also fix spellings - `assest` appears a few times.

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+ * @property {Number} numberOfExternalRequests The number of external requests to fetch the assest data.
```
lilleyse: Slightly better: `The number of external requests required to fetch asset data`

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+ * @property {Number} numberOfDrawCalls The number of draw calls required in the asset.
```
lilleyse: `The number of draw calls required to render the asset.`

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+ * @property {Number} numberOfRenderedPrimitives The number of rendered primitives.
```
lilleyse: For consistency: `The total number of rendered primitives in the asset`. Also add a note that this typically means the number of triangles.

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+    var statistics = new Statistics();
+    statistics.buffersSizeInBytes          = getBuffersSize(gltf.buffers);
+    statistics.numberOfImages              = getNumberOfProperties(gltf.images);
+    statistics.numberOfExternalRequests    = getNumberOfExternalRequests(gltf);
+    statistics.numberOfDrawCalls           = drawStats.numberOfDrawCalls;
+    statistics.numberOfRenderedPrimitives  = drawStats.numberOfRenderedPrimitives;
+    statistics.numberOfNodes               = getNumberOfProperties(gltf.nodes);
+    statistics.numberOfMeshes              = getNumberOfProperties(gltf.meshes);
+    statistics.numberOfMaterials           = getNumberOfProperties(gltf.materials);
+    statistics.numberOfAnimations          = getNumberOfProperties(gltf.animations);
```
lilleyse: Usually we don't format like this, not that its wrong but just not consistent with the style generally. Use formatting like:

```
statistics.buffersSizeInBytes = getBuffersSize(gltf.buffers);
statistics.numberOfImages = getNumberOfProperties(gltf.images);
```

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+        numberOfDrawCalls : numberOfDrawCalls,
+        numberOfRenderedPrimitives : numberOfRenderedPrimitives
+    };
+}
+
+/**
+ * Returns an object containing the statistics for the glTF assest.
+ *
+ * @param {Object} gltf A javascript object containing a glTF asset with extras and defaults.
+ * @returns {Statistics} Object containing the properties of the assest.
```
lilleyse: `Object containing the statistics of the glTF asset.`

lib/computeStatistics.js
```diff
@@ -0,0 +1,182 @@
...
+ * @property {Number} numberOfMaterials The total number of materials in the assest.
+ * @property {Number} numberOfAnimations The total number of animations in the assest.
+ *
+ * @constructor
+ *
+ * @see getStatistics
+ * @see printStatistics
+ */
+function Statistics() {
+    this.bufferSizeInBytes = undefined;
```
lilleyse: Should be `buffersSizeInBytes`.

lib/Pipeline.js
```diff
@@ -149,6 +158,12 @@ Pipeline.processJSONWithExtras  = function(gltfWithExtras, options) {
             if (defined(kmcOptions) && kmcOptions.enable) {
                 generateModelMaterialsCommon(gltfWithExtras, options.kmcOptions);
             }
+
+            // Generate stats for optimized input
```
lilleyse: For this comment and above change `Generate` to `Print`.

lib/computeStatistics.js
shehzan10: I think it's good to have the functions as a part of the class. The usage would then be:

```
var statistics = getStatistics(gltf);
console.log(statistics.toString());
```

shehzan10:
@lilleyse I think I've addressed all the items.

lib/getStatistics.js
```diff
@@ -0,0 +1,179 @@
...
+var DeveloperError = Cesium.DeveloperError;
+var isDataUri = require('./isDataUri');
+var PrimitiveHelpers = require('./PrimitiveHelpers');
+
+var defined = Cesium.defined;
+
+module.exports = getStatistics;
+
+/**
+ * Creates an object type for glTF Statistics.
```
lilleyse: See https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/220#discussion_r97392709

lib/getStatistics.js
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * @property {String} gltfPath The path to the glTF file.
+ * @property {Number} buffersSizeInBytes The total size of all the buffers.
+ * @property {Number} numberOfImages The number of images in asset.
```
lilleyse: `in the asset`

lib/getStatistics.js
```diff
@@ -0,0 +1,179 @@
...
+ * @property {Number} numberOfExternalRequests The number of external requests required to fetch the asset data.
+ * @property {Number} numberOfDrawCalls The number of draw calls required to render the asset.
+ * @property {Number} numberOfRenderedPrimitives The total number of rendered primitives in the asset.
```
lilleyse: Add `(e.g. triangles)` to the end of the description (mimics the toString function).

lib/getStatistics.js
```diff
@@ -0,0 +1,179 @@
...
+
+    return {
+        numberOfDrawCalls : numberOfDrawCalls,
+        numberOfRenderedPrimitives : numberOfRenderedPrimitives
+    };
+}
+
+/**
+ * Returns an object containing the statistics for the glTF asset.
+ * This can be printed to console using object.toString().
```
lilleyse: Probably fine to remove this line since it should appear in documentation.

lib/getStatistics.js
```diff
@@ -0,0 +1,179 @@
...
+ * @property {Number} numberOfMaterials The total number of materials in the asset.
+ * @property {Number} numberOfAnimations The total number of animations in the asset.
+ *
+ * @constructor
+ *
+ * @see getStatistics
+ * @see printStatistics
+ */
+function Statistics() {
+    this.gltfPath = '';
```
lilleyse: I think we can take this out of Statistics entirely. Before printing the stats `Pipeline` can print `Statistics for <input path>`. Sorry for not catching this in the previous review.

lib/Pipeline.js
```diff
@@ -82,12 +83,23 @@ Pipeline.processJSON = function(gltf, options) {
...
+
+    addDefaults(gltfWithExtras, options);
+
+    // Print statistics for unoptimized input
+    if (options.stats) {
+        if (options.stats) {
+            console.log("Statistics for " + options.inputPath + "\n------------------\n" +
+                        getStatistics(gltfWithExtras).toString());
+        }
+    }
```
lilleyse: `options.stats` appears twice.

Use single quotes.

This would probably look cleaner as two different calls to `console.log`.

shehzan10: Oops. Muscle memory.

Yes, I thought of that, but since the old `printStatistics` was a single call, I kept it single as a style thing.

Will change and force push right away.

specs/lib/getStatisticsSpec.js
```diff
@@ -0,0 +1,46 @@
...
+        expect(readGltf('specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf')
+            .then(function(gltf) {
+                var stats = getStatistics(gltf);
+
+                expect(stats.buffersSizeInBytes).toEqual(840);
+                expect(stats.numberOfImages).toEqual(1);
+                expect(stats.numberOfExternalRequests).toEqual(4);
+
+                expect(stats.numberOfDrawCalls).toEqual(1);
+                expect(stats.numberOfRenderedPrimitives).toEqual(0);
```
lilleyse: This should be a warning sign that something isn't right.

In `getStatistics` `primitive.primitive` should be `primitive.mode`. I believe this is legacy from gltf 0.8.

lilleyse:
The newline spacing could be improved:

newlines

shehzan10:
Is this better? I reordered it a little too.
```
$ node ./bin/gltf-pipeline.js --stats -i ./specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf -o output.gltf

Statistics for specs/data/boxTexturedUnoptimized/CesiumTexturedBoxTest.gltf
------------------
Total size of all buffers: 840 bytes
Images: 1
Draw calls: 1
Rendered primitives (e.g., triangles): 0
Nodes: 4
Meshes: 1
Materials: 1
Animations: 0
External requests (not data uris): 4

Statistics for output.gltf
------------------
Total size of all buffers: 840 bytes
Images: 1
Draw calls: 1
Rendered primitives (e.g., triangles): 0
Nodes: 1
Meshes: 1
Materials: 1
Animations: 0
External requests (not data uris): 3

optimize: 70.616ms
```

lilleyse:
> Note: The text pasted above the not only the same as gltf-statistics, but is also used by agi-model-inspector in the Key Statistics drop down.

agi-model-inspector uses gltf-statistics so it explains why it also says 0.

> Is this better? I reordered it a little too.

The new formatting looks great!

lilleyse:
> agi-model-inspector uses gltf-statistics so it explains why it also says 0.

Actually the model inspector seems to be up-to-date but the gltf-statistics still uses the old terminology.

shehzan10:
Pushed the new format.

lilleyse:
Nice - once https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/220#discussion_r97804477 is addressed this should be good to go.

specs/lib/getStatisticsSpec.js
shehzan10: Right it was from legacy. Thanks for catching that. Pushed the fix.

PR #221, opened 2017-01-19
exposed `triangleAxisAlignedBoundingBoxOverlap`

likangning93:


pjcozzi:
Does this warrant an update to CHANGES.md?

likangning93:
Maybe someone will do something weird but cool like write a voxelization stage as an artistic effect.

PR #223, opened 2017-01-20
Cleanup add defaults

lasalvavida:
@lilleyse, this breaks out processModelMaterialsCommon and glTF 0.8->1.0 stuff out of addDefaults.

It adds updateVersion for 0.8 -> 1.0 and moves tests for this and materials common to more appropriate locations.

addDefaults now only adds the default values of non-required variables as per its name.

This is an intermediary pull request for #191. It refactors our 1.0 processing and then #191 will contain the changes from that to get 1.0 -> 2.0, so this should be merged first.


lasalvavida:
Other change of note, if the technique is missing, we just use the default technique as per the glTF 1.0 specification instead of trying to guess a KHR_materials_common technique.

lasalvavida:
I thought about it a bit on the ride home. I liked the idea of having a glTF template object and cloning defaults out of that, but I think with the function properties here, it is a bit unnecessarily complex. I think I'm going to change this back to just being a series of functions.

lilleyse:
I don't have a problem with the template version.

lib/addDefaults.js
```diff
@@ -1,225 +1,358 @@
...
+    type : WebGLConstants.VERTEX_SHADER,
+    extras : {
+        _pipeline : {
+            source : '' +
+                'precision highp float;\n' +
+                '\n' +
+                'uniform mat4 u_modelViewMatrix;\n' +
+                'uniform mat4 u_projectionmatrix;\n' +
+                '\n' +
+                'attribute vec3 position;\n' +
```
lilleyse: `a_position`

lilleyse:
This seems like a Cesium bug (or maybe not a bug?) - when the emission value in techniques is set [0.5, 0.5, 0.5, 1.0] it still uses the material's emission value [0.0, 0.0, 0.0, 1.0]
[cow-nonormals.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/724228/cow-nonormals.gltf.txt)
cow

lilleyse:
If a model comes in without any materials, it would be nice to generate a default material corresponding with the default technique. I don't know whether the pipeline handled this before but there is some interest in this: https://github.com/AnalyticalGraphicsInc/obj2gltf/issues/32 (though this issue is more about exporting without materials).

Example model:
[cow-nonormals-no-materials.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/724259/cow-nonormals-no-materials.gltf.txt)

lilleyse:
`processModelMaterialsCommon` should set `premultipliedAlpha` to true.

lilleyse:
Also a separate bug I noticed: a uniform called `u_extras` is added to the shader in `processModelMaterialsCommon` since `technique.parameters` contains an `extras` object. This can be tested with either of the cow normals (I just had `generateNormals` commented out).

lib/processModelMaterialsCommon.js
```diff
@@ -670,12 +670,7 @@ function lightDefaults(gltf) {
     }
     var khrMaterialsCommon = extensions.KHR_materials_common;
 
-    if (defined(gltf.lights)) {
-        // glTF 0.8 backward compatibility
-        khrMaterialsCommon.lights = gltf.lights;
-        delete gltf.lights;
-    }
```
lasalvavida: Yes! I will add it

lasalvavida:
>  I don't know whether the pipeline handled this before.

It did, but it was tightly coupled with model materials common since that was how it was generating them. I'm not opposed to adding this back, but we do have the `--kmc` command line options if you want to generate non-default materials.

lasalvavida:
I think I may have misinterpreted your comment. If a glTF comes in without a technique, we used to guess technique and shader using KHR_materials_common based on the provided material parameters. We no longer do that, we just use the glTF default technique and shaders. I think this is appropriate since custom materials can be generated if desired.

For missing materials, there is a default material in glTF 2.0 but not in 1.0. That will be added to addDefaults as part of that pull request.

lilleyse:
> For missing materials, there is a default material in glTF 2.0 but not in 1.0. That will be added to addDefaults as part of that pull request.

Ah that's good to hear.

I suppose in 1.0 it's not possible to have a model without materials since `primitives.material` is required.

lasalvavida:
@lilleyse, this should be ready for another look

lilleyse:
Looks good!

PR #224, opened 2017-01-20
File formatting

shehzan10:
Taking @lasalvavida's suggestion, I have moved the file formatting fixes out of #220.

Fixes block tabs -> space tabs
Fixes trailing whitespaces.


lasalvavida:
Looks good to me, thanks @shehzan10!

shehzan10:
As a follow up to this PR, there are a lot of files in `specs` that also run into the exec issue. Should I clean those up too?

lilleyse:
Sure, more cleanup is always good.

PR #225, opened 2017-01-20
Remove +x permission from spec/data files

shehzan10:
Also includes 1 whitespace fix


lasalvavida:
Looks good to me, thanks @shehzan10!

Issue #99, opened 2016-06-20
Uniform function templates


lasalvavida:
As we continue adding pipeline stages, it is going to be important to keep a uniform template for pipeline stage functions so that they can be used in the same way.

I would propose the following:
```javascript
/**
 * Performs some operation on a glTF hierarchy.
 *
 * @param {Object} gltf An object holding a glTF hierarchy.
 * @param {Object} [options] Defines more specific behavior for this stage.
 *
 * @returns {Promise} A promise that resolves when the stage completes.
 */
function stage(gltf, options) {
    ...
}
```

pjcozzi:
Agreed.  I suggested this here: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/75#issuecomment-225271358

For an example in Cesium, see https://github.com/AnalyticalGraphicsInc/cesium/blob/3d-tiles/Source/Scene/Cesium3DTileContent.js

JoshuaStorm:
I can start going through some of our unlabeled functions. 

@lilleyse, thoughts on @lasalvavida's template before I move forward?

lasalvavida:
@JoshuaStorm We probably want to go the whole way with @pjcozzi's suggestion. That is, make a Stage.js that looks something like this:

```javascript
function Stage() {
    this.name = undefined;
    this.before = [];
    this.after = [];
};

Stage.prototype.run = function(gltf, options) {};
```

Then change the stages that currently just export functions to follow the Stage template.

```javascript
module.exports = RemoveUnusedMeshesStage;

function RemoveUnusedMeshesStage() {
    this.name = 'removeUnusedMeshes';
    this.before = ['addPipelineExtras'];
    this.after = [];
}

RemoveUnusedMeshes.prototype.run = function(gltf, options) {
    return new Promise(resolve, reject) {
        removeUnusedMeshes(gltf);
        resolve();
    }
}
```

The function bodies won't need to change until we have something set up for using the before and after to resolve dependencies, but it would be good to go ahead and set them up while we're making this change.

lasalvavida:
We may also want to start keeping stages in a subdirectory of lib and utility functions in another.

lilleyse:
Yup, all the suggestions here should be part of a larger cleanup/restructuring process, which I'm going to open an issue for pretty soon and @lasalvavida will handle.

@JoshuaStorm in the meantime there are a few other stages you can work on like pre-cache vertex optimization and transparent textures.

[...]

PR #229, opened 2017-01-25
Uniform template function for encode images stage

shehzan10:
This is an initial PR for #99.

I picked `encodeImages` to start with as it was a simple function and already returned promises.
I wanted to get feedback on how these uniform templates should be structured etc based on the comments in #99.

One of the things that I need answered is whether for a function like `EncodeImagesStage.run` we should keep the `options` parameter around as it will generate an unused function warning in jsHint.  My guess is that the answer is no and we can add the `options` parameter when needed. The counter argument to that is it would no create a uniform parameter list for all the uniform templates in.

Once I have more feedback on this, I'll work on the other stages.

### DIFF
--- a/lib/EncodeImagesStage.js
+++ b/lib/EncodeImagesStage.js
@@ -4,7 +4,18 @@ var Promise = require('bluebird');
 
 Jimp.prototype.getBufferAsync = Promise.promisify(Jimp.prototype.getBuffer);
 
-module.exports = encodeImages;
+module.exports = EncodeImagesStage;
+
+/**
+ * Encodes Jimp images in pipeline extras to buffers.
+ *
+ * The glTF asset must be initialized for the pipeline.
+ */
+function EncodeImagesStage() {
+    this.name = 'encodeImages';
+    this.before = ['quantizeAttributes'];
+    this.after = [];
+}
 
 /**
  * Encodes Jimp images in pipeline extras to buffers.
@@ -12,12 +23,13 @@ module.exports = encodeImages;
  * The glTF asset must be initialized for the pipeline.
  *
  * @param {Object} gltf A javascript object containing a glTF asset.
+ * @param {Object} options A javascript object containing options for this stage. Currently not used.
  * @returns {Promise} A promise that resolves to the glTF asset when encoding is complete for all images.
  *
  * @see addPipelineExtras
  * @see loadGltfUris
  */
-function encodeImages(gltf) {
+EncodeImagesStage.prototype.run = function(gltf, options) {
     var images = gltf.images;
     var promises = [];
     for (var imageId in images) {
@@ -45,11 +57,8 @@ function encodeImages(gltf) {
             }
         }
     }
-    return Promise.all(promises)
-        .then(function() {
-            return gltf;
-        });
-}
+    return Promise.all(promises);
+};
 
 function loadImageSource(pipelineExtras, mime) {
     var image = pipelineExtras.jimpImage;
--- a/lib/Pipeline.js
+++ b/lib/Pipeline.js
@@ -3,6 +3,7 @@ var Cesium = require('cesium');
 var Promise = require('bluebird');
 var path = require('path');
 
+var EncodeImagesStage = require('./EncodeImagesStage');
 var MergeDuplicateProperties = require('./MergeDuplicateProperties');
 var RemoveUnusedProperties = require('./RemoveUnusedProperties');
 var addDefaults = require('./addDefaults');
@@ -14,7 +15,6 @@ var combinePrimitives = require('./combinePrimitives');
 var compressIntegerAccessors = require('./compressIntegerAccessors');
 var compressTextureCoordinates = require('./compressTextureCoordinates');
 var convertDagToTree = require('./convertDagToTree');
-var encodeImages = require('./encodeImages');
 var generateModelMaterialsCommon = require('./generateModelMaterialsCommon');
 var generateNormals = require('./generateNormals');
 var generateTangentsBitangents = require('./generateTangentsBitangents');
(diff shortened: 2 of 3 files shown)


pjcozzi:
> One of the things that I need answered is whether for a function like EncodeImagesStage.run we should keep the options parameter around as it will generate an unused function warning in jsHint. My guess is that the answer is no and we can add the options parameter when needed. The counter argument to that is it would no create a uniform parameter list for all the uniform templates in.

Keep it; we do in Cesium.  Since this is implementing an interface, its declaration should match the interface.

You should be able to provide a [configuration](http://jshint.com/docs/) to disable the warning.  I don't see it being used in Cesium, but maybe it is a JavaScript vs. Node or JSHint version difference.

lib/EncodeImagesStage.js
```diff
@@ -4,20 +4,32 @@ var Promise = require('bluebird');
...
+module.exports = EncodeImagesStage;
+
+/**
+ * Encodes Jimp images in pipeline extras to buffers.
+ *
+ * The glTF asset must be initialized for the pipeline.
+ */
+function EncodeImagesStage() {
+    this.name = 'encodeImages';
+    this.before = ['quantizeAttributes'];
```
pjcozzi: Does it actually require this stage to be ran first?

lib/EncodeImagesStage.js
```diff
@@ -4,20 +4,32 @@ var Promise = require('bluebird');
 
 Jimp.prototype.getBufferAsync = Promise.promisify(Jimp.prototype.getBuffer);
 
-module.exports = encodeImages;
+module.exports = EncodeImagesStage;
+
+/**
+ * Encodes Jimp images in pipeline extras to buffers.
```
pjcozzi: The doc probably needs
```
* @alias EncodeImagesStage
* @constructor
```
like used in Cesium.

pjcozzi: Also I would not duplicate the doc for `run` here.  This just constructs the object for the stage; it can be very concise and just link to the `run` function.

Please build and verify the doc locally as well.

lib/EncodeImagesStage.js
```diff
@@ -4,20 +4,32 @@ var Promise = require('bluebird');
...
+ *
+ * The glTF asset must be initialized for the pipeline.
+ */
+function EncodeImagesStage() {
+    this.name = 'encodeImages';
+    this.before = ['quantizeAttributes'];
+    this.after = [];
```
pjcozzi: If we are going to fill out the `before`/`after` stages in this PR, we will probably want to do the dependency graph creation and traversal as part of this PR so we know these are right.

@shehzan10 are you familiar with the plan here?

There is also some graph code in Cesium that will be of interest: [Source/Renderer/ShaderSource.js](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Renderer/ShaderSource.js)

Please discuss with me if you need help.

shehzan10: My understanding from the issue comments was that the `before` and `after` would be used to run any stages that are required before/after by the current stage.

For example if stage `Y` requires `X` before and `Z` after, then before would be `X` and after would be `Y`.
If we have stages `A`, `B`, `C` that can all run independently, then none would be before or after in this case.

-----
I was planning on doing the dependency graph last - once all the stages have been made asynchronous. What do you think of this plan?

lib/EncodeImagesStage.js
shehzan10: I don't think it does. But I think I'll probably draw a dependency graph before moving forward so that the dependencies are established correctly.

shehzan10:
> Keep it; we do in Cesium. Since this is implementing an interface, its declaration should match the interface.
> 
> You should be able to provide a configuration to disable the warning. I don't see it being used in Cesium, but maybe it is a JavaScript vs. Node or JSHint version difference.

@pjcozzi This JSHint warning caused the [travis build](https://travis-ci.org/AnalyticalGraphicsInc/gltf-pipeline/jobs/195247341#L555-L569) to fail, hence the concern. I'll see if I can make the warning quiet.

lasalvavida:
@shehzan10, my only concern with this approach is that you would be renaming all of the pipeline stages, completely changing the public API and making it hard to use gltf-pipeline as a utility in other projects with standalone functions.

I think a good compromise might be to create a `lib/stages` directory with these pipeline stage objects that declare dependencies and then use the functions.

By doing this, if you just need a few single functions (like we do in Cesium), that is still easy to do, but you are responsible for making sure that stages have their dependencies met yourself.

lib/EncodeImagesStage.js
pjcozzi: The dependency graph is so, for example, if three stages all depend on stage `A`, stage `A` is only ran once, so it is required for reasonable performance.  It is OK to do it last, but it should be done as part of this PR.

There will also be some cases where we actually _don't_ want to optimize out a `before`/`after` stage and want it to run explicitly and immediately before/after the given stage.  @lasalvavida or @lilleyse should know the exact case(s) we currently have here.

For this, we probably need another set of arrays, `immediatelyBefore`/`immediatelyAfter` or whatever.  There, of course, will be some opportunities for optimization here where one stage might have stage `A` as immediate, and other stages have stage `A` as `before` (but not immediate), and stage `A` would only need to run once.

pjcozzi:
@lasalvavida interesting idea layering this on top of the current functions for individual stages.  I actually had kind of the opposite idea: most users of gltf-pipeline would actually call just one function (and maybe a cleanup function) that describes the pipeline stages they want, and this would be built on top of what @shehzan10 has done here, e.g.,

```javascript
// Not saying the API would look exactly like this
gltfPipline.run([{
        name : 'convert1to2'
    }, {
        name : 'addDefaults '
        someArgument : someValue
    }]);
```

The JSON passed in would match what can be passed to the CLI (also not designed yet).

We'll need this, but I still don't think this helps Cesium since we don't want to pull in all of gltf-pipeline, so @lasalvavida's approach sounds good.  It is layered nicely really: functions have the stage implementations and these objects represent the relationships among stages.

shehzan10:
@pjcozzi Thanks for the [clarification](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/229#discussion_r98088157). Yes, the dependency 

I think the first step from here is to draw out the graph for the various stages and how they will be related.
I will talk to @lasalvavida and @lilleyse and discuss the proposed structure of the code/API.

lilleyse:
Closing due to inactivity / switch to the 2.0 branch.

PR #231, opened 2017-01-26
Simplify and getStatistics

shehzan10:
Based on @pjcozzi suggestion, this PR does 2 things:

* Removed `PrimitiveHelpers.js` and `isDataUri.js` as dependencies of `getStatistics`
  * This require duplicating a few lines of code.
* Adds a `gulp` command `build-cesium-simplify` that exports the listed files into a independent class format that only depends on Cesium already being imported (no require).
  * The files eligible for this must only depend on Cesium and no other files.
  * This can probably be expanded in the future to require local files as well.

Note: The nomenclature (`simple`/`simplify`) maybe naive. So if there is a technical term for it, then I can change it to that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ npm run build-cesium
 
 This will output a portion of the gltf-pipeline code into the `dist/cesium` folder, reformatted into AMD style for use with RequireJS and Cesium in the browser.
 
-### Building for Other integration
+### Building for other integration
 
 Some functionality of gltf-pipeline is used by other projects along with Cesium as a third party library. The necessary files can be generated using:
 ```
--- a/README.md
+++ b/README.md
@@ -77,6 +77,15 @@ npm run build-cesium
 
 This will output a portion of the gltf-pipeline code into the `dist/cesium` folder, reformatted into AMD style for use with RequireJS and Cesium in the browser.
 
+### Building for Other integration
+
+Some functionality of gltf-pipeline is used by other projects along with Cesium as a third party library. The necessary files can be generated using:
+```
+npm run build-cesium-combine
+```
+
+This will output a portion of the gltf-pipeline code into the `dist/cesium-combined` folder, reformatted into self-contained file. Currently, only files that have no other local dependencies are allowed.
+
 ### Running Test Coverage
 
 Coverage uses [istanbul](https://github.com/gotwarlost/istanbul).  Run:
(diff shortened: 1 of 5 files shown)


pjcozzi:
Looks good, please update this part of README.md: https://github.com/AnalyticalGraphicsInc/gltf-pipeline#building-for-cesium-integration

pjcozzi:
> Note: The nomenclature (simple/simplify) maybe naive. So if there is a technical term for it, then I can change it to that.

Used `combined`/`combine`

lib/getStatistics.js
```diff
@@ -58,6 +57,28 @@ Statistics.prototype.toString = function() {
...
 };
 
+function isDataUri(uri) {
+    //Return false if the uri is undefined
+    if (!defined(uri)) {
+        return false;
+    }
+
+    //Return true if the uri is a data uri
+    return /^data\:/i.test(uri);
```
lilleyse: Add one space between `//` and the comment.

lib/getStatistics.js
```diff
@@ -97,11 +118,78 @@ function getNumberOfProperties(object) {
     return Object.keys(object).length;
 }
 
+function getDrawCallStatisticsForNode(gltf, nodeId) {
```
lilleyse: `getDrawCallStatisticsForNode` and  `getDrawCallStatistics` are similar enough that they should share a helper function that gets the statistics for an array of primitives.

lib/getStatistics.js
```diff
@@ -144,18 +232,29 @@ function getDrawCallStatistics(gltf) {
...
  */
-function getStatistics(gltf) {
+function getStatistics(gltf, nodeId) {
     if (!defined(gltf)) {
         throw new DeveloperError('gltf must be defined');
     }
 
+    var statistics = new Statistics();
+
+    if(defined(nodeId)) {   // Only do draw call statistics
```
lilleyse: `if(` -> `if (`

README.md
```diff
@@ -77,6 +77,15 @@ npm run build-cesium
 
 This will output a portion of the gltf-pipeline code into the `dist/cesium` folder, reformatted into AMD style for use with RequireJS and Cesium in the browser.
 
+### Building for Other integration
```
lilleyse: Change Other to other.

shehzan10:
@lilleyse updated.

lilleyse:
Looks good!

PR #232, opened 2017-01-27
Create .appveyor.yml for Appveyor CI (Windows)

### LINKED ISSUE #209: Possible Windows CI Options
There are a few options out there for Windows CI similar to Travis. One of the more popular ones is [AppVeyor](https://www.appveyor.com/). Similar to Travis, AppVeyor has a `appveyor.yml` to configure builds.

Here is some information on configuring for Node with AppVeyor: https://www.appveyor.com/docs/lang/nodejs-iojs/

This could be a good option to start with Windows CI.

There are other CI services out there that offer Windows. They might be worth a look too.

What are everyone's thoughts on getting started with Windows CI?

shehzan10:
This is a build config file for Appveyor, a Travis-like CI tool that runs Windows.

This is how the the console would look: https://ci.appveyor.com/project/shehzan-bot/gltf-pipeline. Appveyor is fairly intuitive as both the config file and the Appveyor setup/website are very similar to Travis, and hence have an almost zero learning curve.

Fixes #209

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,24 @@
+# Builds are whole numbered
+version: '{build}'
+
+clone_depth: 50
+
+environment:
+  matrix:
+    - nodejs_version: "4"
+    - nodejs_version: "6"
+
+install:
+  # Get the latest stable version of Node.js
+  - ps: Install-Product node $env:nodejs_version
+  # install modules
+  - npm install
+
+test_script:
+  - node --version
+  - npm --version
+  - npm run jsHint -- --failTaskOnError
+  - npm run test -- --failTaskOnError --suppressPassed
+
+# Don't actually build.
+build: off


.appveyor.yml
```diff
@@ -0,0 +1,27 @@
+# Builds are whole numbered
+version: '{build}'
+
+init:
+  - git config --global core.autocrlf input
```
mramato: How this this needs to be set?

mramato:
Nice.

Does appveryor support Linux and Mac as well?  At some point it would be great to be able to have a single CI solution for all supports OS's but I've never had the bandwidth to really evaluate them.

.appveyor.yml
shehzan10: It's not mandatory and the builds I've tested without it have been successful.

I kept it here as a safeguard for future actions (like zipping/deploying) so that the unix-style line endings that we have in the code will not be altered when cloning.

mramato: You can do whatever you feel is best, but I think it would be better to keep it as the default, since that's what users will most likely have (and for testing it's normally a good idea to mirror end-user's environments).  But like I said, up to you.  The default is also what AGI developers are also encouraged to keep on their own system.

shehzan10:
> Does appveryor support Linux and Mac as well? At some point it would be great to be able to have a single CI solution for all supports OS's but I've never had the bandwidth to really evaluate them.

Right now no. AppVeyor is Windows only, while Travis has Linux + OS X support (I was thinking of looking into the OS X enabling in the near future).

At this present time, I don't think there is a single+free+hosted CI solution (like Travis, AppVeyor etc) for all 3 major OS.
Jenkins is powerful and more to do that, but it would also mean creating an ecosystem of slaves that allows to run the builds (and all the responsibility that goes along with it).

I think at this time, sticking with a combination of Travis (Linux + OS X) and AppVeyor (Windows) is the best solution to go with.

.appveyor.yml
shehzan10: Ok, I think it makes more sense to remove the line. We can add it in the future.

mramato:
> I think at this time, sticking with a combination of Travis (Linux + OS X) and AppVeyor (Windows) is the best solution to go with.

Yeah, I agree, given what you've found.

Looking forward to a travis/Mac PR, I've experimented briefly with it and think it should work well for this repository.

shehzan10:
@mramato I force pushed the file after removing the git line ending bit.

.appveyor.yml
```diff
@@ -0,0 +1,24 @@
...
+  - npm install
+
+test_script:
+  - node --version
+  - npm --version
+  - npm run jsHint -- --failTaskOnError
+  - npm run test -- --failTaskOnError --suppressPassed
+
+# Don't actually build.
+build: off
```
mramato: Last question, I assume this line is needed because we don't have any kind of actual build step for our code (in the traditional sense)?

mramato:
This looks great to me, not sure if anyone else wanted to look at this, but feel free to merge whenever.

.appveyor.yml
shehzan10: Precisely.

Further reading if interested: https://www.appveyor.com/docs/lang/nodejs-iojs/

PR #233, opened 2017-01-30
Create initial dependency graph structure

shehzan10:
This PR contains an initial version of the dependency graph that will be used for #99.

The algorithm in summary is:
* For a given node, run any `runBefore` function based on 2 conditions:
  * If it has never been run before (marked by `hasRun`)
  * If the `runBefore` stage was a `runAfter` stage for another stage.
* Next, run the `runImmediatelyBefore` stages (always run - no conditions).
* Run the current stage. Mark `hasRun` to true and `needsToRun` to false.
* Run the `runImmediatelyAfter` stages (always run - no conditions).
* Lastly, for all `runAfter` stages, we only mark the `needsToRun` flag so that it can execute any time in the future. The following condition will make a `needsToRun` stage execute:
  * If it is a `runBefore`, `runImmediatelyBefore` or `runImmediatelyAfter` stage for any future stage.
  * If it is called explicitly.
  * If the pipeline is at an end, if will be forced to complete by the `finish` function.

The `finish` function acts like a destructor, ensuring that all stages finish.

I have discussed this with @lilleyse and it seems to hold good. I have added a test so that it can be seen in initial action.

The following are improvements that can/need to be made:
* Have a local `options` object in each node that will override any global options.
* Use an enum `state` to store one of `neverRun`, `hasRun`, or `needsToRun` flags. This will remove the need for the 2 booleans - `hasRun` and `needsToRun`.
* Figure out how to call the `run` function for each stage. On a browser it's easy but not so much in node.js. The **naive** option is to maintain a `switch-case` matching the `node.name` to the `run` function.

**This PR need not be merged. It is more for comments/review before we move forward.**

### DIFF
--- a/lib/DependencyGraph.js
+++ b/lib/DependencyGraph.js
@@ -1,0 +1,154 @@
+'use strict';
+var Cesium = require('cesium');
+
+var defined = Cesium.defined;
+var defaultValue = Cesium.defaultValue;
+var DeveloperError = Cesium.DeveloperError;
+
+module.exports = DependencyGraph;
+
+/**
+ * A node for pipeline stage.
+ *
+ * @property {String} name Stores the stage function name.
+ * @property {Array} runBefore Stages that need to run at least once before this stage.
+ * @property {Array} runImmediatelyBefore Stages that need to run exactly before this stage, irrespective of whether they have run before or not.
+ * @property {Array} runImmediatelyAfter Stages that need to run exactly after this stage, irrespective of whether they have run before or not.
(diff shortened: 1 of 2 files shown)


pjcozzi:
At quick glance, this seems brute force.  Are we sure that we shouldn't just build an actual graph data structure (like the shader generation code I referenced elsewhere) and traverse breath first with "coloring" nodes (basically setting flags) like standard graph algorithms?

The API and tests also need some work; @lilleyse can help you iterate.

lilleyse:
Closing due to inactivity / switch to the 2.0 branch.

PR #236, opened 2017-02-03
Enable OS X builds on Travis. Change coveralls -> coverage in .travis.yml

shehzan10:
* Run all tests on OSX. Skip compressed textures on linux
* Run coverage on OSX and not linux because linux will fail the compress texture tests.
* Previously, travis was running `npm run coveralls`, however, this should run `npm run coverage` to generate the report. That said, **should this run `coverage` AND `coveralls`**? 

The reason for this being the original PR where this was added https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/4/files.

I believe the workflow was to run `npm run test`, which would then run `jsHint` and `coverage` and then `coveralls` would upload this. However, now, `test` does not run coverage and there are no files for `coveralls` target to upload.

Another evidence of this is that the coveralls report has not been updated since [Jul 13, 2016](https://coveralls.io/builds/6996014).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,22 @@
+os:
+  - linux
+  - osx
+
 language: node_js
+
 node_js:
   - "4"
   - "6"
+
 script:
   - npm run jsHint -- --failTaskOnError
-  - npm run test -- --failTaskOnError --suppressPassed --excludeCompressedTextures
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]];
+    then npm run test -- --failTaskOnError --suppressPassed --excludeCompressedTextures;
+    else npm run test -- --failTaskOnError --suppressPassed;
+    fi
+
 after_success:
-## We only need to run coveralls for one node version (doesn't matter which one).
-## We also ignore publishing failures, since restarting an existing travis build would otherwise break.
-  - if node --version | grep -q ^v6 ; npm run coveralls ; fi
+  ## We only need to run coveralls for one node version (doesn't matter which one).
+  ## We also ignore publishing failures, since restarting an existing travis build would otherwise break.
+  ## Run coverage on OSX and not on linux as Linux will run compressed texture tests which will fail.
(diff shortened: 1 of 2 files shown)


pjcozzi:
> Previously, travis was running npm run coveralls, however, this should run npm run coverage to generate the report. That said, should this run coverage AND coveralls?

We want to use coveralls to track the code coverage of each PR/commit so we know if it went up or down, so please set this up however is correct.  It worked when I first enabled it, but something could have changed.

shehzan10:
@pjcozzi Using `npm run coverage && npm run coveralls` got it to work. Here is the latest [coveralls report](https://coveralls.io/builds/10005548).

pjcozzi:
Looks good!

pjcozzi:
What's the deal with the test failure: https://travis-ci.org/AnalyticalGraphicsInc/gltf-pipeline/builds/198655665

shehzan10:
> What's the deal with the test failure: https://travis-ci.org/AnalyticalGraphicsInc/gltf-pipeline/builds/198655665

This is a failure that I have seen on OS X (both locally and on travis) and have made @lilleyse aware of it. According to him, because is a "longer" test (processing each image into multiple formats at 2 levels of compression), it might be timing out.

In my observation, it has happened intermittently hence you do not see it on the other OS X build.

pjcozzi:
Let's come up with a reasonable solution so the tests are reliable, e.g.,
* Break the test into multiple tests, or
* Use a smaller image, or
* Use faster compression options, or
* Disable the test

lilleyse:
> Break the test into multiple tests

This is the easiest to do. One test per format, which a common helper function among them.

lilleyse:
All else looks good to me too.

shehzan10:
@lilleyse What I did is to create a test within a for-loop. So the test itself becomes the "helper" function and the for-loop creates the tests for each format.

shehzan10:
I think this should be safe to merge now.

PR #237, opened 2017-02-03
processModelMaterialsCommon fixes

lilleyse:
Just some minor fixes, one related to pre-multiplied alpha and the other related to deleting the extensions that are added.

@lasalvavida could you take a look here?


lasalvavida:
Deleting the extensions fix I'm good with; there's something similar in the `updateVersion` PR already. What is the purpose of multiplying the color by the transparency? Doesn't using the transparency in the alpha already apply it?

lilleyse:
The shader is premultiplying the alpha for diffuse.a already and should do it for u_transparency too.

lasalvavida:
Okay, sounds good, thanks @lilleyse.

PR #238, opened 2017-02-09
Support shader generation for different joint and weight types

lasalvavida:
Makes `processModelMaterialsCommon` handle cases where `JOINT` and `WEIGHT` are not `vec4` (they don't have to be), and cases where two primitives use the same `KHR_materials_common` but have different `type`s for `JOINT` and `WEIGHT`.

See: https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#skinned-mesh-attributes

```
Implementation note: The number of joints that influence one vertex is usually limited to 4, so that the joint indices and weights can be stored in vec4 elements.
```

This doesn't work in Cesium yet; I'll take a look at that after the model converter PR is in.

Also, Pulls in some `addPipelineExtras` improvements from `updateVersion`. I'm not getting any failing tests without the texture compression shim that was added, but someone should probably check to make sure removing it isn't breaking anything.

### DIFF
--- a/lib/addPipelineExtras.js
+++ b/lib/addPipelineExtras.js
@@ -10,8 +10,11 @@ module.exports = addPipelineExtras;
 var exceptions = {
     attributes: true,
     uniforms: true,
-    extensions: true
+    extensions: true,
+    values: true,
+    samplers: true
 };
+
 /**
  * Adds extras._pipeline to each object that can have extras in the glTF asset.
  *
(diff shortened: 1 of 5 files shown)


lilleyse:
This all looks good to me. Have you encountered any models that don't use vec4 for `JOINT` and `WEIGHT`? It would be good to test one of those.

I also confirmed that the compressed texture models still work after this change since `loadGltfUris` will create `extras._pipeline` if it doesn't exist.

lasalvavida:
Right now everyone seems to use `VEC4` because COLLADA2GLTF forced it to be that way and it seems like everyone followed the sample models However, a model like CesiumMan does have more than 4 influences specified for some vertices in the COLLADA file and it would be more appropriate to use `MAT3` in that case. I have the new converter still shimmed like this since most loaders can't handle it.

This is kind of an in-between pull request in that even if you do try to run a model that uses something besides `VEC4` it crashes when we try to put it through the Cesium geometry pipeline.

This should be fine to merge since it doesn't negatively affect models that do use `VEC4`, and models that use other types don't work right now anyway.

PR #240, opened 2017-02-14
Make generate normals optional

lilleyse:
Also adds a fix when trying to normalize zero vectors.

### DIFF
--- a/lib/generateNormals.js
+++ b/lib/generateNormals.js
@@ -129,11 +129,7 @@ function generateFaceNormals(gltf, primitive, positionSemantic) {
                     Cartesian3.subtract(positionThree, positionOne, scratchNormalTwo);
                     var normal = new Cartesian3();
                     Cartesian3.cross(scratchNormalOne, scratchNormalTwo, normal);
-                    if (Cartesian3.equalsEpsilon(normal, Cartesian3.ZERO, CesiumMath.EPSILON7)) {
-                        Cartesian3.clone(Cartesian3.UNIT_X, normal);
-                    } else {
-                        Cartesian3.normalize(normal, normal);
-                    }
+                    Cartesian3.normalize(normal, normal);
                     generatedNormals.push(normal);
                     generatedNormals.push(normal);
                     generatedNormals.push(normal);
--- a/lib/generateNormals.js
+++ b/lib/generateNormals.js
@@ -3,6 +3,7 @@ var Cesium = require('cesium');
 var clone = require('clone');
 
 var Cartesian3 = Cesium.Cartesian3;
+var CesiumMath = Cesium.Math;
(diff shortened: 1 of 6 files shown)


pjcozzi:
Update README.md with the new CLI argument.

@shehzan10 do you want to review and merge?

lib/generateNormals.js
```diff
@@ -128,7 +129,11 @@ function generateFaceNormals(gltf, primitive, positionSemantic) {
                     Cartesian3.subtract(positionThree, positionOne, scratchNormalTwo);
                     var normal = new Cartesian3();
                     Cartesian3.cross(scratchNormalOne, scratchNormalTwo, normal);
-                    Cartesian3.normalize(normal, normal);
+                    if (Cartesian3.equalsEpsilon(normal, Cartesian3.ZERO, CesiumMath.EPSILON7)) {
+                        Cartesian3.clone(Cartesian3.UNIT_X, normal);
+                    } else {
+                        Cartesian3.normalize(normal, normal);
+                    }
```
shehzan10: In https://github.com/AnalyticalGraphicsInc/cesium/pull/5010, we decided to go for a stage for degenerating these triangles. I think we should make an issue for that and not use this code for consistency.

specs/lib/generateNormalsSpec.js
```diff
@@ -42,11 +42,11 @@ describe('generateNormals', function(){
...
     });
 
-    it('generates hard normals if they do not exist', function() {
+    it('generates face normals if they do not exist', function() {
         var gltf = gltfNoNormals;
         var byteLengthBefore = 168;
         generateNormals(gltf, {
-            hardNormals : true
+            faceNormals : true
         });
```
shehzan10: Can you add a test for smooth normals too?

lilleyse: I clarified the wording in the other test, smooth normals is the default so it will be tested there.

lilleyse:
Updated.

lib/generateNormals.js
```diff
@@ -3,6 +3,7 @@ var Cesium = require('cesium');
 var clone = require('clone');
 
 var Cartesian3 = Cesium.Cartesian3;
+var CesiumMath = Cesium.Math;
```
shehzan10: This can be removed as well (from jsHint failure).

lilleyse: Ah thanks, fixed now.

lilleyse:
This is ready. (Forgot to bump earlier)

shehzan10:
The README needs to be updated with the `smoothNormals` option right.

lilleyse:
The README.md has it already, but I did add a note to `CHANGES.md`

PR #243, opened 2017-02-21
Uint16 check

### LINKED ISSUE #230: After optimizations, meshes with greater than 65535 vertices still written to ushort buffer.
This causes an exception in node. I'm seeing the issue in `combinePrimitives`.

likangning93:
Resolves #230.

`combinePrimitives` now checks if a group of primitives to be concatenated can be concatenated without overflowing `uint16` (or `uint32`, although a primitive that big scares me), and if not, it splits the group into two or more groups.

I got all excited at first thinking this was some kind of knapsack problem, but then I found out it's kind of more like the [bin packing problem](https://en.wikipedia.org/wiki/Bin_packing_problem) but possibly more complex. I tried adapting the [first fit decreasing](https://en.wikipedia.org/wiki/Bin_packing_problem#First-fit_algorithm) strategy, and under testing in a separate project it seems the implementation at least produces 'correct' results.


pjcozzi:
Thanks @likangning93!  @lilleyse can you please review?

Also note that this will be a non-issue with glTF 2.0, CC @lasalvavida, and we'll want to have fewer primitives so we have better batching and rendering performance.

lilleyse:
@tfili Does this PR work for your use case?

lilleyse:
The code and tests look good to me.

tfili:
> @tfili Does this PR work for your use case?

Yep

PR #245, opened 2017-02-23
Fix output path and handle directories as output path

### LINKED ISSUE #244: Unexpected output location
When I run the pipeline with output like this:

`node gltf-pipeline.js ... -o /d/my_folder/my_file.gltf`

I get an unexpected result, a new `output` folder is created and the actual path written is:

`/d/my_folder/output/my_file.gltf`

This is mentioned in a [comment buried in a large cleanup issue](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/130#issuecomment-232756157), but I feel this warrants its own dedicated issue.  It seems quite unnatural for it to create a subfolder when I specifically targeted a folder and filename.  It's bad for scripting too.

shehzan10:
This PR changes the following behavior with respect to output location
* If embed options are true, the output is specified by filename, it does not create a new "output" directory.
* If either embed options are false, then "output" directory is created (even if output is specified by filename).
* If output is not specified and embed are true, the then output file name is `/path/to/input-optimized.gltf` (from `/path/to/input.gltf`)
* If output is not specified and embed are false, the then output file name is `/path/to/output/input.gltf` (from `/path/to/input.gltf`)

This PR also introduces handling directories as output. In this case,
* Create directory is always false (irrespective of embed).
* The output path is `outputDir/input.gltf` (from `-i /path/to/input.gltf -o outputDir`)

I have also introduced code to handle the arguments like `-i input output` correctly. (`input -o output` was handled correctly by default before).

Fixes #244

### DIFF
--- a/lib/Pipeline.js
+++ b/lib/Pipeline.js
@@ -292,7 +292,7 @@ function writeFile(gltf, outputPath, options) {
     var binary = defaultValue(options.binary, false);
     var embed = defaultValue(options.embed, true);
     var embedImage = defaultValue(options.embedImage, true);
-    var createDirectory = defaultValue(options.createDirectory, true);
+    var createDirectory = defaultValue(options.createDirectory, !(embed && embedImage));
     var writeOptions = {
         outputPath : outputPath,
         embed : embed,
--- a/lib/parseArguments.js
+++ b/lib/parseArguments.js
@@ -229,7 +229,7 @@ function parseArguments(args) {
         return;
     }
 
-    var createDirectory = !args.s && !argv.t;
+    var createDirectory = argv.s || argv.t;
 
     var outputFileExtension;
     var fileExtension;
@@ -244,7 +244,8 @@ function parseArguments(args) {
         fileName = path.basename(gltfPath, fileExtension);
         var filePath = path.dirname(gltfPath);
         // Default output.  For example, path/asset.gltf becomes path/asset-optimized.gltf
-        outputPath = path.join(filePath, fileName + '-optimized' + outputFileExtension);
+        var outputFileName = fileName + (createDirectory ? '' : '-optimized') + outputFileExtension;
+        outputPath = path.join(filePath, outputFileName);
     } else if(path.extname(outputPath) === '') { // Handle the case where output path is directory
         createDirectory = false;
         if (argv.b) {
--- a/lib/parseArguments.js
+++ b/lib/parseArguments.js
@@ -222,7 +222,7 @@ function parseArguments(args) {
     }
 
     var gltfPath = defaultValue(argv.i, argv._[0]);
-    var outputPath = defaultValue(argv.o, argv._[1]);
+    var outputPath = defaultValue(argv.o, argv._[defined(argv.i) ? 0 : 1]);
 
     if (!defined(gltfPath)) {
         yargs.showHelp();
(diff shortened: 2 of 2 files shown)


shehzan10:
One further change we could make is to change the default created directory name from "output" to the filename of the input.

So `/path/to/myasset.gltf` will result in `/path/to/myasset/myasset.gltf`

mramato:
To be similar to other CLI tools, I don't think we should ever create a directory that isn't specified by the user.  i.e. if the user sets `/foo/bar/test.glb` as the output then all files should go in bar.  If bar (or foo) doesn't exist, we'll create them but we shouldn't create anything under bar.

This allows the user to be able to to determine what the output will be without knowing any special rules.

shehzan10:
> To be similar to other CLI tools, I don't think we should ever create a directory that isn't specified by the user.

To incorporate this suggestion, I think only the following change needs to be made. From

> If output is not specified and embed are false, the then output file name is /path/to/output/input.gltf (from /path/to/input.gltf)

We should make the output filename `/path/to/input-optimized.gltf` and other accompanying files would be `/path/to/input-optimized.*`

Does that sound good?

shehzan10:
> /path/to/input-optimized.*

Doing this style might be more complicated than I initially thought.

The problem comes down to how the properties are named within the glTF object. An example is the `CesiumTexturedBoxTestEmbedded.gltf` which produces the following sources when not embedded:
```
CesiumTexturedBoxTestEmbedded-optimized.gltf
CesiumTexturedBoxTest.bin
Image0001.png
CesiumTexturedBoxTest0FS.glsl
CesiumTexturedBoxTest0VS.glsl
```
These filenames are coming from the object property names. So there is no clear way to determine which filename to replace with the suggested ones.
I feel like unless we run through all cases for these names, we may not have a neat solution.

One solution could be to do something like `CesiumTexturedBoxTest0FS.glsl` -> `CesiumTexturedBoxTest0FS-optimized.glsl`, ie. add the `-optimized` suffix to all sources.

pjcozzi:
Is anything needed to move this along?

shehzan10:
> Is anything needed to move this along?

There needs to be consensus on how we handle the problem I described in my previous comment.

lilleyse:
For now just deriving the filename from the id is probably fine.

With gltf 2.0 not having ids anymore, the filenames can be a combination of the filename, `name` (if it exists), and its index.

mramato:
I agree with @lilleyse, other "sidecar" generated data is usually prefixed with the filename.  For example, if you have `foo.tif` and ask gdal to output `foo.png`, it outputs `foo.png` and `foo.png.xml` (which has the geolocation information.  So we should use `model.gltf` and then `model.gltf.<index>.bin` or something similar.

However, I'm not sure why we would wait to gltf and change things again.  Can't we just switch to a single convention now that will work in going forward?

shehzan10:
@mramato Just to clarify, the offline discussion that @lilleyse  and I had was to maintain status quo wrt. the "sidecar" files, ie, the files would be like this:
```
CesiumTexturedBoxTestEmbedded-optimized.gltf (main gltf file)
CesiumTexturedBoxTest.bin
Image0001.png
CesiumTexturedBoxTest0FS.glsl
CesiumTexturedBoxTest0VS.glsl
```
And any changes we make would be after 2.0.

But your suggestion would make the sidecars as:
```
CesiumTexturedBoxTestEmbedded-optimized.gltf (main gltf file)
CesiumTexturedBoxTestEmbedded-optimized.gltf.CesiumTexturedBoxTest.bin
CesiumTexturedBoxTestEmbedded-optimized.gltf.Image0001.png
CesiumTexturedBoxTestEmbedded-optimized.gltf.CesiumTexturedBoxTest0FS.glsl
CesiumTexturedBoxTestEmbedded-optimized.gltf.CesiumTexturedBoxTest0VS.glsl
```

Admittedly, the input filename is quite long so all the filenames look long. But this could be a viable approach.

Comments?

mramato:
How common is it for files to have shared textures?  It just occurred to me that renaming the texture output names could result in that use case being "broken".  Assuming that's not a use case we have to worry about, then the post 2.0 version looks fine with me.

How popular is the separate use case anyway (outside of textures)?  I pretty much always use combined models.

lilleyse:
```
CesiumTexturedBoxTestEmbedded-optimized.gltf (main gltf file)
CesiumTexturedBoxTestEmbedded-optimized_0.bin // Index in case of multiple buffers
Image0001.png
CesiumTexturedBoxTestEmbedded-optimized_0FS.glsl
CesiumTexturedBoxTestEmbedded-optimized_0VS.glsl
```

This would be my preference as I rarely see the sidecar convention with gltf models. Textures ought to stay the same name.

mramato:
@lilleyse's suggestion sounds good to me.

lilleyse:
This is fixed in the 2.0 branch.

PR #246, opened 2017-02-27
KHR_materials_common transparency fix

tfili:
The spec says that there should be a `transparent` bool set if the material has transparency. When we were generating KMC materials we weren't ever setting that flag. The caused models that needed normals generated and had transparency to be rendered opaque.


lilleyse:
Looks good.

PR #253, opened 2017-04-05
updated getBinaryGltf for gltfs with special characters

likangning93:
`getBinaryGltf` used to have a problem where gltfs with special characters would cause weird byte offset issues, since the stringified gltf's string length wouldn't be the same as the byte length.

### DIFF
--- a/lib/getBinaryGltf.js
+++ b/lib/getBinaryGltf.js
@@ -32,7 +32,7 @@ function updateBinaryObject(gltf, objects, name, pipelineExtras, state) {
                     state.currentBinaryView++;
                 }
 
-                var objectSource = new Buffer(object.extras._pipeline.source);
+                var objectSource = Buffer.from(object.extras._pipeline.source);
                 var bufferViewId = 'binary_bufferView' + state.currentBinaryView;
                 KHR_binary_glTF.bufferView = bufferViewId; //Create bufferview
                 bufferViews[bufferViewId] = {
@@ -120,7 +120,7 @@ function getBinaryGltf(gltf, embed, embedImage) {
 
     // Create padded binary scene buffer and calculate total length
     var sceneString = JSON.stringify(gltf);
-    var sceneBuffer = Buffer.alloc(Buffer.byteLength(sceneString), sceneString);
+    var sceneBuffer = Buffer.from(sceneString);
     var sceneLength = sceneBuffer.length;
     var paddingLength = 4 - (sceneLength % 4); // pad out to 4 bytes as per glb specification
     sceneLength += paddingLength;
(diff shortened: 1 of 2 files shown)


lib/getBinaryGltf.js
```diff
@@ -118,12 +118,16 @@ function getBinaryGltf(gltf, embed, embedImage) {
     // Remove extras objects before writing
     removePipelineExtras(gltf);
 
-    // Create padded binary scene string and calculate total length
+    // Create padded binary scene buffer and calculate total length
     var sceneString = JSON.stringify(gltf);
-    var sceneLength = Buffer.byteLength(sceneString);
-    sceneLength += 4 - (sceneLength % 4);
-    var padding = new Array(sceneLength + 1).join(' ');
```
likangning93: I'm not 100% sure if there was a reason for the padding to be so big before, and then get sliced down to size. Any insights?

lib/getBinaryGltf.js
```diff
@@ -118,12 +118,15 @@ function getBinaryGltf(gltf, embed, embedImage) {
...
-    sceneString = (sceneString + padding).substring(0, sceneLength);
+    var sceneBuffer = new Buffer(sceneString);
+    var sceneLength = sceneBuffer.length;
+    var paddingLength = 4 - (sceneLength % 4); // pad out to 4 bytes as per glb specification
+    sceneLength += paddingLength;
+    var paddingBuffer = new Buffer(paddingLength).fill(' ');
```
lilleyse: Use `Buffer.alloc(paddingLength, ' ')` instead. We should avoid `new Buffer` throughout but that can be a sweeping change handled later.

lib/getBinaryGltf.js
```diff
@@ -118,12 +118,15 @@ function getBinaryGltf(gltf, embed, embedImage) {
...
+    var scene = (Buffer.concat([sceneBuffer, paddingBuffer]));
```
lilleyse: No need for the outer parentheses.

lib/getBinaryGltf.js
lilleyse: I'm not sure either...

likangning93:
@lilleyse updated! Looks like a texture compression test is failing to finish in time though.

lib/getBinaryGltf.js
```diff
@@ -120,18 +120,18 @@ function getBinaryGltf(gltf, embed, embedImage) {
 
     // Create padded binary scene buffer and calculate total length
     var sceneString = JSON.stringify(gltf);
-    var sceneBuffer = new Buffer(sceneString);
+    var sceneBuffer = Buffer.alloc(Buffer.byteLength(sceneString), sceneString);
```
lilleyse: Use `Buffer.from(string)` instead

likangning93:
@lilleyse yayyy updated and everything passes!

Issue #239, opened 2017-02-09
addDefaults edge case


lilleyse:
@lasalvavida 

This model fails when running through `addDefaults`: 
[mar.glb.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/762399/mar.glb.txt)

Seems to be an issue with the `states` section.

PR #269, opened 2017-04-26
`processModelMaterialsCommon` fix

likangning93:
The `processModelMaterialsCommon` for generating glTF shaders/materials/techniques based on the `KH_model_materials_common` extension used to produce slightly out-of-spec glTF techniques.

I only caught this while trying to run a model with the extension through the pipeline twice, which caused a bizarre crash in `addDefaults`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 Change Log
 ==========
 
+### 0.1.0-alpha13 - ??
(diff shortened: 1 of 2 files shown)


lilleyse:
This may be the cause of https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/239

lilleyse:
Yes it is.. thanks for finding this.

lilleyse:
Fixes #239

PR #272, opened 2017-04-28
Update version -> 2.0 (#271)

lasalvavida:
@lilleyse, @pjcozzi, this is the PR for merging 2.0 into master


lib/uninterleaveAndPackBuffers.js
```diff
@@ -24,148 +24,122 @@ module.exports = uninterleaveAndPackBuffers;
...
+    var otherDataLength = 0;
 
-    // Change the accessor bufferViews to point to the new bufferViews
-    for (var accessorId in accessors) {
-        if (accessors.hasOwnProperty(accessorId)) {
-            var accessor = accessors[accessorId];
-            accessor.bufferView = accessor.extras._pipeline.bufferView;
+    // compute the total size for each arraybuffer type
+    ForEach.accessor(gltf, function (accessor) {
+        var bufferView = gltf.bufferViews[accessor.bufferView];
```
fanzhanggoogle: @lasalvavida I think in the most recent spec. Accessor doesn't have to have a bufferView property?

lib/uninterleaveAndPackBuffers.js
```diff
@@ -24,148 +24,122 @@ module.exports = uninterleaveAndPackBuffers;
...
-                var byteLength = byteLengthForComponentType(accessor.componentType);
-                var accessorIds = accessorsByByteLength[byteLength];
-                if (!defined(accessorIds)) {
-                    accessorIds = [];
-                    accessorsByByteLength[byteLength] = accessorIds;
-                }
-                accessorIds.push(accessorId);
-            }
-        }
+    var bufferViews = [{
```
fanzhanggoogle: In this file, it packs all the bufferViews if they are belong to an accessor and write over the original array of bufferViews. However, if there are some bufferViews are not belong to any accessor, e.g. is a property of an extension, then it won't work.

lib/uninterleaveAndPackBuffers.js
pjcozzi: Thanks for the review, @fanzhanggoogle!  You are correct, see https://github.com/KhronosGroup/glTF/issues/863

pjcozzi:
> @lilleyse, @pjcozzi, this is the PR for merging 2.0 into master

@lilleyse I most likely will not have time to review this so please merge when ready without me.

lib/uninterleaveAndPackBuffers.js
lasalvavida: @fanzhanggoogle you are correct, however the only situation where an accessor doesn't have a bufferView property is if it is a sparse accessor. That isn't something that we support here yet, however a guard to make sure it is defined may be a good idea.

lib/uninterleaveAndPackBuffers.js
lasalvavida: @fanzhanggoogle, that is a good point, but I'm not really sure how to work around it, since you'll have the same problem with the `removeUnused` functions as well. This is an optimization pipeline after all. I think the best solution we have is to support known extensions. So whichever extension specifies a bufferView will also be checked and added here.

If your example use case comes from the mesh compression you guys have been working on, I would be happy to add support based on an extension spec draft.

lib/uninterleaveAndPackBuffers.js
fanzhanggoogle: Yes, I agree with using a guard here. Just a note, cctually in the most recently 2.0 spec, there is:
"When nor sparse, neither bufferView is defined, min and max properties could have any values. This is intended for use cases when binary data is supplied by external means (e.g., via extensions)."
Which means `bufferView` could just be omitted.

lib/uninterleaveAndPackBuffers.js
fanzhanggoogle: Sounds good! Thanks. I'm actually working on a branch to add support for the mesh compression extension in glTF-pipeline, I'll make a PR once it's done. It would be great to have you review it.
I agree with you that supporting known extension is enough here.

lilleyse:
`uninterleaveAndPackBuffers` may need some further adjustments for byte alignment in 2.0:
https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/279#issuecomment-300627826

fanzhanggoogle:
Hi @lilleyse @lasalvavida , I was wondering if the PR is still active. Which is the active glTF-pipeline repo for glTF2.0? Thanks!

lilleyse:
This PR hasn't been visibly active, but there is a separate pull request coming in here soon. A lot of the code here will also be used in Cesium for converting 1.0 models to 2.0.

Long term however, this PR will stay parked on a branch until a full cleanup is complete that puts gltf 2.0 first. The working branch for that is `cleanup` and will be ready a couple months from now.

fanzhanggoogle:
Thanks. Good to know!

lilleyse:
Closing - but the keeping the branch alive since this is the basis for Cesium's implementation. Development for 2.0 will now be on the `cleanup` branch.

PR #274, opened 2017-04-28
2.0 pbr

lasalvavida:
@lilleyse, here is the PR as discussed offline.
This adds the function `pbrToMaterialsCommon`, a utility function for doing lossy mapping from PBR -> Blinn shading. This enables partial loading of PBR models in Cesium until we have a more robust rendering solution for PBR.

[Damaged Helmet](https://sketchfab.com/models/b81008d513954189a063ff901f7abfe4) sketchfab model rendered this way for reference:
damagedhelmet

@emackey, any input you have here would also be appreciated


lasalvavida:
Here's a few more examples:
lantern
corset

pjcozzi:
Thanks @lasalvavida, this will get us some mileage until we are able to get full PBR integrated.  Even after that, I wonder if this or some version of this would be useful to keep for slow devices.

We plan to derive full PBR support from https://github.com/moneimne/WebGL-PBR with help from @moneimne, @sbtron, @snagy, and crew.

pjcozzi:
@lilleyse just like #274, please merge when ready.

lib/pbrToMaterialsCommon.js
```diff
@@ -0,0 +1,119 @@
...
+ * Convert PBR materials to use a KHR_materials_common BLINN shader.
+ * This is a lossy conversion, and as such should only be used for
+ * compatibility reasons. Normal texture and metallic roughness information
+ * will be lost, and some color or texture information may be lost.
+ *
+ * @param {Object} gltf A javascript object containing a glTF asset.
+ * @returns {Object} The glTF asset with materials converted to KHR_materials_common
+ */
+function pbrToMaterialsCommon(gltf) {
+    var materialJointCount = getJointCountForMaterials(gltf);
```
lilleyse: If possible, `transparent` should be incorporated into this.

lib/pbrToMaterialsCommon.js
```diff
@@ -0,0 +1,119 @@
...
+
+        if (defined(pbrMetallicRoughness)) {
+            var baseColorFactor = pbrMetallicRoughness.baseColorFactor;
+            if (defined(baseColorFactor)) {
+                values.diffuse = baseColorFactor;
+            }
+            var baseColorTexture = pbrMetallicRoughness.baseColorTexture;
+            if (defined(baseColorTexture)) {
+                values.diffuse = [baseColorTexture.index];
+            }
```
lilleyse: Is it possible to estimate `specular` and `shininess` from roughness?

emackey: Should be possible to calculate, not estimate, I would expect.

lib/pbrToMaterialsCommon.js
```diff
@@ -0,0 +1,119 @@
...
+ */
+function pbrToMaterialsCommon(gltf) {
+    var materialJointCount = getJointCountForMaterials(gltf);
+    ForEach.material(gltf, function(material, materialId) {
+        var values = {
+            ambient : [ 0.0, 0.0, 0.0, 1.0 ],
+            diffuse : [ 0.0, 0.0, 0.0, 1.0 ],
+            emission : [ 0.0, 0.0, 0.0, 1.0 ],
+            specular : [ 0.0, 0.0, 0.0, 1.0],
+            shininess : [ 0.0 ]
```
emackey: Just a float, not an array of one float.

https://github.com/KhronosGroup/glTF/tree/2.0/extensions/Khronos/KHR_materials_common

lasalvavida: Eh, that's massively out of date... I don't really have a good answer other than that glTF 2.0 materials pre-PBR required all values to be arrays. I will put this in line with what is there currently, but it will probably change to something closer to this when it gets updated.

lib/pbrToMaterialsCommon.js
lasalvavida: @emackey do you know of any resources for this that I should be looking at?

lib/pbrToMaterialsCommon.js
lasalvavida: I think I'm actually going to keep this as is, just considering the dated nature of the KHR_materials_common spec. For example, the spec doesn't include `doubleSided` or `jointCount` as properties.

lasalvavida: Note, this will probably be replaced by: https://github.com/KhronosGroup/glTF/issues/947 at some point, so I wouldn't worry about it too much

lasalvavida:
This is updated. I'm made the structural changes that needed to be made. As far as material changes go, I've left it mostly as-is. I would say this is good to be merged into the 2.0 branch; material and specifically KHR_materials_common changes should probably be addressed at a later time.

lilleyse:
Sounds good @lasalvavida. There are a couple jsHint errors.

lasalvavida:
Updated

lasalvavida:
Okay, this should be good now

PR #275, opened 2017-04-28
Infer missing bufferView targets

lasalvavida:
@pjcozzi, @lilleyse, this infers missing bufferView targets as discussed offline

### DIFF
--- a/lib/ForEach.js
+++ b/lib/ForEach.js
@@ -1,7 +1,6 @@
 'use strict';
 var Cesium = require('cesium');
 
-var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
 
 module.exports = ForEach;
@@ -17,8 +16,13 @@ ForEach.object = function(arrayOfObjects, handler) {
     if (defined(arrayOfObjects)) {
         for (var i = 0; i < arrayOfObjects.length; i++) {
             var object = arrayOfObjects[i];
-            var offset = defaultValue(handler(object, i), 0);
-            i += offset;
+            var returnValue = handler(object, i);
+            if (typeof returnValue === 'number') {
+                i += returnValue;
+            }
+            else if (returnValue) {
+                break;
+            }
         }
     }
 };
--- a/lib/addDefaults.js
+++ b/lib/addDefaults.js
@@ -1,6 +1,7 @@
 'use strict';
 var Cesium = require('cesium');
 var addToArray = require('./addToArray');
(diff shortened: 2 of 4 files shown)


lib/addDefaults.js
```diff
@@ -428,6 +429,47 @@ function optimizeForCesium(gltf) {
     }
 }
 
+function inferBufferViewTargets(gltf) {
+    // If bufferView elements are missing targets, we can infer their type from their use
+    var needsTarget = {};
+    var shouldTraverse = false;
```
pjcozzi: Is this optimization worth it: change this boolean to an integer with the number of bufferViews that need a `target`, decrease it and set `needsTarget[bufferViewId] = false` every time a target is set, then exit the outer most `ForEach.mesh` early when it is zero.  The return value of the callback could indicate `break` if it doesn't already.

Given that this is simple, it could be worth it, I just think most models will either always have target defined or never have target defined.

pjcozzi:
LGTM, just that one comment.

lasalvavida:
@pjcozzi, updated

lib/ForEach.js
```diff
@@ -17,8 +16,13 @@ ForEach.object = function(arrayOfObjects, handler) {
     if (defined(arrayOfObjects)) {
         for (var i = 0; i < arrayOfObjects.length; i++) {
             var object = arrayOfObjects[i];
-            var offset = defaultValue(handler(object, i), 0);
-            i += offset;
+            var returnValue = handler(object, i);
+            if (typeof returnValue === 'number') {
```
pjcozzi: Is this needed?  I didn't expect it and it doesn't look like it is used.

lib/ForEach.js
```diff
@@ -17,8 +16,13 @@ ForEach.object = function(arrayOfObjects, handler) {
...
+                i += returnValue;
+            }
+            else if (returnValue) {
+                break;
```
pjcozzi: Can you add a unit test for this?

lib/ForEach.js
lasalvavida: It is used by the removal functions: [RemoveUnusedProperties.js#L42](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/2.0/lib/RemoveUnusedProperties.js#L42), to make sure that the `ForEach` adjusts the index when elements get removed.

lasalvavida:
Updated

lilleyse:
Looks good.

Issue #277, opened 2017-05-03
Model fails to round trip


mramato:
Here is a quantized version of the BrainStem sample model: [BrainStem-q.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/973843/BrainStem-q.zip)

Running `gltf-pipeline -i /BrainStem-q.gltf -o out.gltf` results in a crash with the below call stack.  It looks like we're passing an invalid offset when constructing the `Uint8Array` from the buffer.

```
Unhandled rejection RangeError: Invalid typed array length
    at new Uint8Array (native)
    at packGltfBuffer (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\lib\uninterleaveAndPackBuffers.js:64:29)
    at uninterleaveAndPackBuffers (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\lib\uninterleaveAndPackBuffers.js:32:13)
    at cesiumGeometryToGltfPrimitive (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\lib\cesiumGeometryToGltfPrimitive.js:100:5)
    at optimizeForVertexCache (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\lib\optimizeForVertexCache.js:94:21)
    at Function.Pipeline.processJSONWithExtras (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\lib\Pipeline.js:130:9)
    at C:\Git\AnalyticalGraphicsInc\gltf-pipeline\lib\Pipeline.js:261:29
    at tryCatcher (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\node_modules\bluebird\js\release\promise.js:512:31)
    at Promise._settlePromise (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\node_modules\bluebird\js\release\promise.js:569:18)
    at Promise._settlePromise0 (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\node_modules\bluebird\js\release\promise.js:614:10)
    at Promise._settlePromises (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\node_modules\bluebird\js\release\promise.js:693:18)
[...]

PR #279, opened 2017-05-09
Fix byte boundary issue

lilleyse:
There was an issue related to byte offsets when loading models converted here into Cesium. Before an index accessor might look like:

```
    "accessor_2": {
      "bufferView": "bufferView_1",
      "byteOffset": 2,
      "byteStride": 0,
      "componentType": 5125,
      "count": 784734,
      "min": [
        0
      ],
      "max": [
        169568
      ],
      "type": "SCALAR"
    }
```

Which is valid since `accessor_2` is still 4-byte aligned when `bufferView_1` itself has the correct offset from the central `buffer`.

However since Cesium creates a typed array from the buffer view, and not the buffer as a whole, `accessor_2` would begin on byte 2 of a `Uint32Array`, causing the indices to be incorrect.

The fix here forces the first accessor of each target to start with a `byteOffset` of 0 to prevent situations like this.

@mramato can you test this?


mramato:
Testing now.

mramato:
I can confirm this fixes the problem I was seeing.  Thanks!

mramato:
This also fixes #277, which was also offset related.

mramato:
@lilleyse were you going to do more work here, or is this OK to merge? I know this module is still alpha, so not sure if we're doing unit tests yet or not.

lilleyse:
Yeah I should add a unit test, I'll update soon.

mramato:
Awesome, thanks.  I have a feeling this will fix quite a few issues I've ran into recently.

lilleyse:
Added test.

mramato:
Thanks @lilleyse!  The new test fails in master and passes with this fix.

pjcozzi:
@lilleyse w.r.t the glTF spec does Cesium need a fix?  The _intent_ of the spec is to allow creating "pointers" with typed arrays instead of copying buffers.

lilleyse:
Yeah Cesium would need a fix because nothing was wrong with the previous output. Cesium would need to adjust the byte offset of where it creates the index buffer's typed array.

Does the gltf spec allow an accessor to exist on a non-aligned byte offset? That would be a more serious issue, and require a copy.

pjcozzi:
Please submit an issue to Cesium.

> Does the gltf spec allow an accessor to exist on a non-aligned byte offset? That would be a more serious issue, and require a copy.

@lexaknyazev?

lexaknyazev:
@lilleyse @pjcozzi 
Here's relevant spec section.
https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#data-alignment

The intent is that no copy is needed.

lilleyse:
Ok that is good to know. Thanks for pointing me there @lexaknyazev.

lilleyse:
Closed https://github.com/AnalyticalGraphicsInc/cesium/issues/5309

Issue #282, opened 2017-05-16
`loadGltfUris` can run out of file handles


likangning93:
`loadGltfUris` currently runs out of file handles for models that reference lots of external resources. I have a model that references over 9000 images, and it fails with:
```
Unhandled rejection Error: EMFILE: too many open files
```

likangning93:
Hmm. looks like this is because `loadGltfUris` uses `fs` instead of `fs-extra`. `fs-extra` under the hood wraps `graceful-fs`, which avoids `EMFILE` exceptions by queuing file requests or something.

PR #283, opened 2017-05-16
fs to fs-extra for loadGltfUris

likangning93:
Fixes https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/282.

Fixes a crash in the somewhat atypical use case where a gltf references a huge number of files.


lilleyse:
Nice quick fix.

PR #284, opened 2017-05-19
Fix issue with transparent diffuse texture setting the technique state

lilleyse:
In `addDefaults.enabledDiffuseTransparency` it checks if any materials have a transparent diffuse texture and if so sets the alpha render state for that technique. The problem occurs when multiple materials share the same technique and one transparent texture causes the whole model to be transparent.

The solution was to check if the technique has other users, and if so copy the technique before messing with the render state.

Sample models that show the problem are here: https://groups.google.com/forum/#!topic/cesium-dev/4cG1nDU_QwU

### DIFF
--- a/lib/addDefaults.js
+++ b/lib/addDefaults.js
@@ -347,6 +347,42 @@ function addDefaultTechnique(gltf) {
     }
 }
 
+function isTechniqueShared(gltf, techniqueId) {
+    var users = 0;
+    var materials = gltf.materials;
+    for (var materialId in materials) {
+        if (materials.hasOwnProperty(materialId)) {
+            var material = materials[materialId];
+            if (material.technique === techniqueId) {
+                ++users;
+            }
+        }
+    }
+    return users > 1;
+}
+
+function makeTechniqueTransparent(technique) {
+    technique.states = {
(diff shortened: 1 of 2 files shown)


lib/addDefaults.js
```diff
@@ -347,6 +347,42 @@ function addDefaultTechnique(gltf) {
     }
 }
 
+function isTechniqueShared(gltf, techniqueId) {
+    var users = 0;
```
pjcozzi: Nitpick, but `references` or `numberOfReferences` would have better semantics.

lib/addDefaults.js
```diff
@@ -347,6 +347,42 @@ function addDefaultTechnique(gltf) {
...
+    var materials = gltf.materials;
+    for (var materialId in materials) {
+        if (materials.hasOwnProperty(materialId)) {
+            var material = materials[materialId];
+            if (material.technique === techniqueId) {
+                ++users;
```
pjcozzi: Why not return early, i.e., `return true;` when this is `> 1`, then just always `return false;` below.

lilleyse:
Updated

PR #286, opened 2017-05-30
Add regex to remove leading underscore from shader identifiers

shehzan10:
This fixes attributes names that have leading underscores, ex. `_batchid`, which then cause attribute names to be `a__batchid`.
The double underscore is reserved in GLSL and is not allowed for identifiers.

### DIFF
--- a/lib/processModelMaterialsCommon.js
+++ b/lib/processModelMaterialsCommon.js
@@ -812,7 +812,9 @@ function ensureSemanticExistenceForPrimitive(gltf, primitive) {
                 var accessorId = attributes[semantic];
                 var accessor = accessors[accessorId];
                 var lowerCase = semantic.toLowerCase();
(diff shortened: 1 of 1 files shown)


lib/processModelMaterialsCommon.js
```diff
@@ -812,7 +812,9 @@ function ensureSemanticExistenceForPrimitive(gltf, primitive) {
                 var accessorId = attributes[semantic];
                 var accessor = accessors[accessorId];
                 var lowerCase = semantic.toLowerCase();
+                lowerCase = lowerCase.replace(/^_*/g, "");  // Remove all starting _ from name. GLSL does not allow __ double underscores
```
lilleyse: Style: change `""` to `''`

shehzan10: Whoops. My bad. Copy paste mistake. Fixed in the same commit.

lilleyse:
Looks good!

Issue #280, opened 2017-05-12
RemoveUnusedProperties in removePrimitiveAttributes tries to remove undefined technique


bkuster:
After removing the normals in the Pipeline with `removeNormals` a material is left without a `technique`. The technique referenced _before_ the normals where removed is also no longer available. In the `removeAll` step of the pipeline following the `removeNormals`, the removePrimitiveAttributes tries to access the material and its technique at *line 400* of `RemoveUnusedProperties.js` . The material exists, but the technique returns undefined, causing an unhandled rejection.

Im not sure if a) the material should have been removed at the same time as the technique its using was removed in `removeNormals` or b) the `removePrimitiveAttributes` function should be able to handle a missing technique.

bkuster:
If I try to _ignore_ the missing technique in the material, the issue reemerges in `processModelMaterialCommons.ensureSemanticExistenceForPrimitive` at line 806, when the program should be accessed. This leads me to believe the issue actually lies in the `removeNormals` step?

lilleyse:
`removeNormals` should regenerate the technique, but I guess it doesn't update the material properly. I'll look into this in a bit. Thanks for the detailed report!

PR #287, opened 2017-06-05
Fix remove normals

lilleyse:
Fixes https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/280

@shehzan10 can you check that this also fixes your problem?

### DIFF
--- a/lib/removeNormals.js
+++ b/lib/removeNormals.js
@@ -34,7 +34,7 @@ function removeNormals(gltf, options) {
             for (var i = 0; i < primitivesLength; i++) {
                 var primitive = primitives[i];
                 if (removePrimitiveNormals(primitive)) {
-                    generateMaterial(gltf, primitive, generatedMaterials, options);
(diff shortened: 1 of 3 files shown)


shehzan10:
Combined with master (for the `a__batchid` fix), this fixes the problem I had faced.

lib/removeNormals.js
```diff
@@ -1,25 +1,30 @@
...
  * @param {Object} gltf A javascript object containing a glTF asset.
+ * @param {Object} [options] Options passed to processModelMaterialsCommon.
  * @returns {Object} The glTF asset with removed normals.
  *
  * @see addPipelineExtras
  * @see loadGltfUris
  */
-function removeNormals(gltf) {
-    var materials = gltf.materials;
+function removeNormals(gltf, options) {
```
shehzan10: @lilleyse CI is failing because options is unused.

lilleyse: Fixed now.

PR #288, opened 2017-06-12
Add ESLint shareable config, remove JShint

omh1280:
Part of https://github.com/AnalyticalGraphicsInc/cesium/pull/5431.

- Add `.eslintrc.json` files in place of `.jshintrc`'s
- Update the `package.json` as appropriate
- Add eslint hints where necessary to make sure `npm run eslint` passes.
- Remove `.idea/jsLinters/jshint.xml`

This is part of an effort to move to [shareable configs](http://eslint.org/docs/developer-guide/shareable-configs): https://www.npmjs.com/package/eslint-config-cesium


omh1280:
@mramato How does this look? I'll apply the same "treatment" to the other repos if this looks OK.

.npmignore
```diff
@@ -5,7 +5,6 @@
 /output
 /specs
 /test
-.jshintrc
```
mramato: We should add the eslint file here as well.

index.js
```diff
@@ -1,3 +1,4 @@
+/*eslint strict: ["error", "function"]*/
```
mramato: Rather than ignoring the error, just add a `'use strict';` here.

lib/removeObject.js
```diff
@@ -9,7 +9,7 @@ module.exports = removeObject;
  */
 function removeObject(gltf, name, usedId) {
     // Iterate through objects and remove those that are not in the hash
-    var numberRemoved = 0;
+    var numberRemoved = 0; //eslint-disable-line no-unused-vars
```
mramato: Just remove this variable, since it looks like `numberRemoved` isn't ever read.

index.js
mramato: Rather than ignoring the error, just add a `'use strict';` here.

omh1280: Got it. For files like `specs/matchers/equals.js` as well?

omh1280:
Bump

specs/.eslintrc.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../.eslintrc.json",
+  "env": {
+    "jasmine": true
+  },
+  "rules": {
+    "no-unused-vars": "off"
```
mramato: Are we sure we want this rule here?

omh1280: I copied it from https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/288/files/30d340f9d3c9b7d222f59431bd1e11a5f078c952#diff-6deaf0988d1fe3f5d08d3f8bca32b624L4. Want me to turn it off and remove unused vars?

mramato: Yes, there are many cases in this repository where variables are really unused and should be removed

In most places where there is a false positive, disable the rule at the line level, except for files like `addDefaultMatchers` where is makes more sense at the file level.

Thanks.

omh1280:
Bump!

specs/lib/compressTexturesSpec.js
```diff
@@ -41,7 +41,7 @@ function compressGltfTexture(gltfPath, imagePath, options) {
                 basePath : basePath
             };
             return Pipeline.processJSON(gltf, pipelineOptions)
-                .then(function(gltf) {
+                .then(function(gltf) { //eslint-disable-line no-unused-vars
```
mramato: If `gltf` isn't used here, remove it.

specs/lib/removePipelineExtrasSpec.js
```diff
@@ -7,7 +7,7 @@ describe('removePipelineExtras', function() {
     var gltf;
 
     beforeAll(function(done) {
-        fs.readFile(gltfExtrasPath, function(err, data) {
+        fs.readFile(gltfExtrasPath, function(err, data) { //eslint-disable-line no-unused-vars
```
mramato: Remove this comment and fix the function by adding

```javascript
if(defined(err) {
    done.fail(err);
    return;
}
```
to the top of the function.

mramato:
Just those comments and then I'll merge.  Thanks @omh1280!

omh1280:
Thanks! Bump

specs/lib/removePipelineExtrasSpec.js
```diff
@@ -1,13 +1,18 @@
 'use strict';
 var fs = require('fs');
+var defined = require('cesium').defined;
```
mramato: Coding guidelines request that this be written as 

```javascript
var defined = Cesium.defined;
```

Plus, the `var removePipelineExtras = require('../../lib/removePipelineExtras');` should be moved up to be with the other requires, and defined should be after that.

omh1280:
thanks. Bump!

mramato:
Thanks.

PR #293, opened 2017-06-23
Fix errors from ESLint shared config 2.0

omh1280:
For https://github.com/AnalyticalGraphicsInc/cesium/issues/5456.

Fix all the errors ESLint throws with the new 2.0 additions. Includes [global-require](http://eslint.org/docs/rules/global-require), [no-buffer-constructor](http://eslint.org/docs/rules/no-buffer-constructor), and [no-new-require](http://eslint.org/docs/rules/no-new-require) rules.

I disabled the `quotes` rule because of how much it is broken in this repo. I assume the double quotes are for JSON objects. If it's decided that we should use single quotes, it will be an easy change with `--fix`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,3 +1,7 @@
 {
-  "extends": "cesium/node"
+  "extends": "cesium/node",
+  "rules": {
+    "no-buffer-constructor": "off",
+    "quotes": "off"
+  }
 }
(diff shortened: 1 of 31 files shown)


pjcozzi:
@likangning93 do you want to review this?

likangning93:
Changes look good!

PR #297, opened 2017-07-03
Fix shader comparison

### LINKED ISSUE #296: gltf -> glb conversion breaks model
1. Download http://assets.agi.com/models/orbview3.gltf
2. Run `node bin/gltf-pipeline.js -i orbview3.gltf -o orbview3.glb -b`
3. Crash

```
Unhandled rejection TypeError: shaderOne.extras._pipeline.source.equals is not a function
    at shaderEquals (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\lib\MergeDuplicateProperties.js:227:46)
    at Function.MergeDuplicateProperties.mergeShaders (C:\Git\AnalyticalGraphicsInc\gltf-pipeline\lib\MergeDuplicateProperties.js:189:17)
```

The problem is that `sharedEquals` assumes that `shaderOne.extras._pipeline.source` and `shaderTwo.extras._pipeline.source` are both buffers, but they are actually both strings in this case.

Because of this, I'm not sure where the bug actually is, should these values always be buffers?  Always be strings?  Or can they be either?

I feel like this is a bug we keep running into where we assume source is one thing and then we get a crash because it's another.

If I update the code to just do `===`, it no longer crashes, but the model loads incorrectly with bad textures.

CC @likangning93 @lilleyse

lilleyse:
Solves the crash in #296, but not the conversion problem

@mramato

### DIFF
--- a/specs/lib/MergeDuplicatePropertiesSpec.js
+++ b/specs/lib/MergeDuplicatePropertiesSpec.js
@@ -139,8 +139,6 @@ describe('MergeDuplicateProperties', function() {
 
     var mergeShaders = MergeDuplicateProperties.mergeShaders;
     describe('mergeShaders', function() {
-        var testShaderBufferOne = Buffer.from('test shader one', 'utf8');
-        var testShaderBufferTwo = Buffer.from('test shader two', 'utf8');
         it('merges duplicate shaders', function() {
             var gltf = {
                 programs : {
@@ -158,7 +156,7 @@ describe('MergeDuplicateProperties', function() {
                         type : WebGLConstants.VERTEX_SHADER,
                         extras : {
                             _pipeline : {
-                                source : testShaderBufferOne
+                                source : 'test shader one'
                             }
                         }
                     },
@@ -166,7 +164,7 @@ describe('MergeDuplicateProperties', function() {
                         type : WebGLConstants.FRAGMENT_SHADER,
                         extras : {
                             _pipeline : {
-                                source : testShaderBufferOne
+                                source : 'test shader one'
                             }
                         }
                     },
@@ -174,7 +172,7 @@ describe('MergeDuplicateProperties', function() {
                         type : WebGLConstants.VERTEX_SHADER,
                         extras : {
                             _pipeline : {
-                                source : testShaderBufferTwo
+                                source : 'test shader two'
                             }
                         }
                     },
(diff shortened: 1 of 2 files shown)


mramato:
Doesn't this change imply we're missing some pretty critical unit tests?  We've had several problems in this area.

likangning93:
The spec for `MergeDuplicateProperties` spoofs glTF with pipeline extras and assumes that shaders are stored as buffers in `extras._pipeline.source`, I think that's why we didn't catch this before.

mramato:
> The spec for MergeDuplicateProperties spoofs glTF with pipeline extras and assumes that shaders are stored as buffers in extras._pipeline.source, I think that's why we didn't catch this before.

Sounds like we have bad specs as well then.  We should be mocking/spying on as little as possible in our specs precisely because it leads to problems like this.

lilleyse:
Thanks for pinpointing the issue @likangning93. Fixed the test. It passed before because shaders referenced the same Buffer objects which would pass both the `===` check and the `.equals` check.

lilleyse:
@likangning93 if this looks good can you merge? Then I'll push a new release out.

specs/lib/MergeDuplicatePropertiesSpec.js
```diff
@@ -139,8 +139,6 @@ describe('MergeDuplicateProperties', function() {
 
     var mergeShaders = MergeDuplicateProperties.mergeShaders;
     describe('mergeShaders', function() {
-        var testShaderBufferOne = Buffer.from('test shader one', 'utf8');
-        var testShaderBufferTwo = Buffer.from('test shader two', 'utf8');
         it('merges duplicate shaders', function() {
```
mramato: This test is basically useless, why doesn't it should load and use real gltf files, not mock what it expects the source to look like.

Given you want to get out a new release, I'm okay with @likangning93 merging this, but I'll write up an issue to fix this and other tests that do this.

mramato:
Submitted #300

specs/lib/MergeDuplicatePropertiesSpec.js
lilleyse: I agree here. In general we have strayed away from creating too many test gltf files on disk, but maybe a middle ground is good.

Issue #242, opened 2017-02-17
Check for degenerate triangles


lilleyse:
Generating smooth normals goes through Cesium's `GeometryPipeline.computeNormal` function which will throw when normalizing a zero-vector normal. Rather than patching up broken normals we should fix the root cause which seems to be degenerate triangles. For this we can use some of the code lifted from @likangning93. This also affects face normals, so that may be an easier place to start. 

Sample model: `node bin/gltf-pipeline.js face.gltf -f` or `node bin/gltf-pipeline.js face.gltf -m`
[face.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/783520/face.gltf.txt)

```javascript
'use strict';
var Cesium = require('cesium');

var Cartesian3 = Cesium.Cartesian3;
var CesiumMath = Cesium.Math;
var defined = Cesium.defined;
var DeveloperError = Cesium.DeveloperError;

module.exports = isTriangleDegenerate;

var triangleSideAScratch = new Cartesian3();
var triangleSideBScratch = new Cartesian3();
var crossProductResultScratch = new Cartesian3();

/**
 * Checks whether or not the triangle defined by the given positions has an area of zero.
 *
 * @param {Cartesian3} p0 Position of one triangle vertex.
 * @param {Cartesian3} p1 Position of one triangle vertex.
 * @param {Cartesian3} p2 Position of one triangle vertex.
 *
 * @returns {Boolean} Whether the triangle is degenerate.
 */
function isTriangleDegenerate(p0, p1, p2) {
    if (!defined(p0) || !defined(p1) || !defined(p2)) {
        throw new DeveloperError('p0, p1, and p2 are required.');
    }

    // are any of the points basically identical?
    if (Cartesian3.equalsEpsilon(p0, p1, CesiumMath.EPSILON10)) {
        return true;
    }
    if (Cartesian3.equalsEpsilon(p1, p2, CesiumMath.EPSILON10)) {
        return true;
    }
    if (Cartesian3.equalsEpsilon(p2, p0, CesiumMath.EPSILON10)) {
        return true;
    }

    // is the area 0?
    var sideA = Cartesian3.subtract(p1, p0, triangleSideAScratch);
    var sideB = Cartesian3.subtract(p2, p0, triangleSideBScratch);
    var cross = Cartesian3.cross(sideA, sideB, crossProductResultScratch);
[...]

PR #298, opened 2017-07-06
Fixed alignment issues

tfili:
Fixed issues with merging buffers that weren't a multiple of 4.

I wrote up #299 because there are some issues that prevented me from writing a test for the degenerate triangles. It opened a can of worms that was outside the scope of this PR.

### DIFF
--- a/lib/generateNormals.js
+++ b/lib/generateNormals.js
@@ -5,6 +5,7 @@ var clone = require('clone');
 var Cartesian3 = Cesium.Cartesian3;
 var GeometryPipeline = Cesium.GeometryPipeline;
 var WebGLConstants = Cesium.WebGLConstants;
+var CesiumMath = Cesium.Math;
 var defaultValue = Cesium.defaultValue;
 var defined = Cesium.defined;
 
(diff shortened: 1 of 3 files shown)


lib/mergeBuffers.js
```diff
@@ -41,8 +41,19 @@ function mergeBuffers(gltf, bufferId) {
                         bufferView.buffer = bufferId;
                     }
                 }
-                buffersToMerge.push(buffer.extras._pipeline.source);
-                lengthSoFar += buffer.extras._pipeline.source.length;
+
+                var sourceBuffer = buffer.extras._pipeline.source;
+                var currentBufferLength = sourceBuffer.length;
+                if (currentBufferLength % 4)
```
likangning93: Is this 4 because of `Float32` based on the implementation note here? https://github.com/KhronosGroup/glTF/tree/master/specification/1.0#bufferview-and-accessor-byte-alignment

 It might be worthwhile to add a blurb summarizing why the 4-byte alignment is needed.

lib/generateNormals.js
```diff
@@ -127,8 +128,23 @@ function generateFaceNormals(gltf, primitive, positionSemantic) {
                     Cartesian3.subtract(positionTwo, positionOne, scratchNormalOne);
                     Cartesian3.subtract(positionThree, positionOne, scratchNormalTwo);
                     var normal = new Cartesian3();
-                    Cartesian3.cross(scratchNormalOne, scratchNormalTwo, normal);
-                    Cartesian3.normalize(normal, normal);
+
+                    // Check for degenerate triangles
```
likangning93: When this is merged, please update https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/242 to note that this check is here.

lib/mergeBuffers.js
tfili: Yep. One thing that the note doesn't say is that a float must also start on 4-byte boundary. My particular case had something like 102 bytes of shorts containing batch ids so every float array after it in the buffer was unable to be read. The C++ converter handles this, but during the processing the padding is thrown out, so the problem creeps back in again.

likangning93:
Thanks @tfili!

PR #301, opened 2017-07-11
Added processPbrMetallicRoughness

moneimne:
@lilleyse, this is the file that has been tested within Cesium.


lib/processPbrMetallicRoughness.js
```diff
@@ -0,0 +1,774 @@
...
+
+        vertexShader += 'void main(void) {\n';
+        vertexShader += vertexShaderMain;
+        vertexShader += '}\n';
+
+        // Fragment shader lighting
+        fragmentShader += 'const float M_PI = 3.141592653589793;\n';
+
+        var lambertianDiffuse = '';
+        lambertianDiffuse += 'vec3 lambertianDiffuse(vec3 baseColor) {\n';
```
lilleyse: In GLSL code we put the `{` on a separate line.

lib/processPbrMetallicRoughness.js
```diff
@@ -0,0 +1,774 @@
...
+        GGX += 'float GGX(float roughness, float NdotH) {\n';
+        GGX += '  float roughnessSquared = roughness * roughness;\n';
+        GGX += '  float f = (NdotH * roughnessSquared - NdotH) * NdotH + 1.0;\n';
+        GGX += '  return roughnessSquared / (M_PI * f * f);\n';
+        GGX += '}\n\n';
+
+        fragmentShader += lambertianDiffuse + fresnelSchlick2 + fresnelSchlick + smithVisibilityG1 + smithVisibilityGGX + GGX;
+
+        var fragmentShaderMain = '';
+        fragmentShaderMain += 'void main(void) {\n';
```
lilleyse: Combine these two.

lib/processPbrMetallicRoughness.js
```diff
@@ -0,0 +1,774 @@
...
+        vertexShader += vertexShaderMain;
+        vertexShader += '}\n';
+
+        // Fragment shader lighting
+        fragmentShader += 'const float M_PI = 3.141592653589793;\n';
+
+        var lambertianDiffuse = '';
+        lambertianDiffuse += 'vec3 lambertianDiffuse(vec3 baseColor) {\n';
+        lambertianDiffuse += '  return baseColor / M_PI;\n';
+        lambertianDiffuse += '}\n\n';
```
lilleyse: Doing a multiline string is usually cleaner. Also use 4 spaces of indentation throughout.

```
var lambertianDiffuse =
    'vec3 lambertianDiffuse(vec3 baseColor) {\n' +
    '    return baseColor / M_PI;\n' +
    '}\n\n';
```

lib/processPbrMetallicRoughness.js
```diff
@@ -0,0 +1,774 @@
...
+            } else {
+                fragmentShaderMain += '  float roughness = 1.0;\n';
+            }
+        }
+        fragmentShaderMain += '  vec3 v = -normalize(v_positionEC);\n';
+
+        // Generate fragment shader's lighting block
+        var fragmentLightingBlock = '';
+        fragmentLightingBlock += '  vec3 lightColor = vec3(1.0, 1.0, 1.0);\n';
+        fragmentLightingBlock += '  vec3 l = normalize(czm_sunDirectionEC);\n';
```
lilleyse: Using `czm_sunDirectionEC` should only happen if `optimizeForCesium` is true. Otherwise the light direction should originate from the camera (0.0, 0.0, 1.0) like how `processModelMaterialsCommon` handles it.

lib/processPbrMetallicRoughness.js
```diff
@@ -0,0 +1,774 @@
...
+
+        fragmentShader += lambertianDiffuse + fresnelSchlick2 + fresnelSchlick + smithVisibilityG1 + smithVisibilityGGX + GGX;
+
+        var fragmentShaderMain = '';
+        fragmentShaderMain += 'void main(void) {\n';
+
+        // Add normal mapping to fragment shader
+        if (hasNormals) {
+            fragmentShaderMain += '  vec3 ng = normalize(v_normal);\n';
+            if (defined(parameterValues.normalTexture) && (hasTangents || frameState.context._standardDerivatives)) {
```
lilleyse: Standard derivatives will need to handled differently since we don't have this info available at conversion time. Check if there is a `#define` associated with the `GL_OES_standard_derivatives` extension, and add that in here. Check out https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Shaders/ShadowVolumeFS.glsl as an example.

lib/processPbrMetallicRoughness.js
```diff
@@ -0,0 +1,774 @@
...
+        for (var additional in material) {
+            if (material.hasOwnProperty(additional) && ((additional.indexOf('Texture') >= 0) || additional.indexOf('Factor') >= 0) || additional === 'doubleSided') {
+                parameterValues[additional] = material[additional];
+            }
+        }
+
+        var vertexShader = 'precision highp float;\n';
+        var fragmentShader = 'precision highp float;\n';
+
+        var skin = gltf.skins[0];
```
moneimne: It doesn't seem to say explicitly in the spec, but all of the test models at least have an empty `gltf.skins` . This could be the default behavior of exporters/coverters though, so I'll put a check just in case.

lib/processPbrMetallicRoughness.js
```diff
@@ -0,0 +1,720 @@
...
+    fragmentShader += '    float G = smithVisibilityGGX(alpha, NdotL, NdotV);\n';
+    fragmentShader += '    float D = GGX(alpha, NdotH);\n';
+
+    fragmentShader += '    vec3 diffuseContribution = (1.0 - F) * lambertianDiffuse(baseColor);\n';
+    fragmentShader += '    vec3 specularContribution = F * G * D / (4.0 * NdotL * NdotV);\n';
+    fragmentShader += '    vec3 color = NdotL * lightColor * (diffuseContribution + specularContribution);\n';
+
+    if (optimizeForCesium) {
+        fragmentShader += '    vec3 diffuseIrradiance = vec3(0.5);\n';
+        fragmentShader += '    vec3 specularIrradiance = textureCube(czm_cubeMap, r).rgb;\n';
```
abwood: Should this be u_SpecularEnvSampler instead of using a Cesium builtin?

lib/processPbrMetallicRoughness.js
```diff
@@ -0,0 +1,720 @@
...
+        fragmentShader += '    specularIrradiance = mix(specularIrradiance, diffuseIrradiance, roughness);\n'; // Fake LOD
+        fragmentShader += '    vec2 brdfLUT = texture2D(czm_brdfLUT, vec2(NdotV, 1.0 - roughness)).rg;\n';
```
abwood: Same comment as above, should you only use a Cesium builtin if optimizeForCesium?

lib/processPbrMetallicRoughness.js
```diff
@@ -0,0 +1,720 @@
...
+    fragmentShader += '    vec3 F = fresnelSchlick2(r0, r90, VdotH);\n';
+    fragmentShader += '    float G = smithVisibilityGGX(alpha, NdotL, NdotV);\n';
+    fragmentShader += '    float D = GGX(alpha, NdotH);\n';
+
+    fragmentShader += '    vec3 diffuseContribution = (1.0 - F) * lambertianDiffuse(baseColor);\n';
+    fragmentShader += '    vec3 specularContribution = F * G * D / (4.0 * NdotL * NdotV);\n';
+    fragmentShader += '    vec3 color = NdotL * lightColor * (diffuseContribution + specularContribution);\n';
+
+    if (optimizeForCesium) {
+        fragmentShader += '    vec3 diffuseIrradiance = vec3(0.5);\n';
```
abwood: Will there be an option to sample a diffuse environment map too? Similar to how we sample diffuse in the glTF-WebGL-PBR reference implementation.

lib/processPbrMetallicRoughness.js
moneimne: I should probably clarify the name, but I added the specular environment map as a Cesium builtin for now since anyone using Cesium will use the same map.

lib/processPbrMetallicRoughness.js
moneimne: Yup! Made this change in a recent commit.

moneimne: Yup! Made this change in a recent commit.

lib/processPbrMetallicRoughness.js
moneimne: As for now, we are using a very simple gradient as the environment map, so I thought a diffuse gray would be acceptable for the diffuse irradiance. The change to using a proper diffuse environment map would be simple to add, but I think this can come with a switch to a more complex environment map.

moneimne:
@lilleyse, all the models from the sample repo have been tested and load correctly into Cesium when passed through the pipeline.

lilleyse:
The latest code changes look good.

There are some jsHint errors and test failures.

(Yes, this branch and `2.0` are pre-eslint).

lilleyse:
Thanks @moneimne. Will merge this into the `2.0` branch and then leave that parked for a little while.

PR #305, opened 2017-08-10
Use uniqueProgram/uniqueTechniques to retrive unique programs/techniques

shehzan10:
* This is already being done for materials
* If this is not done, then the comparison is happening between objects
where one has name/extras and the other doesn't. This the comparison
will fail.
* I pseudo-randomly added `extras` property to program/technique/material objects in `MergeDuplicateProperties` spec such that the properties should merge regardless of the `extras` property.


lilleyse:
Nice simple fix!

PR #306, opened 2017-08-11
Skip merge accessors before combine nodes/meshes

shehzan10:
Running merge accessors before nodes/meshes when multiple accessors can
be independent but have the same data causes one of the accessors to be
deleted and both meshes to use the same accessor. When this happens,
combineNodes can no longer merge the meshes.

Skipping merge accessors allows the nodes to be merged first and the
accessors can be merged during the second call to
mergeDuplicateProperties.

### DIFF
--- a/lib/MergeDuplicateProperties.js
+++ b/lib/MergeDuplicateProperties.js
@@ -30,7 +30,7 @@ function MergeDuplicateProperties() {}
  */
 MergeDuplicateProperties.mergeAll = function(gltf, skipMergeAccessors) {
     skipMergeAccessors = defaultValue(skipMergeAccessors, true);
-    if(!skipMergeAccessors) {
+    if (!skipMergeAccessors) {
         MergeDuplicateProperties.mergeAccessors(gltf);
     }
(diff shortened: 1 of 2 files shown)


lilleyse:
Good fix for now. This sort of issue will be fixed in the cleanup branch.

lasalvavida:
I'm not sure you actually want to do this.

The reason that combineNodes doesn't merge meshes that share attribute data is because we flatten out transforms into the data, so if two meshes use the same data and the transforms are different, we can't apply different transforms to the same data.

If you skip this step, a duplicated accessor may get its data transformed by combineNodes and then they won't be combined later because the data has changed, bloating the model unnecessarily. 

I think what you really want here is to add a smarter check to combineNodes for the case where two meshes contain primitives with shared attribute accessors, but the transform that would be applied is the same, so there is no conflict.

lilleyse:
Hit merge right as you left your comment... we can fix/revert if needed.

lasalvavida:
Haha sorry guess I need to learn to type faster

PR #308, opened 2017-08-22
Lets primitives combine if the non-transformable attribute is shared

tfili:
This undoes #306, which was the wrong solution as discussed in the comments of that PR.

When combining nodes, we only care if data is shared because we flatten the tree and bake the transforms into the vertex data. We only transform `POSITION` and `NORMAL` attributes, so other data can be shared and it won't matter as it remains unmodified.

CC @lasalvavida @shehzan10 @lilleyse


lib/PrimitiveHelpers.js
```diff
@@ -44,6 +45,30 @@ function primitivesShareAttributeAccessor(primitive, comparePrimitive) {
 }
 
 /**
+ * Compare two primitives to check if they share an transformable attribute (position and normal) accessor.
+ *
+ * @param {Object} primitive The first primitive to compare.
+ * @param {Object} comparePrimitive The second primitive to compare.
+ * @returns {Boolean} True if primitives share an attribute, false otherwise.
+ */
+function primitivesShareTransformableAttributeAccessor(primitive, comparePrimitive) {
```
lasalvavida: Maybe add an extra optional argument to `primitivesShareAttributeAccessor` that is an object with the attributes to match against if it is present. Then call the function as `primitivesShareAttributeAccessor(primitive, comparePrimitive, {POSITION: true, NORMAL: true})`

shehzan10:
Shouldn't we change this
https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/aabfa2558690bfa1658fc4aee1a23fc4a1116210/lib/PrimitiveHelpers.js#L251 to `primitivesShareTransformableAttributeAccessor` as well?

tfili:
> Shouldn't we change this to primitivesShareTransformableAttributeAccessor as well?

@shehzan10 I think so. I'm not super familiar with this code. @lilleyse is there something I'm missing or is good to change it here as well?

If so I can remove the original version of this function? @lilleyse what is the policy on removing functions from the `PrimitivesHelper` class, as the function was public.

shehzan10:
The reason I found that to be a suitable change is because my test model has a color buffer and with this PR, I found that it is merging nodes, but not merging primitives as all the primitives use the same color accessor. In this case, this condition fails.

tfili:
@shehzan10 @lilleyse I think this is done.

lasalvavida:
`combinePrimitives` uses the primitiveConflicts to decide whether primitives can be combined or not. No transforms are done to the data during this process, so only checking "transformable attributes" is almost surely going to result in something you're not expecting at some point.

@shehzan10, if you can provide the model that isn't combining, I may be able to provide some insight into why that is happening

shehzan10:
@lasalvavida Here is the data. This glTF is a naive merge of very similar 6 glTFs. They are similar in that they have the same number of primitives and their color buffers are identical. My assumption is that because the color buffers are identical, gltf-pipeline is removing the duplicates and then not merging because all the primitives use the same accessor for colors.

[merged.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/1249894/merged.gltf.txt)

lasalvavida:
The combining primitives logic is defined for two cases: primitives that have separate indices accessors and shared attribute accessors where we can just append the indices accessors into a single primitive, and primitives that have separate indices and attribute accessors, where we have to append the attributes together and modify the indices of each primitive with the new offsets. 

There is no reasonable strategy for combining primitives with both shared and unshared attributes. Duplicating the shared attributes and appending them along with the unshared attributes bloats the model.

Since the color attributes contain the same data, they get replaced with a single accessor. Now you have primitives with shared and unshared attributes. It is expected behavior for these primitives to not be combined in order to reduce the model size.

I hope that clears this up.

tfili:
Thanks. @lasalvavida that makes sense.

tfili:
Ok, so I wrote some code to check to make sure we don't leave any primitives in a mixed state (some accessors shared and some not). Now I'm thinking, is there really a point to removing duplicate accessors?

Here is my thinking:
* We only remove duplicate accessors if the size, count, type and actual data is the same.
* The accessors for each primitive will only be merged if there is another primitive where all accessors can be merged as well. For instance for any accessor below to be merged all position, normal and color accessors must be identical. In the same mesh, that seems really unlikely as they would be rendered in the same spot. Checking all the data is slow for detecting a problem that isn't likely to happen.
```
{
  POSITION: accessorA1,
  NORMAL: accessorA2,
  COLOR: accessorA3
},
{
  POSITION: accessorB1,
  NORMAL: accessorB2,
  COLOR: accessorB3
}
```
* Now if we add another primitive like this, then if `accessorC1` is different from `accessorB1`, then no accessors can be merged at all since `accessorB3` can't be merged at all and either can any `accessorB*`. This is where all the problems were coming from with it merging batch ids and colors, but nothing else.
```
{
  POSITION: accessorC1,
  NORMAL: accessorC2,
  COLOR: accessorB3
}
```
* Now if there is a case where there is duplicate primitives between meshes, there is probably a really good chance its because they are in different nodes with different transforms. We need to make the call what we want here. Do we want to be able to merge nodes, bake in transforms and reduce draw calls? Or do we want to merge primitives and reduce the data size and increase draw calls?
* If we aggressively merge accessors, then we force the user to have more draw calls regardless of what they do. If we remove it, we allow the user to create a model with shared accessors and more draw calls or separate accessors and less draw calls.

Did we actually have problems with models having a bunch of identical accessors or did we preemptively write a `mergeAccessor` function because we just wrote functions to merge duplicates of all type?

@lilleyse @likangning93 @lasalvavida Thoughts?

lasalvavida:
> so I wrote some code to check to make sure we don't leave any primitives in a mixed state
I think you're misunderstanding me here. We do this on purpose, it isn't something we want to prevent. This project optimizes for smaller model sizes to improve transmission times, not for fewer draw calls. 

> Did we actually have problems with models having a bunch of identical accessors
Yes. There are tons of COLLADA files that contain duplicate data and the old COLLADA2GLTF wasn't smart enough to optimize it out, so it would just map the COLLADA data 1:1 with glTF accessors.

You are correct in that if you wanted to optimize for fewer draw calls, you wouldn't want to merge accessors. If you want to add a flag like `--optimizeDrawCalls` for your use case that skips merging accessors and allows mixed primitives to combine, that would be a more appropriate solution.

tfili:
Yes, I understand that. I'm just not sure if that is what we want moving forward. Speeding up a one time download and slowing rendering that happens 30 times a second seems like the wrong approach in normal cases.

Anyway, I added a `optimizeDrawCalls` option that doesn't merge accessors.

This should be ready for another look.

lasalvavida:
> Speeding up a one time download and slowing rendering that happens 30 times a second seems like the wrong approach in normal cases.

I don't disagree, and maybe optimizing for draw calls should be the default setting.

However, I think it's worth saying that optimizing for the attribute buffer to be smaller doesn't just speed up a one time download, it also reduces the model's footprint in video memory. That might be something that you want depending on your use case, so I think it's worth keeping it as an optimization pathway.

lasalvavida:
A few other things to do if you want to optimize for draw calls:
 * A stage should be added that looks for accessors shared by primitives, clones the accessor and separates the references.
 * In combineNodes if a mesh is instantiated on two different nodes (like the parts in a lot of CAD models), you would want to duplicate the mesh and its primitives so that the data can be transformed up the node tree without data conflicts.

To elaborate on my last point; duplicating the data in this way could potentially increase the size of the optimized model several times over, especially for CAD style models. On mobile chipsets with limited VRAM, this could be the difference between the model loading somewhat slower performance vs. the model just not loading at all.

Doing this also destroys the node/mesh/primitive hierarchy along with all of the names of the elements (this is probably a better reason not to optimize for draw calls by default).

tfili:
Cool. Thanks @lasalvavida for the suggestions. Having separate options probably does make sense.

lib/Pipeline.js
```diff
@@ -85,6 +85,7 @@ Pipeline.processJSON = function(gltf, options) {
  * @param {Object} [options.quantize] Flag to run quantizeAttributes stage.
  * @param {Object} [options.preserve=false] Flag to turn optimization pipeline stages on/off to preserve the original glTF hierarchy.
  * @param {Boolean} [options.mergeVertices=false] Flag to merge duplicate vertices, which can produce a smaller model size but greatly increases conversion time. This setting only applies when options.preserve is false.
+ * @param {Boolean} [options.optimizeDrawCalls=false] Flag to merge duplicate vertices, which can produce a smaller model size but greatly increases conversion time. This setting only applies when options.preserve is false.
```
lilleyse: Fix description.

tfili: Good catch. Should be all good now @lilleyse

PR #313, opened 2017-08-29
Add minimum lighting limit

lilleyse:
Fixes https://github.com/AnalyticalGraphicsInc/cesium/issues/5787

Prevents models using the `KHR_materials_common` extension from getting too dark, but only in Cesium.


lilleyse:
@shehzan10 can you review?

PR #316, opened 2017-09-08
Bump npm module versions

mramato:
These two modules had major out of date versions but were none of our code needed to be updated to support them.

We should really set up greenkeeper.


lilleyse:
👍

PR #318, opened 2017-09-19
Fixes in modelMaterialsCommon

shehzan10:
* Fixes how `hasNormals` is determined - take `meshes.primitives.attributes.NORMAL` into account.
* Set the model view semantic after checking `CESIUM_RTC` extension.
* Lighting fix from 5655411

### DIFF
--- a/lib/processModelMaterialsCommon.js
+++ b/lib/processModelMaterialsCommon.js
@@ -230,10 +230,16 @@ function generateTechnique(gltf, materialId, khrMaterialsCommon, lightParameters
     var hasNormals = (lightingModel !== 'CONSTANT' && meshesHaveNormals);
 
     // Add techniques
+
+    var modelViewSemantic = 'MODELVIEW';
+    if (gltf.extensionsUsed.indexOf('CESIUM_RTC') >= 0) {
+        modelViewSemantic = 'CESIUM_RTC_MODELVIEW';
+    }
+
     var techniqueParameters = {
         // Add matrices
         modelViewMatrix: {
-            semantic: 'MODELVIEW',
+            semantic: modelViewSemantic,
             type: WebGLConstants.FLOAT_MAT4
         },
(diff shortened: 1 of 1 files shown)


lib/processModelMaterialsCommon.js
```diff
@@ -208,13 +208,39 @@ function generateTechnique(gltf, khrMaterialsCommon, lightParameters, optimizeFo
...
 
-    var hasNormals = (lightingModel !== 'CONSTANT');
+
+    // Check if normals exist
+    var meshesHaveNormals = false;
+    var meshes = gltf.meshes;
+    for (var meshId in meshes) {
+        if (meshes.hasOwnProperty(meshId)) {
+            var mesh = meshes[meshId];
+            if (mesh.hasOwnProperty('primitives')) {
```
lilleyse: Use `defined` instead.

lib/processModelMaterialsCommon.js
```diff
@@ -208,13 +208,39 @@ function generateTechnique(gltf, khrMaterialsCommon, lightParameters, optimizeFo
...
+                for (var p = 0; p < mesh.primitives.length; p++) {
```
lilleyse: Factor out length into its own variable before the loop.

lib/processModelMaterialsCommon.js
```diff
@@ -208,13 +208,39 @@ function generateTechnique(gltf, khrMaterialsCommon, lightParameters, optimizeFo
...
+                    if (mesh.primitives[p].material === materialId) {
+                        if (defined(mesh.primitives[p].attributes.NORMAL)) {
+                            meshesHaveNormals = true;
+                            break;
```
lilleyse: Might be worth putting this whole section into a separate function and using `return` instead to break out of both loops.

shehzan10:
@lilleyse updated

PR #322, opened 2017-09-28
Make ForEach a proper static function

### LINKED ISSUE #310: Fix `ForEach` in 2.0 branch
https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/2.0/lib/ForEach.js#L12

`function ForEach() {}` should be `var ForEach = {}` instead.  `ForEach` doesn't need a constructor.

hpinkos:
Fixes #310

### DIFF
--- a/lib/ForEach.js
+++ b/lib/ForEach.js
@@ -3,13 +3,10 @@ var Cesium = require('cesium');
 
 var defined = Cesium.defined;
 
-module.exports = ForEach;
-
 /**
  * Contains traversal functions for processing elements of the glTF hierarchy.
- * @constructor
  */
-function ForEach() {}
+var ForEach = {};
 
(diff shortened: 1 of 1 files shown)


lilleyse:
Does `module.exports = ForEach` need to be at the bottom?

hpinkos:
No, it just has to be after `var ForEach = {}`.  I just thought organizationally it made more sense to put it at the bottom

lilleyse:
Maybe that's why we used `function ForEach() {}` instead of `var ForEach = {}`, so that the `module.exports` could go in the same place for all files. Is this change mainly for style purposes?

hpinkos:
`ForEach` is a static class, so it shouldn't have a constructor

lilleyse:
Ok, sounds good then.

PR #323, opened 2017-09-28
Remove default scene for gltf 2.0

hpinkos:
glTF 2.0 doesn't have scenes

Or if this is part of the 1.0 to 2.0 conversion, this function should be

```js
    if (defined(gltf.scenes) && !defined(gltf.scene)) {
        gltf.scene = gltf.scenes[0];
    }
```
instead.

### DIFF
--- a/lib/addDefaults.js
+++ b/lib/addDefaults.js
@@ -380,14 +380,8 @@ function addDefaultByteOffsets(gltf) {
 }
 
 function selectDefaultScene(gltf) {
-    var scenes = gltf.scenes;
-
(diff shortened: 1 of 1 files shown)


lilleyse:
Slight edit - should be

```js
    if (defined(gltf.scenes) && !defined(gltf.scene)) {
        gltf.scene = 0;
    }
```

By the way, glTF 2.0 still uses scenes.

hpinkos:
@lilleyse whoops, skimmed through the 2.0 spec to quickly.  This is ready now, thanks!

PR #324, opened 2017-09-28
Make gltf-pipeline and Cesium consistent

lilleyse:
Some changes so that running `build-cesium` produces the version actually used in Cesium. Once this is merged I'll update Cesium's gltf-pipeline copy since there are other formatting things fixed here.

@hpinkos do you want to look at this?


lilleyse:
Actually I need to make a couple more changes, some tests are failing. I'll update when ready.

lilleyse:
This ended up being longer than expected. These are the fixes:

* Some scalar material values where surrounded with `[]`. I think this was a remnant of an earlier version of the 2.0 spec. Along with this removed `makeTechniqueValuesArrays`.
* Textures were not using the correct glTF 2.0 form in `bakeAmbientOcclusion`
* Build for cesium was adding `'/*global define*/` which has since been removed in Ceisum
* Use of Cesium's internal `ShaderSource` functions was failing because it wasn't passing in a context, which was a recent Cesium change to detect webgl 2. (This ought to be fixed in the cleanup branch)
* After parsing glb's the code now converts the buffer.extras._pipeline.source from a Uint8Array to Buffer. The confusion is that the components bundled in Cesium need to use the Uint8Array version, but Node needs to use the Buffer version.
* Added compressed texture stuff in `updateVersion` which had only made it into Cesium and not here
* Removed `requireAccessorMinMax`. This function was unsafe to use in Cesium because it required glTFs to have proper byte alignment, and glTF 1.0 models often fail to be aligned. I'll open an issue to add this functionality back.
* Removed tests that were checking that correct alpha-blending techniques were being created for materials. We removed this from Cesium because it was error-prone and required examining if textures were translucent, which is very slow.

lilleyse:
Given that this PR has grown to many more areas in the code, @shehzan10 can you review?

I'm waiting for Travis to pass as well.

PR #336, opened 2017-11-22
Fix for default materials

lilleyse:
Fixes two separate bugs for when a model with pbr materials that also has a default material.

### DIFF
--- a/lib/processPbrMetallicRoughness.js
+++ b/lib/processPbrMetallicRoughness.js
@@ -42,12 +42,13 @@ function processPbrMetallicRoughness(gltf, options) {
         splitIncompatibleSkins(gltf);
 
         ForEach.material(gltf, function(material) {
-            if (defined(material.pbrMetallicRoughness)) {
-                var pbrMetallicRoughness = material.pbrMetallicRoughness;
+            var pbrMetallicRoughness = material.pbrMetallicRoughness;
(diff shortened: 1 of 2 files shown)


lib/processPbrMetallicRoughness.js
```diff
@@ -41,23 +41,15 @@ function processPbrMetallicRoughness(gltf, options) {
...
                 var pbrMetallicRoughness = material.pbrMetallicRoughness;
                 var technique = generateTechnique(gltf, material, options);
 
-                var newMaterial = {
-                    values : pbrMetallicRoughness,
-                    technique : technique
-                };
-                materials.push(newMaterial);
+                material.values = pbrMetallicRoughness;
+                material.technique = technique;
```
likangning93: Seems to work as advertised, but should this also be removing the `pbrMetallicRoughness` field?
Otherwise it ends up in the output despite (from what I can tell) not being needed, since the resulting gltf bundles techniques, programs, and shaders.

Is the intent to retain compatibility of the output with other engines? Or is this just for Cesium?

lilleyse: Oh that's just an oversight on part, will delete `pbrMetallicRoughness`

lilleyse:
Updated

likangning93:
Thanks @lilleyse!

PR #338, opened 2017-11-27
Catch Jimp.read errors and ignore it

shehzan10:
Doesn't affect how the pipeline works if the images is not loaded. But
saves the processing from crashing due to unhandled errors from jimp.


lilleyse:
There is an eslint error.

shehzan10:
@lilleyse Fixed.

PR #340, opened 2017-12-07
Fix semantic check for pbr

lilleyse:
Copied from the same area in `processModelMaterialsCommon`. This change must have been added there but forgot to be added to `processPbrMetallicRoughness`.

Discovered through a bug report on the forum: https://groups.google.com/forum/#!topic/cesium-dev/qMKvqBUGsoE


likangning93:
Confirmed to work with the data in the issue linked. Thanks @lilleyse!

PR #350, opened 2018-03-02
Fix a couple issues with supplied tangent processing.

emackey:
Not sure if this is the correct upstream branch to target with my PR.

This is just a direct copy of changes I made in AnalyticalGraphicsInc/cesium#6302.


lilleyse:
Thanks @emackey. Look good to me.

PR #351, opened 2018-03-09
Fix generating shaders for case where material is not used by a primitive

lilleyse:
Fixes https://groups.google.com/forum/#!topic/cesium-dev/xat4Uy4sabQ


lilleyse:
Added a test that fails on master and works here. We don't have tests for `processPbrMetallicRoughness` so I only included it for `processModelMaterialsCommon`.

shehzan10:
LGTM.

PR #355, opened 2018-03-19
Fix default value of alphaCutoff.

emackey:
This is a copy of https://github.com/AnalyticalGraphicsInc/cesium/pull/6346.


lilleyse:
Thanks @emackey.

PR #357, opened 2018-03-26
glTF BLEND mode didn't respect double-sided flag.

emackey:
This is a copy of https://github.com/AnalyticalGraphicsInc/cesium/pull/6371.


lilleyse:
Looks good.

PR #359, opened 2018-03-27
Update NPM modules

mramato:
No code changes needed except for some new eslint failures due to double line breaks and I had to fix one test suite which was using beforeAll instead of beforeEach.

@lilleyse this has the one bad failing test we just talked about, hopefully it's an easy fix.


specs/lib/removeUnusedVerticesSpec.js
```diff
@@ -1,592 +1,591 @@
-'use strict';
```
mramato: The line endings in this file were wrong, which is why changing updated every line in the file.  All I did was remove an extra line break, no need to review.

lilleyse:
Fixed the test. I'll merge once CI passes.

mramato:
Thanks!

emackey:
Will this get merged down into 2.0 and 2.0-cesium?

mramato:
> Will this get merged down into 2.0 and 2.0-cesium?

I don't see why not (though the changes are super trivial and shouldn't effect either at all)

PR #361, opened 2018-04-03
More careful splitting of glTF materials

emackey:
This is a copy of https://github.com/AnalyticalGraphicsInc/cesium/pull/6421.


lilleyse:
Since there are already no tests in `processPbrMetallicRoughness` this is good to merge.

PR #363, opened 2018-04-24
Add code to remove unused buffer data in mergeBuffers

FrankGalligan:
- mergeBuffers will only copy data from buffers if it is referenced in a
bufferView.

### DIFF
--- a/lib/mergeBuffers.js
+++ b/lib/mergeBuffers.js
@@ -25,15 +25,24 @@ function mergeBuffers(gltf) {
         var bufferViewIds = bufferViewsForBuffers[bufferId];
         var length = bufferViewIds.length;
         for (var i = 0; i < length; ++i) {
-            var bufferView = bufferViews[bufferViewIds[i]];
-            bufferView.byteOffset += lengthSoFar;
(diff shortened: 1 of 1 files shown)


lib/mergeBuffers.js
```diff
@@ -25,15 +25,24 @@ function mergeBuffers(gltf) {
...
             bufferView.buffer = 0;
-        }
 
-        var sourceBuffer = getBufferPadded(buffer.extras._pipeline.source);
-        var sourceBufferLength = sourceBuffer.length;
-        buffersToMerge.push(sourceBuffer);
-        lengthSoFar += sourceBufferLength;
+            buffersToMerge.push(sourceBufferViewData);
+            lengthSoFar += sourceBufferViewData.length;
+        }
```
lilleyse: Cool, I like the approach. It could even be simplified more now by just iterating over the buffer views directly. I don't think there's any need for the `getBufferViewsForBuffers` stuff now.

FrankGalligan: Done.

lilleyse:
Thanks looks good!

Issue #365, opened 2018-04-27
Error with animated draco


soadzoor:
Hi,

So today I played some time with the new draco feature. Everything seemed to work fine, expect for when I tried to compress animated gltfs:

Without draco:
![without_draco](https://user-images.githubusercontent.com/10392261/39370893-61d07c64-4a40-11e8-9998-5519798ead30.png)

With draco:
![with_draco](https://user-images.githubusercontent.com/10392261/39370902-66fb33f0-4a40-11e8-81d4-b0a3be27cf26.png)

The validator on [donmccurdy's viewer](https://gltf-viewer.donmccurdy.com/) says the following error: `Length of 0x004e4942 chunk is not aligned to 4-byte boundaries`

When I export it as a `gltf+bin` (instead `glb`), the error is not showing up, but the visual artifact is the same. As well, as in the babylon sandbox. When I remove the `-d` from the command, the exported file's animation works well, so I'm pretty sure draco has something to do with it.

The command I use:
`node bin/gltf-pipeline.js -i input/cube.gltf -s -d -b -o output/cube.glb`

My files:
[animation_test.zip](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/1955722/animation_test.zip)

FrankGalligan:
So currently encoding a mesh with morph targets with default settings could cause errors as you observed. I'm going to make a change to gltf-pipeline to fix these types of files. The compression won't be as good as I'm going to force sequential encoding which will preserve the order. 

Later on, we will make changes to properly handle animation and morph targets.

FrankGalligan:
@soadzoor I tested your files with https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/367 and it works for me.

lilleyse:
Fixed in #367 

> Length of 0x004e4942 chunk is not aligned to 4-byte boundaries

This error seems to have been fixed at some point in the meantime.

soadzoor:
Great, thanks guys for the quick fix!

PR #367, opened 2018-05-01
If primitive contains targets set sequential encoding

FrankGalligan:
Check if the primitive contians any morph targets. If it does, set sequential encoding to preserve the order of vertices.

This fixes https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/365


lilleyse:
Works for the model in #365 and [AnimatedMorphSphere](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedMorphSphere).

Thanks @FrankGalligan

PR #372, opened 2018-05-30
Replace original accessor properties with decompressed data properties

FrankGalligan:
This addresses https://github.com/KhronosGroup/glTF/pull/1343 for compressed data.

### DIFF
--- a/lib/compressDracoMeshes.js
+++ b/lib/compressDracoMeshes.js
@@ -251,7 +251,7 @@ function addCompressionExtensionToPrimitive(gltf, primitive, attributeToId, enco
         attributes: attributeToId
     };
 
-    gltf = replaceWithDecompressedPrimitive(gltf, primitive, encodedPointsCount, encodedFacesCount, primitive.extensions.KHR_draco_mesh_compression);
+    gltf = replaceWithDecompressedPrimitive(gltf, primitive, encodedPointsCount, encodedFacesCount);
 }
 
 function copyCompressedExtensionToPrimitive(primitive, compressedPrimitive) {
--- a/lib/compressDracoMeshes.js
+++ b/lib/compressDracoMeshes.js
@@ -184,16 +184,19 @@ function compressDracoMeshes(gltf, options) {
                 encoder.SetEncodingMethod(encoderModule.MESH_SEQUENTIAL_ENCODING);
             }
 
+            encoder.SetTrackEncodedProperties(true);
             var encodedLength = encoder.EncodeMeshToDracoBuffer(mesh, encodedDracoDataArray);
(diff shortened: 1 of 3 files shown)


lilleyse:
Does the encoder have information about how many faces and points are in the encoded geometry, or is there a plan to add that?

FrankGalligan:
Not currently, but we are going to look into adding it today. If we can add it, I can change this PR.

FrankGalligan:
@lilleyse We added the encoder api to get number of points and faces in draco npm 1.3.3. I updated the PR to use this new API.

lib/compressDracoMeshes.js
```diff
@@ -246,6 +250,8 @@ function addCompressionExtensionToPrimitive(gltf, primitive, attributeToId, enco
         bufferView: bufferViewId,
         attributes: attributeToId
     };
+
+    gltf = replaceWithDecompressedPrimitive(gltf, primitive, encodedPointsCount, encodedFacesCount, primitive.extensions.KHR_draco_mesh_compression);
```
lilleyse: Small thing, but instead of passing in both `primitive` and `primitive.extensions.KHR_draco_mesh_compression`, maybe just pass in `primitive` and get the extension in the function itself.

lib/replaceWithDecompressedPrimitive.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Object} gltf A javascript object containing a glTF asset.
+ * @param {Object} primitive A javascript object containing a glTF primitive.
+ * @param {Number} encodedPointsCount Number of points after the mesh is decompressed.
+ * @param {Number} encodedFacesCount Number of faces after the mesh is decompressed.
+ * @param {Object} dracoExtension A javascript object containing KHR_draco_mesh_compression.
+ * @returns {Object} The glTF asset with the decompressed primitive.
+ *
+ * @private
+ */
+function replaceWithDecompressedPrimitive(gltf, primitive, encodedPointsCount, encodedFacesCount, dracoExtension) {
```
lilleyse: Now that this function is simpler I think it's fine to move it into `compressDracoMeshes`.

lib/compressDracoMeshes.js
FrankGalligan: Done.

lib/replaceWithDecompressedPrimitive.js
FrankGalligan: I'm going to keep for now, as the next branch will add more code here.

lilleyse: Alright!

Issue #330, opened 2017-10-10
2.0 Roadmap


lilleyse:
**Background:**

The [2.0 branch](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/tree/2.0) is used by Cesium to upgrade 1.0 models to 2.0. The upgraded models use the [KHR_technique_webgl](https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_technique_webgl) extension, which is supported by Cesium but hardly any other engines. PBR is not supported in this branch.

There is another branch called [cleanup](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/tree/cleanup) which is a cleanup of glTF-pipeline that supports 1.0-2.0 conversion, conversions between glTF and glb, and other basic tasks. The other pipeline stages will eventually be moved over here once the core is solid.

The `2.0` branch will be renamed to `2.0-cesium` and the `cleanup` branch renamed to `2.0`.

One confusing aspect of the project right now is a subset of the files are used within Cesium for upgrading 1.0 models. Cesium includes gltf-pipeline which includes Cesium. Our early attempts at using browserify didn't work, so we instead added a gulp task called `build-cesium` which saves those gltf-pipeline files with AMD syntax to be packaged directly with Cesium. Additionally those files do not use the Node.js API or any third party packages (except Cesium). If anyone can think of an alternative to this approach that would be very helpful.

Another side effect of including gltf-pipeline in Cesium is we try to make those files as performant as possible. We try to avoid any large buffer copies, and because of that glTF 1.0 models that are upgraded to 2.0 are not always byte aligned correctly. This is best for Cesium but not best for gltf-pipeline, so there needs to be some compromise here.

**To do:**

* [ ] Bring the new `2.0` branch up to good enough quality so others can start working on it. It is almost there.
* [ ] `updateVersion` needs to fix accessors that have invalid byte alignment: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#data-alignment
* [ ] Support upgrading glTF 1.0 materials to PBR in `updateVersion`. This means removing shaders and techniques and making some best guesses for upgrading to PBR.
* [ ] A lot of models are using the old `KHR_materials_common` extension. There should be a stage to convert this to PBR.
* [ ] Transfer all stages over.
* [ ] Support exporting with specular-glossiness. This will require a metallic-roughness to specular-glossiness converter.
[...]

PR #387, opened 2018-06-27
Update technique render states and support for KHR_blend

ggetz:
Following #376, processes legacy technique render states, setting material `doubleSided` and `alphaMode` properties where neccisary, and moving blending functions to the [`EXT_blend`](https://github.com/KhronosGroup/glTF/pull/1302) extension. 

@lilleyse As far as this PR, is there anything else I have to do to address `material.alphaMode` (specifically for `MASK`, and `material.alphaCutoff`)?

Part of https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/330


lib/updateMaterialProperties.js
```diff
@@ -0,0 +1,86 @@
...
+                    };
+                }
+            }
+
+            // If CULL_FACE is not enabled, the material should be doubleSided
+            if (!isStateEnabled(renderStates, WebGLConstants.CULL_FACE)) {
+                materialProperties.doubleSided = true;
+            }
+
+            delete techniqueLegacy.state;
```
lilleyse: Should be `techniqueLegacy.states`

(Though I don't think it actually matters since it won't be included in the final `KHR_techniques_webgl` technique)

specs/lib/updateMaterialPropertiesSpec.js
```diff
@@ -0,0 +1,236 @@
...
+            extensionsRequired: []
+        };
+
+        var gltfWithUpdatedMaterials = updateMaterialProperties(gltf);
+        var material = gltfWithUpdatedMaterials.materials[0];
+        expect(material.doubleSided).toBe(true);
+
+        gltf = {
+            programs: {
+                program_0: {
```
lilleyse: To make this more condensed clone the previous gltf and add the new state.

This can probably be done to make all the tests shorter. Have a single glTF above the tests that is cloned/edited by each test.

specs/lib/updateMaterialPropertiesSpec.js
```diff
@@ -0,0 +1,236 @@
...
+                    }
+                }
+            },
+            materials: [
+                {
+                    technique: 'technique0'
+                }
+            ],
+            extensionsUsed: [],
+            extensionsRequired: []
```
lilleyse: Remove these two if they aren't used.

specs/lib/updateMaterialPropertiesSpec.js
```diff
@@ -0,0 +1,236 @@
...
+        };
+
+        var gltfWithBlendExtension = updateMaterialProperties(gltf);
+        expect(gltfWithBlendExtension.extensionsUsed.indexOf('EXT_blend')).toEqual(-1);
+        expect(gltfWithBlendExtension.extensionsRequired.indexOf('EXT_blend')).toEqual(-1);
+
+        var material = gltfWithBlendExtension.materials[0];
+        expect(material.alphaMode).toBeUndefined();
+        expect(material.extensions).toBeDefined();
```
lilleyse: At first I was confused why the extensions were defined, and then noticed `KHR_techniques_webgl`. Maybe just remove that.

lib/updateMaterialProperties.js
```diff
@@ -0,0 +1,86 @@
...
+
+            // If BLEND is enabled, the material should have alpha mode BLEND
+            // and save the blend functions to the EXT_blend extension
+            if (isStateEnabled(renderStates, WebGLConstants.BLEND)
+                    || defined(renderStates.blendEquationSeparate) || defined(renderStates.blendFuncSeparate)) {
+                materialProperties.alphaMode = 'BLEND';
+                if (defined(renderStates.functions)) {
+                    blendingForTechnique[techniqueIndex] = {
+                        blendEquation: renderStates.functions.blendEquationSeparate,
+                        blendFactors: renderStates.functions.blendFuncSeparate
```
lilleyse: It looks like gltf 1 has more blend factors than the draft extension.

```
ZERO
ONE
SRC_COLOR
ONE_MINUS_SRC_COLOR
SRC_ALPHA
ONE_MINUS_SRC_ALPHA
DST_ALPHA
ONE_MINUS_DST_ALPHA
DST_COLOR
ONE_MINUS_DST_COLOR
CONSTANT_COLOR
ONE_MINUS_CONSTANT_COLOR
CONSTANT_ALPHA
ONE_MINUS_CONSTANT_ALPHA
SRC_ALPHA_SATURATE
```
vs

```

ZERO               
ONE                
SRC_COLOR          
ONE_MINUS_SRC_COLOR
SRC_ALPHA          
ONE_MINUS_SRC_ALPHA
DST_ALPHA          
ONE_MINUS_DST_ALPHA
DST_COLOR          
ONE_MINUS_DST_COLOR
```

If one of the old ones are encountered maybe just switch to the default value? Or not add the extension?

lilleyse:
> @lilleyse As far as this PR, is there anything else I have to do to address material.alphaMode (specifically for MASK, and material.alphaCutoff)?

I think what you have here is good!

lib/updateMaterialProperties.js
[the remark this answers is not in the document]
ggetz: Actually they are required by the extension, so I should definitely provide defaults.

lib/updateMaterialProperties.js
ggetz: I'm going to keep this here, if only because it allows this stage to run independently of later stages.

ggetz:
Thanks @lilleyse, updated !

I bumped https://github.com/KhronosGroup/glTF/pull/1302 to ask about the prefix and if there will be any more updates.

specs/lib/moveTechniqueRenderStatesSpec.js
```diff
@@ -0,0 +1,219 @@
...
+    materials: [
+        {
+            technique: 'technique0'
+        }
+    ]
+};
+
+describe('updateMaterialProperties', function() {
+    it('sets material.doubleSided property if CULL_FACE is not enabled', function () {
+        var gltfWithUpdatedMaterials = moveTechniqueRenderStates(gltf);
```
lilleyse: `gltf` should be cloned first.

specs/lib/moveTechniqueRenderStatesSpec.js
```diff
@@ -0,0 +1,219 @@
...
+        var material = gltfWithUpdatedMaterials.materials[0];
+        expect(material.doubleSided).toBeUndefined();
+
+        var gltfDoubleSided = JSON.parse(JSON.stringify(gltf));
```
lilleyse: Maybe use Cesium's `clone` function instead. `clone(gltf, true)`.

ggetz: I was having issues with the depth of the cloning when it came to arrays (even when using the deep clone argument). This was much more reliable.

lilleyse: Is that a Cesium bug? I don't think that should be happening...

specs/lib/moveTechniqueRenderStatesSpec.js
lilleyse: This may also be causing intermittent failures when tests are run in different order, which I think jasmine does now?

specs/lib/moveTechniqueRenderStatesSpec.js
```diff
@@ -0,0 +1,219 @@
...
+        }
+    },
+    materials: [
+        {
+            technique: 'technique0'
+        }
+    ]
+};
+
+describe('updateMaterialProperties', function() {
```
lilleyse: Update the name to `moveTechniqueToRenderStates`.

specs/lib/moveTechniqueRenderStatesSpec.js
ggetz: The documentation isn't partially clear on what the behavior should be: https://cesiumjs.org/Cesium/Build/Documentation/clone.html

lilleyse: Alright, if this way works consistently then let's go with it.

ggetz:
@lilleyse Alright, updated! And the tests are now consistently passing.

lilleyse:
Cool - will merge once we get word on the extension name.

ggetz:
@lilleyse Extension name updated to `KHR_blend` and updated this PR accordingly.

lilleyse:
Thanks @ggetz!

PR #389, opened 2018-07-02
Require min/max only for accessors containing position attributes

### LINKED ISSUE #325: Add back requireAccessorMinMax
From https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/324

> Removed requireAccessorMinMax. This function was unsafe to use in Cesium because it required glTFs to have proper byte alignment, and glTF 1.0 models often fail to be aligned. I'll open an issue to add this functionality back.

For the 2.0 branch. Ideally no operations in gltf-pipeline should require byte alignment since we always have to worry about 1.0 models passing through, and would prefer not to copy the memory. The byte alignment condition was new to the 2.0 spec.

lilleyse:
Fixes #325

According to the spec: 

> POSITION accessor must have min and max properties defined.

And it is optional for other accessors.

Since `requireAccessorMinMax` is by far the slowest part of the 1.0 - 2.0 upgrade I think it makes sense to only compute the min/max when needed. This will help speed up engines using gltf-pipeline to upgrade models on the fly. In fact I had previously removed `requireAccessorMinMax` from the `cesium-2.0` branch because it was too slow.

Thankfully min/max are already present on most glTF 1.0 position accessors so `findAccessorMinMax` should not be called too often.

### DIFF
--- a/lib/updateVersion.js
+++ b/lib/updateVersion.js
@@ -803,7 +803,8 @@ function moveByteStrideToBufferView(gltf) {
 }
 
 function requireAccessorMinMax(gltf) {
-    ForEach.accessor(gltf, function(accessor) {
+    ForEach.accessorWithSemantic(gltf, 'POSITION', function(accessorId) {
+        var accessor = gltf.accessors[accessorId];
         if (!defined(accessor.min) || !defined(accessor.max)) {
             var minMax = findAccessorMinMax(gltf, accessor);
             accessor.min = minMax.min;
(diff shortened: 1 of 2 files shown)


lib/updateVersion.js
```diff
@@ -803,7 +803,8 @@ function moveByteStrideToBufferView(gltf) {
 }
 
 function requireAccessorMinMax(gltf) {
```
ggetz: Might want to just clarify the helper function name here by specifying position.

ggetz:
Thanks @lilleyse !

PR #390, opened 2018-07-03
Allow for early returns in ForEach functions

ggetz:
Restructure `ForEach` function to allow for early returns when defined values are returned from a handler function. This is often used in Cesium and I think the restructure is easier here than jumping through hoops there.

### DIFF
--- a/lib/ForEach.js
+++ b/lib/ForEach.js
@@ -20,12 +20,16 @@ function ForEach() {
  * Fallback for glTF 1.0
  * @private
  */
-ForEach.objectLegacy = function(objects, handler) {
+ForEach.objectLegacy = function (objects, handler) {
     if (defined(objects)) {
         for (var objectId in objects) {
             if (objects.hasOwnProperty(objectId)) {
                 var object = objects[objectId];
-                handler(object, objectId);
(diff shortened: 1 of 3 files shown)


lib/ForEach.js
```diff
@@ -20,12 +20,16 @@ function ForEach() {
...
+ForEach.objectLegacy = function (objects, handler) {
     if (defined(objects)) {
         for (var objectId in objects) {
             if (objects.hasOwnProperty(objectId)) {
                 var object = objects[objectId];
-                handler(object, objectId);
+                var value = handler(object, objectId);
+
+                if (defined(value)) {
+                    return value;
```
lilleyse: `ForEachSpec` should be updated to test early returns.

lilleyse: Are there any area in gltf-pipeline that can take advantage of early returns now? https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/2.0/lib/writeResources.js#L141-L152 comes to mind. If anything else comes up we can just fix as we notice them.

ggetz:
Thanks @lilleyse, updated!

lilleyse:
Looks good, thanks!

PR #393, opened 2018-07-13
Decode uri before reading it

lilleyse:
To fix reading uri's like `"Cesium%20Logo%20Flat%20Low.png"`

@likangning93 can you review?


likangning93:
Thanks @lilleyse!

PR #395, opened 2018-07-13
Avoid adding duplicate shader for KHR_techniques_webgl

lilleyse:
I ran into an issue where two programs used the same vertex shader, and that shader object would get added twice to `extras.KHR_techniques_webgl.shaders`. This caused problems when writing resources later on.

@ggetz could you review?


lilleyse:
I noticed it in a few places so I updated `addToArray` to check for duplicates (optionally).

ggetz:
Thanks for the update @lilleyse!

PR #396, opened 2018-07-16
Ensure gltf.asset.version is defined

ggetz:
Some gltf assets, such as some sample b3dms, have `gltf.asset` defined, but not `gltf.asset.version`, which was causing failures.


lilleyse:
Thanks!

PR #397, opened 2018-07-16
Added removeExtension helper

lilleyse:
Added a helper function `removeExtension` that removes an extension from the glTF. Added to `index.js` as well since there are a few other projects that could use this helper.

### DIFF
--- a/lib/removeExtension.js
+++ b/lib/removeExtension.js
@@ -1,0 +1,32 @@
+'use strict';
+var Cesium = require('cesium');
+var removeExtensionsUsed = require('./removeExtensionsUsed');
+
+var defined = Cesium.defined;
+
+module.exports = removeExtension;
+
+/**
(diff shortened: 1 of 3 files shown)


lib/removeExtension.js
```diff
@@ -0,0 +1,32 @@
...
+var removeExtensionsUsed = require('./removeExtensionsUsed');
+
+var defined = Cesium.defined;
+
+module.exports = removeExtension;
+
+/**
+ * Removes an extension from gltf.extensions, gltf.extensionsUsed, and gltf.extensionsRequired if it is present.
+ * <p>
+ * Does not remove the extension from other objects in the glTF.
```
lilleyse: I might actually add support for this quickly... I'll update when ready.

lilleyse:
For some reason this PR wasn't including the latest commits in the branch, maybe a side effect of the github outage? I opened https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/398 instead which is up to date and ready to look at.

Issue #392, opened 2018-07-12
Fresh clone on Ubuntu shows a bunch of CRLF -> LF changes


likangning93:
...which makes doing basic stuff like checkout out branches impossible. Reproduced on multiple systems with really barebones git configs... what happened?

If I commit the line ending changes to a throwaway branch and then switch to `2.0` I also get problems with a couple files in there.

likangning93:
Herm, seems like no longer able to reproduce outside my VM.
Gonna close this for now.

likangning93:
no no no there's still CRLF files in here

PR #399, opened 2018-07-17
manual CRLF -> LF

likangning93:
Part of #392


likangning93:
Test failures seem to be around binary gltf buffer stuff and how different line endings can throw that off... 
Seems like a sign that the existing tests are kind of fragile.

likangning93:
> remove shader line endings from glb tests

This seemed like the "simplest" fix here, and it works for me both on Linux and Windows.
Let me know if it's too weird though.

lilleyse:
This is fine for now. I just went through and made sure each file had a newline at the end. https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/399/commits/9a091f47290989935a82afb783394ac114fe0a06

lilleyse:
Travis failed due to a timeout on a compression test.

PR #404, opened 2018-07-25
Remove checkTransparency

lilleyse:
Removes image decoding and transparency checking. The `checkTransparency` option wasn't actually being used by the pipeline. `writeResources` is now synchronous.


lib/writeResources.js
```diff
@@ -45,64 +41,52 @@ function writeResources(gltf, options) {
...
+    removeUnusedElements(gltf);
+    mergeBuffers(gltf);
+    writeBuffer(gltf, gltf.buffers[0], 0, options);
+    return gltf;
 }
 
 function writeBuffer(gltf, buffer, i, options) {
     if (defined(options.bufferStorage)) {
         options.bufferStorage.buffer = buffer.extras._pipeline.source;
-        return Promise.resolve();
```
ggetz: Do we still want to early return here?

lilleyse: Ah yes, thanks for catching that. Fixed.

ggetz:
Thanks, looks good @lilleyse!

PR #406, opened 2018-07-26
Update attribute component types, and remove sampler name properties

ggetz:
* Remove sampler `name` property to be consistant with glTF validator.
* Update accessors referenced by JOINTS_0 and WEIGHTS_0 attributes to use correct component types.
> The number of joints that influence one vertex is limited to 4, so referenced accessors must have VEC4 type and following component formats:
>
> JOINTS_0: UNSIGNED_BYTE or UNSIGNED_SHORT
> WEIGHTS_0: FLOAT, or normalized UNSIGNED_BYTE, or normalized UNSIGNED_SHORT

### DIFF
--- a/lib/updateAccessorComponentTypes.js
+++ b/lib/updateAccessorComponentTypes.js
@@ -1,0 +1,67 @@
+'use strict';
+var Cesium = require('cesium');
+var addToArray = require('./addToArray');
+var ForEach = require('./ForEach');
+var readAccessorPacked = require('./readAccessorPacked');
+
+var ComponentDatatype = Cesium.ComponentDatatype;
+var WebGLConstants = Cesium.WebGLConstants;
+
+module.exports = updateAccessorComponentTypes;
+
+/**
+ * Update accessors referenced by JOINTS_0 and WEIGHTS_0 attributes to use correct component types.
+ *
+ * @param {Object} gltf A javascript object containing a glTF asset.
+ * @returns {Object} The glTF asset with compressed meshes.
+ *
+ * @private
+ */
+function updateAccessorComponentTypes(gltf) {
+    var componentType;
+    ForEach.accessorWithSemantic(gltf, 'JOINTS_0', function(accessorId) {
+        var accessor = gltf.accessors[accessorId];
+        componentType = accessor.componentType;
+        if (componentType === WebGLConstants.BYTE) {
+            convertType(gltf, accessor, componentType, ComponentDatatype.UNSIGNED_BYTE);
+        } else if (componentType !== WebGLConstants.UNSIGNED_BYTE
+                && componentType !== WebGLConstants.UNSIGNED_SHORT) {
+            convertType(gltf, accessor, componentType, ComponentDatatype.UNSIGNED_SHORT);
+        }
+    });
+    ForEach.accessorWithSemantic(gltf, 'WEIGHTS_0', function(accessorId) {
+        var accessor = gltf.accessors[accessorId];
+        componentType = accessor.componentType;
+        if (componentType === WebGLConstants.BYTE) {
+            convertType(gltf, accessor, componentType, ComponentDatatype.UNSIGNED_BYTE);
+        } else if (componentType === WebGLConstants.SHORT) {
+            convertType(gltf, accessor, componentType, ComponentDatatype.UNSIGNED_SHORT);
+        }
+    });
+
+    return gltf;
+}
+
+function convertType(gltf, accessor, componentType, updatedComponentType) {
+    var data = ComponentDatatype.createTypedArray(componentType, readAccessorPacked(gltf, accessor));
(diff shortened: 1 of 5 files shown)


lib/updateVersion.js
```diff
@@ -480,7 +481,7 @@ function objectsToArrays(gltf) {
     });
     ForEach.animation(gltf, function(animation) {
         var samplerMapping = {};
-        animation.samplers = objectToArray(animation.samplers, samplerMapping);
+        animation.samplers = objectToArray(animation.samplers, samplerMapping, 'samplers');
```
lilleyse: It would be cleaner to add a separate function called `removeAnimationSamplerNames` and called after `objectsToArrays`.

From the spec it looks look `gltf.samplers` can have names but `animation.samplers` cannot.

lib/updateAccessorComponentTypes.js
```diff
@@ -0,0 +1,67 @@
...
+        }
+    });
+
+    return gltf;
+}
+
+function convertType(gltf, accessor, componentType, updatedComponentType) {
+    var data = ComponentDatatype.createTypedArray(componentType, readAccessorPacked(gltf, accessor));
+    var typedArray = ComponentDatatype.createTypedArray(updatedComponentType, data);
+    var newBuffer = Buffer.from(typedArray.buffer);
```
lilleyse: `Buffer.from` will not work from when this file is bundled in Cesium. I wonder if doing `Uint8Array(typedArray.buffer)` will work instead and across different versions of Node.

lib/updateAccessorComponentTypes.js
```diff
@@ -0,0 +1,67 @@
...
+            convertType(gltf, accessor, componentType, ComponentDatatype.UNSIGNED_SHORT);
+        }
+    });
+
+    return gltf;
+}
+
+function convertType(gltf, accessor, componentType, updatedComponentType) {
+    var data = ComponentDatatype.createTypedArray(componentType, readAccessorPacked(gltf, accessor));
+    var typedArray = ComponentDatatype.createTypedArray(updatedComponentType, data);
```
lilleyse:
```json
var data = readAccessorPacked(gltf, accessor);
var typedArray = ComponentDatatype.createTypedArray(updatedComponentType, data);
```

Remove the first `ComponentDatatype.createTypedArray`, I don't think it is needed.

specs/lib/updateAccessorComponentTypeSpec.js
```diff
@@ -0,0 +1,177 @@
...
+                    buffer: 0,
+                    byteLength: 12
+                }
+            ],
+            buffers: [
+                {
+                    byteLength: source.byteLength,
+                    extras: {
+                        _pipeline: {
+                            source: source
```
lilleyse: We moved all the tests away from setting `_pipeline` directly. Use data uris and `readResources` instead. See [readAccessorPackedSpec](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/2.0/specs/lib/readAccessorPackedSpec.js) for an example.

lib/updateAccessorComponentTypes.js
```diff
@@ -0,0 +1,67 @@
...
+    var buffer = {
+        byteLength: newBufferLength,
+        extras: {
+            _pipeline: {
+                source: newBuffer
+            }
+        }
+    };
+
+    delete gltf.bufferViews[accessor.bufferView];
```
lilleyse: The buffer view will be removed automatically in `removeUnusedElements` which is called before the glTF is written. This line isn't needed.

lib/updateAccessorComponentTypes.js
lilleyse: This worked in Node 8 and Cesium, but failed in Node 6. A quick fix would be to convert `Uint8Array` buffer view data into a `Buffer` in `mergeBuffers`.

lilleyse:
Rigged simple is looking good - just check out https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/406#discussion_r206230263

lib/updateAccessorComponentTypes.js
ggetz: I ran this in Node 6.10 and it works. Likewise it works in CI with 6.14.

lilleyse: Using 6.14.3 I get:

> C:\Code\gltf-pipeline>node bin/gltf-pipeline.js -i test/RiggedSimple.gltf
> TypeError: "list" argument must be an Array of Buffers

ggetz: Upgraded and was able to duplicate. Should be fixed now.

ggetz:
Thanks @lilleyse, fixed, see https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/406#discussion_r206295982

lilleyse:
Thanks!

PR #407, opened 2018-07-26
Require animation sampler accessor min and max

ggetz:
To comply with
> Animation Sampler's input accessor must have min and max properties defined.

### DIFF
--- a/lib/updateVersion.js
+++ b/lib/updateVersion.js
@@ -815,6 +815,17 @@ function requirePositionAccessorMinMax(gltf) {
     });
 }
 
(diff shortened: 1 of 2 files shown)


lilleyse:
Looks great! Can you remind me which model failed validation so I confirm this fix?

ggetz:
`RiggedSimple` will work, but I just validated the output and I'm still getting the error. I'll bump when updated.

ggetz:
@lilleyse Forgot to iterate through the outer animations, updated.

lilleyse:
Ah... thanks!

PR #410, opened 2018-07-30
Remove empty nodes when upgrading version

ggetz:
Remove empty nodes when updating version, checks parents for deletion recursively, and removes deleted nodes from the scene list. I do not believe we have to check anywhere else to update after deleting, as if the empty nodes were referenced, it wouldn't be a valid glTF in the first place.

Can test this with `CesiumMilkTruck` 1.0 model.


lilleyse:
Nice looks great!

PR #414, opened 2018-08-09
Add default material to primitive

lilleyse:
Adds a default material to each primitive that does not specify a material.

Test model: [CUBE.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/2272530/CUBE.gltf.txt)

@ggetz can you review?

### DIFF
--- a/lib/addDefaults.js
+++ b/lib/addDefaults.js
@@ -1,5 +1,6 @@
 'use strict';
 var Cesium = require('cesium');
+var addToArray = require('./addToArray');
 var ForEach = require('./ForEach');
 var getAccessorByteStride = require('./getAccessorByteStride');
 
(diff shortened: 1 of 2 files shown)


lib/addDefaults.js
```diff
@@ -31,6 +32,12 @@ function addDefaults(gltf) {
     ForEach.mesh(gltf, function(mesh) {
         ForEach.meshPrimitive(mesh, function(primitive) {
             primitive.mode = defaultValue(primitive.mode, WebGLConstants.TRIANGLES);
+            if (!defined(primitive.material)) {
+                if (!defined(gltf.materials)) {
+                    gltf.materials = [];
+                }
+                primitive.material = addToArray(gltf.materials, {});
```
ggetz: Maybe give this a name, like "default"?That way it's not a completely empty object, and we'll be able to easily identify the default material added later.

lilleyse:
Updated.

PR #418, opened 2018-08-20
Fix error when updating component type for accessors

ggetz:
Additional fix after https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/406

The accessor's `byteOffset` needs to be set to 0 for the new buffer.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,11 @@
 Change Log
 ==========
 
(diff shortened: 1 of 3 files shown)


lilleyse:
Good catch. Code look good, just update `CHANGE.md`.

ggetz:
Thanks @lilleyse, updated!

PR #421, opened 2018-09-05
Created accessor bufferViews should be byte aligned

ggetz:
This is why `Cesium_Air.gltf` was not rendering in https://github.com/AnalyticalGraphicsInc/cesium/pull/6805

The bufferViews created when splitting accessors with different byte strides must be byte aligned.

### DIFF
--- a/lib/updateVersion.js
+++ b/lib/updateVersion.js
@@ -790,7 +790,8 @@ function moveByteStrideToBufferView(gltf) {
                 var accessorByteLength = accessor.count * accessorByteStride;
                 delete accessor.byteStride;
 
-                var nextAccessorByteStride = (i < accessorsLength - 1) ? computeAccessorByteStride(gltf, accessors[i + 1]) : undefined;
+                var hasNextAccessor = (i < accessorsLength - 1);
+                var nextAccessorByteStride = hasNextAccessor ? computeAccessorByteStride(gltf, accessors[i + 1]) : undefined;
                 if (accessorByteStride !== nextAccessorByteStride) {
                     var newBufferView = clone(bufferView, true);
                     if (bufferViewHasVertexAttributes[bufferViewId]) {
@@ -804,7 +805,8 @@ function moveByteStrideToBufferView(gltf) {
                         accessor.bufferView = newBufferViewId;
                         accessor.byteOffset = accessor.byteOffset - currentByteOffset;
                     }
-                    currentByteOffset = accessorByteOffset + accessorByteLength + accessorByteLength % 4;
+                    // Set current byte offset to next accessor's byte offset
+                    currentByteOffset = hasNextAccessor ? accessors[i + 1].byteOffset : undefined;
                     currentIndex = i + 1;
                 }
             }
--- a/lib/updateVersion.js
+++ b/lib/updateVersion.js
@@ -804,7 +804,7 @@ function moveByteStrideToBufferView(gltf) {
                         accessor.bufferView = newBufferViewId;
                         accessor.byteOffset = accessor.byteOffset - currentByteOffset;
                     }
-                    currentByteOffset = accessorByteOffset + accessorByteLength;
+                    currentByteOffset = accessorByteOffset + accessorByteLength + accessorByteLength % 4;
                     currentIndex = i + 1;
                 }
             }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 ### 2.0.1 - ????
 
 * Fixed a bug where the buffer `byteOffset` was not set properly when updating 1.0 accessor types to 2.0 allowed values. [#418](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/418)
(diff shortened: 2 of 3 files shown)


lilleyse:
What is the error in Cesium? I took a look at Cesium_Air after going through gltf-pipeline and it seems like Cesium should still render it despite the validation errors.

The model seems to meet this condition at least:
> The offset of an accessor into a bufferView (i.e., accessor.byteOffset) and the offset of an accessor into a buffer (i.e., accessor.byteOffset + bufferView.byteOffset) must be a multiple of the size of the accessor's component type.

Though doesn't follow

> For performance and compatibility reasons, each element of a vertex attribute must be aligned to 4-byte boundaries inside bufferView (i.e., accessor.byteOffset and bufferView.byteStride must be multiples of 4).

Which is not good, but Cesium should still render it.

The reason I bring this all up is because editing the byte offset is not enough, padding also needs to get added to the buffer which could be a longer fix.

ggetz:
[Sandcastle example](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/gltf-update-pipeline/Build/Apps/Sandcastle/index.html#c=ZVPbbtswDP0VIS9xAU92bu2WJsGydMAGrFjXDHsyMKg2HQuTJUOSc9mQfx8l5+K6gh8o8vCQPJS3TJMthx1oMicSdmQFhtcl/eV9QT/115WSlnEJuh+Sf4kkeLjM1Se1J1OSM2EgbLwGBKSWK/lVZjxlVukuoGCZ2hn0Wl1fnaoW2VLyklk4hRJ5vLlPZCLzWnpGkmrA8KPKQAS1FiEpgG8Ke3PuqBmDgrTccjBUQ6m2sBQiaIg8BsetlOGecH4edsW0RYvJEc21Kh9gowFM8G4wHNH4bjy+HXwIyXhM40k8uotvL5Xvr6QFsIzLzZXzkdmCWvWMbiZNMBhN2vCK27RAcNzyaSVEx1VUnbV8aco8ufRnxAenumHDGHqSdiGlOQrCXs/7U2NLudKloUWH8EeNGmuJ8OCsU+jaeC2h1/iAhF3NWZYFp3W4I1np9um2dXVe9J9ezFa03fC0fWthSvcGMNqq5E6t+ZtiHs7xZdXlE9+DWPO/rqPB8H0XxPYOtE6ZcIBhjOeKODbm8SJtM7fVLP0D2eezHI0uiDk6tRLZfrJ9SqPLt6wqE61ZWQl4YJZFfiITNftZcn2yfqNJN8Lm+N9NsCEauwZ6YW9m7EHAomnmIy8rpa2bPEBuC0iLZU30UmN3lqbGuLRZdE6aZXxLeDZPep3fO+mRVDBjMJLXwouV9BazCPGv0oTyr+b7FrRgBwcpBotvjZNSOovw+jbLKiVemG4x/gc)

Getting the following WebGL errors:
```
bucket.html:1 [.Offscreen-For-WebGL-0000008F71D9A340]GL ERROR :GL_INVALID_OPERATION : glVertexAttribPointer: offset not valid for type
45115:48:29.071 [.Offscreen-For-WebGL-0000008F71D9A340]GL ERROR :GL_INVALID_OPERATION : glDrawElements: attached to enabled attrib 0 : no buffer
```

The `byteOffset` was being set to 2, which was what was causing the error because of the later 

> For performance and compatibility reasons, each element of a vertex attribute must be aligned to 4-byte boundaries inside bufferView (i.e., accessor.byteOffset and bufferView.byteStride must be multiples of 4).

lib/updateVersion.js
```diff
@@ -804,7 +804,7 @@ function moveByteStrideToBufferView(gltf) {
                         accessor.bufferView = newBufferViewId;
                         accessor.byteOffset = accessor.byteOffset - currentByteOffset;
                     }
-                    currentByteOffset = accessorByteOffset + accessorByteLength;
+                    currentByteOffset = accessorByteOffset + accessorByteLength + accessorByteLength % 4;
```
lilleyse: @ggetz I stared at this for a really long time. The fix here is pretty close, but I think it needs to be

`currentByteOffset = (i < accessorsLength - 1) ? accessors[i + 1].byteOffset : undefined;`

Which is setting the current byte offset to next accessor's byte offset rather than the end of the current accessor.
Cesium_Air brought up a discrepancy where these two values are not always the same if there is padding between the current accessor and the next accessor.

This approach should work in Cesium without requiring the more involved changes of always aligning byte offsets and byte strides to 4 bytes, which I still think involves buffer copies of some sort.

ggetz: Got it, thanks @lilleyse! This means we don't have to pad the buffer correct? That's good news!

lilleyse: Yeah no padding required for this fix thankfully.

ggetz:
Thanks again @lilleyse, updated. I can confirm Cesium Air now loads in https://github.com/AnalyticalGraphicsInc/cesium/pull/6805 with this change.

lilleyse:
Nice!

PR #422, opened 2018-09-05
Fixed removePipelineExtras for use in browsers

ggetz:
`Object.keys` casts `undefined` to an empty object in Node, but fails in the browser. This adds an extra check for `undefined`.


lilleyse:
Thanks!

PR #423, opened 2018-09-05
Use Node 6 and 8 in Appveyor builds

lilleyse:
The `dependency-tree` dev dependency doesn't seem to work in Node 4 and is causing CI builds to fail. Updated `.appveyor.yml` to use Node 6 and 8 instead.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,8 +5,8 @@ clone_depth: 50
 
 environment:
   matrix:
-    - nodejs_version: "4"
     - nodejs_version: "6"
+    - nodejs_version: "8"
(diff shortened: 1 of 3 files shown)


ggetz:
There's a failing test in Node 8: https://ci.appveyor.com/project/AnalyticalGraphics/gltf-pipeline/build/732

lilleyse:
@ggetz The failing tests should be fixed. There were two tests that weren't setting up their buffers incorrectly which was leading to out of bound reads.

ggetz:
Looks good @lilleyse, thanks!

PR #430, opened 2018-10-30
Split techniques for primitives that use different attributes

lilleyse:
Take into account whether a primitive has normals or texture coordinates when deciding whether to split incompatible techniques/materials.

This targets the 1.0 branch.

### DIFF
--- a/specs/lib/processModelMaterialsCommonSpec.js
+++ b/specs/lib/processModelMaterialsCommonSpec.js
@@ -326,7 +326,7 @@ describe('processModelMaterialsCommon', function() {
         expect(techniqueMat3.parameters.joint.type).toEqual(WebGLConstants.FLOAT_MAT3);
     });
 
-    it('splits two materials with different attributes', function() {
+    it('splits a material used by meshes with different attributes', function() {
         var gltf = {
             accessors: {
                 positionAccessor_0: {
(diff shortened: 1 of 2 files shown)


likangning93:
@lilleyse works as described for input gltf that already has `KHR_materials_common`, but should this also address generally when a primitive doesn't have enough attributes for the specified material/technique? Models without kmc that run with `kmc` output also don't hit this code path.

lilleyse:
> should this also address generally when a primitive doesn't have enough attributes for the specified material/technique

I'm not totally sure what the spec would say about that, but it sounds like that model would be invalid.

> Models without kmc that run with kmc output also don't hit this code path.

I think by this point a valid glTF would have already been split already, so the same kmc material wouldn't be used by two primitives with different attributes.

specs/lib/processModelMaterialsCommonSpec.js
```diff
@@ -325,4 +325,84 @@ describe('processModelMaterialsCommon', function() {
         expect(techniqueVec4.parameters.joint.type).toEqual(WebGLConstants.FLOAT_VEC4);
         expect(techniqueMat3.parameters.joint.type).toEqual(WebGLConstants.FLOAT_MAT3);
     });
+
+    it('splits two materials with different attributes', function() {
```
likangning93: Small thing: I'm reading this spec title as "splitting two materials," like splitting two materials into four materials, instead of something like "splits a material used by meshes with different attributes."

lilleyse: Fixed. I used your wording.

likangning93:
Thanks @lilleyse!

PR #434, opened 2018-11-06
Upgrade to gulp4

mramato:
`gulp.task` was replaced with module exports, and sync tasks now need to at least return a resolved promise.

No other changes required.


lilleyse:
Thanks @mramato.

PR #439, opened 2018-11-29
Fix code samples in readme

lilleyse:


lilleyse:
@shehzan10 could you review this quickly?

shehzan10:
Changes look good and I verified that there are no other errors.

However, do we want to change `var` to `const`/`let`? `var` are currently eslint errors right, so maybe we should not use them in the readme either.

lilleyse:
@shehzan10 Updated to use const.

shehzan10:
Thanks @lilleyse.

PR #454, opened 2019-02-09
Quantization volume option

lilleyse:
Adds an option to `compressDracoMeshes` that defines the explicit quantization volume to use rather the automatically computed ones.


bagnell:
👍

PR #464, opened 2019-03-08
readAccessorPacked: Fix accessor reading with undefined byteOffsets

Squareys:
Hi @likangning93 !

First of all: thanks for sharing the code of gltf-pipeline, I am currently trying to some custom gltf processing and this is really helping me out.

On the way I found a small bug that caused accessors being read incorrectly if no byteOffset is set on either the bufferView, accessor or source. According to the spec ([accessor.byteOffset](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#accessorbyteoffset) & [bufferView.byteOffset](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#accessorbyteoffset)) these should default to 0.

This may not be an issue with the command line tool, I suspect it may fill in default values before processing. In this case, maybe I should be calling something before doing this kind of processing?

Cheers, Jonathan.

### DIFF
--- a/lib/readAccessorPacked.js
+++ b/lib/readAccessorPacked.js
@@ -33,7 +33,9 @@ function readAccessorPacked(gltf, accessor) {
 
     const bufferView = gltf.bufferViews[accessor.bufferView];
     const source = gltf.buffers[bufferView.buffer].extras._pipeline.source;
-    let byteOffset = accessor.byteOffset + bufferView.byteOffset + source.byteOffset;
+    let byteOffset = (accessor.byteOffset || 0)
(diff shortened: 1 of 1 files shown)


Squareys:
I suppose that would have been `addDefaults`. Feel free to close this pullrequest, if that is the case.

lilleyse:
Yeah any processing will go through `addDefaults` first if you are using `processGltf`, `processGlb`, `gltfToGlb` or `glbToGltf`, and that's why most of the internal functions are not exposed in the API. The best way to run your own processing functions is to use the `customStages` option for the functions above.

PR #465, opened 2019-03-10
Remove more unused elements

Squareys:
Hi everybody!

While writing a custom stage for gltf-pipeline, I had the need to remove unused materials, meshes and nodes.

I figured this might not always be desirable, which is why I added `--removeUnused material,node,accessor` to `gtlf-pipeline.js` to allow specification of which types of elements should be removed.
To keep backwards compatibility, this defaults to `accessor,buffer,bufferView`.

Since order of removal matters, I ensured that the order is always the same independent of the order of given element types.

Are you interested in this pullrequest? Then I will make sure to test this and clean it up, as this is currently just some code extracted from my custom stage.

Looking forward to your reply,
Jonathan.

### DIFF
--- a/lib/removeUnusedElements.js
+++ b/lib/removeUnusedElements.js
@@ -18,9 +18,12 @@ module.exports = removeUnusedElements;
  *
  * @private
  */
-function removeUnusedElements(gltf, elementTypes=['node', 'mesh', 'material', 'accessor', 'bufferView', 'buffer']) {
+function removeUnusedElements(gltf, elementTypes) {
+    if(elementTypes === undefined) {
+        elementTypes = ['node', 'mesh', 'material', 'accessor', 'bufferView', 'buffer'];
+    }
     ['mesh', 'node', 'material', 'accessor', 'bufferView', 'buffer'].forEach(function(type) {
-        if(elementTypes.includes(type)) {
+        if (elementTypes.includes(type)) {
             removeUnusedElementsByType(gltf, type);
         }
     });
@@ -46,7 +49,7 @@ function removeUnusedElementsByType(gltf, type) {
         const length = arrayOfObjects.length;
 
         for (let i = 0; i < length; ++i) {
-            if(!usedIds[i]) {
+            if (!usedIds[i]) {
                 Remove[type](gltf, i - removed);
                 removed++;
             }
@@ -171,11 +174,11 @@ Remove.mesh = function(gltf, meshId) {
     meshes.splice(meshId, 1);
 
     ForEach.node(gltf, function(n) {
-        if(defined(n.mesh)) {
-            if(n.mesh > meshId) {
+        if (defined(n.mesh)) {
+            if (n.mesh > meshId) {
                 --n.mesh;
-            } else if(n.mesh === meshId) {
-                /* Remove reference to deleted mesh */
+            } else if (n.mesh === meshId) {
+                // Remove reference to deleted mesh
                 delete n.mesh;
             }
         }
@@ -184,25 +187,39 @@ Remove.mesh = function(gltf, meshId) {
 
 Remove.node = function(gltf, nodeId) {
     const nodes = gltf.nodes;
-
-    /* Remove this node from parent */
-    const parentId = findParentNode(gltf, nodeId);
-    if(parentId >= 0) {
-        nodes[parentId].children = nodes[parentId].children.filter(x => x !== nodeId);
-    }
-
     nodes.splice(nodeId, 1);
 
-    /* Shift all node references */
-    nodes.forEach(other => {
-        if(defined(other.children)) {
-            other.children = other.children.map(x => x > nodeId ? x - 1 : x);
+    // Shift all node references
+    ForEach.skin(gltf, function(s) {
+        if (s.skeleton) {
+            s.skeleton -= s.skeleton > nodeId ? 1 : 0;
+        }
+
+        s.joints -= s.joints.map(function(x) { x > nodeId ? x - 1 : x });
+    });
+    ForEach.animation(gltf, function(animation) {
+        if(animation.target && animation.target.node) {
+            animation.target.node -= animation.target.node > nodeId ? 1 : 0;
         }
     });
-    ForEach.scene(gltf, (s, i) => {
+    ForEach.technique(gltf, function(technique) {
+        ForEach.techniqueUniform(technique, function(uniform) {
+            if (defined(uniform.node)) {
+                uniform.node -= uniform.node > nodeId ? 1 : 0;
+            }
+        });
+    });
+    ForEach.node(gltf, function(n) {
+        if (!n.children) return;
+
+        n.children = n.children
+            .filter(function(x) { x !== nodeId }) // Remove
+            .map(function(x) { x > nodeId ? x - 1 : x }); // Shift indices
+    });
+    ForEach.scene(gltf, function(s, i) {
         s.nodes = s.nodes
-            .filter(x => x !== nodeId) /* Remove */
-            .map(x => x > nodeId ? x - 1 : x); /* Shift indices */
+            .filter(function(x) { x !== nodeId }) // Remove
+            .map(function(x) { x > nodeId ? x - 1 : x }); // Shift indices
     });
 };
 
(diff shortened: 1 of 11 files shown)


lilleyse:
Hi @Squareys,

Yes we're interested in this. Just one thing, I think we should always default to removing all unused elements and have a `--keep-unused-elements` option to disable that behavior. I'm not sure if we need the level of granularity currently in the PR, at least not yet.

Then the question is what to do about accessors, bufferViews, and buffers that need to be cleaned up from internal stages like `compressDracoMeshes` when `--keep-unused-elements` is passed in. I think each stage needs to be more careful about cleaning up after itself. I could open a PR for just that.

lilleyse:
Could you also fill out a CLA so we can review this?

https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla

Squareys:
@lilleyse Done! Sorry I missed that :)

No need to review _yet_, I wanted to roughly probe interest for merging this kind of work first! I should do some cleanup in case there is.

Squareys:
> --keep-unused-elements

I can imagine a usecase where you'd want to keep "empty" nodes (as weapon sockets or snapping something to in games for example), but still want to remove meshes and buffers. 

Similarly, you may want to keep materials, e.g. some gltf for a configurator application that contains the mesh and different unreferenced materials to switch at runtime. But still want to optimize the rest of the scene 🤔

pjcozzi:
Thanks again for the contribution @Squareys, we received your CLA.

Squareys:
@lilleyse In case you missed it, I had answered with some concerns whether the granularity may be useful after all. Wanted to make sure we come to a decision before cleaning this up.

lilleyse:
@Squareys Those are good use cases but I'm not sure if it happens much in practice where the glTF has a mix of intentionally unused properties and unintentionally unused properties. I still think we should keep the API simple to start.

Squareys:
@lilleyse Internally, I kept the granularity to be able to provide backwards compatible behaviour of draco mesh compression, update version, split primitives and a couple of other pipeline stages. (Otherwise most test cases would break, as usually there are materials, meshes and other elements that are not referenced and would be removed.)

I added the `--keepUnusedObjects` which will keep unused nodes, materials and meshes, restoring the behaviour of the current version.

I will look into the CI failure ASAP (maybe later today), I think that's the linting.

lilleyse:
@Squareys Thanks for the update, all sounds good.

Squareys:
@lilleyse This is ready for review!

I fixed the code style issues and improved test coverage a bit, added to README and CHANGES and finally rewrote the git commit history.

Squareys:
Rebased onto latest master

lilleyse:
Thanks, I hope to review this soon.

Squareys:
Hey @lilleyse, I don't want to be obtrusive, just wanted to make sure you push this PR back into my responsibility in case there is any need to clean up more. If any of the changes are unclear, I'm happy to answer any questions :)

lilleyse:
Sorry for the delay, it's still on my radar though. I might have some time later in the week to start the review.

lib/removeUnusedElements.js
```diff
@@ -260,3 +324,81 @@ getListOfElementsIdsInUse.bufferView = function(gltf) {
...
+        return false;
+    }
+
+    /* Empty if no children or children are all empty nodes */
+    return !defined(node.children)
+        || node.children.filter(n => !nodeIsEmpty(gltf, gltf.nodes[n])).length === 0;
+}
+
+getListOfElementsIdsInUse.node = function(gltf) {
+    const usedNodeIds = {};
```
lilleyse: Also handle:

* `skin.skeleton` and `skin.joints` referencing nodes
* [`KHR_techniques_webgl`](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_techniques_webgl) referencing a node. Which would look like:

```
    ForEach.technique(gltf, function (technique) {
        ForEach.techniqueUniform(technique, function (uniform) {
            if (defined(uniform.node)) {
                usedNodeIds[uniform.node] = true;
            }
        });
    });
```

lib/removeUnusedElements.js
```diff
@@ -12,20 +12,28 @@ module.exports = removeUnusedElements;
  * This function currently only works for accessors, buffers, and bufferViews.
```
lilleyse: Some overall comments for this file:

* Avoid any features above es5 because certain files of gltf-pipeline including this one are bundled in CesiumJS which still supports IE11. It's an annoying limitation that isn't made very clear to new contributors, so sorry about that. `const` and `let` are fine though and will work once https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/471 is merged. A summary of the bundling situation is here: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/330.
* For style consistency, change `if(` to `if (` everywhere.
* Also style related, use `//` for single line comments rather than `/*  */`

lib/removeUnusedElements.js
```diff
@@ -158,6 +166,62 @@ Remove.bufferView = function(gltf, bufferViewId) {
...
+
+    /* Remove this node from parent */
+    const parentId = findParentNode(gltf, nodeId);
+    if(parentId >= 0) {
+        nodes[parentId].children = nodes[parentId].children.filter(x => x !== nodeId);
+    }
+
+    nodes.splice(nodeId, 1);
+
+    /* Shift all node references */
```
lilleyse: Also shift node references for skeletons, joints, animations, and `KHR_technique_webgl`.

lib/removeUnusedElements.js
```diff
@@ -158,6 +166,62 @@ Remove.bufferView = function(gltf, bufferViewId) {
...
+    });
+};
+
+Remove.material = function(gltf, materialId) {
+    const materials = gltf.materials;
+    materials.splice(materialId, 1);
+
+    /* Shift other material ids */
+    ForEach.mesh(gltf, (m, j) => {
+        m.primitives.forEach(p => {
```
lilleyse: Use `ForEach.meshPrimitive`

lib/removeUnusedElements.js
```diff
@@ -158,6 +166,62 @@ Remove.bufferView = function(gltf, bufferViewId) {
...
+    /* Remove this node from parent */
+    const parentId = findParentNode(gltf, nodeId);
+    if(parentId >= 0) {
+        nodes[parentId].children = nodes[parentId].children.filter(x => x !== nodeId);
+    }
+
+    nodes.splice(nodeId, 1);
+
+    /* Shift all node references */
+    nodes.forEach(other => {
```
lilleyse: Use `ForEach.node`.

lib/removeUnusedElements.js
```diff
@@ -158,6 +166,62 @@ Remove.bufferView = function(gltf, bufferViewId) {
...
+        if(defined(other.children)) {
+            other.children = other.children.map(x => x > nodeId ? x - 1 : x);
+        }
+    });
+    ForEach.scene(gltf, (s, i) => {
+        s.nodes = s.nodes
+            .filter(x => x !== nodeId) /* Remove */
+            .map(x => x > nodeId ? x - 1 : x); /* Shift indices */
```
lilleyse: Could this same filter/map code be used above to combine the remove-from-parent and shift-indices steps? Maybe something like:

```js
    const nodes = gltf.nodes;
    nodes.splice(nodeId, 1);

    ForEach.node(gltf, (n) => {
        n.children = n.children
            .filter(x => x !== nodeId) /* Remove */
            .map(x => x > nodeId ? x - 1 : x); /* Shift indices */
    });
```

Squareys:
Oh! Sorry, this time the delay was on my side...

Will get to this over the weekend, will comment some replies in case I have any questions there 👍 
Thanks for the review!

lib/removeUnusedElements.js
Squareys: Awesome! Was expecting something like this to pop up, thanks!

lib/removeUnusedElements.js
Squareys: > bundled in CesiumJS which still supports IE11

Ah! I was very curious about why you don't use some of the newer features. I suspect in the tests they are ok, cause they are not bundled, right?

lilleyse: Yeah, exactly.

Squareys: Alright, I changed:

* `if(` to `if (`
* `x => {}` to `function(x) {}`
* `/* */` to `//`
* default parameter of removeUnusedElements

I hope the two were all the ES6 functions I used (Went through [this resource](http://es6-features.org/) to find them)

Squareys:
@lilleyse Done! Once you're happy with the changes, I can rebase to auto squash in those `fixup!` commits.

Tell me if there is anything I missed or anything else you want me to change :)

lilleyse:
@Squareys I pushed some fixes and did a consistency/cleanup pass over the code. For the sake of not holding this PR up any longer I'm going to go ahead and merge.

Thanks again!

PR #466, opened 2019-03-14
Fixed ReadResource to handle absolute uris

tfili:
`readFile` function in `readResources.js` handled absolute paths, but not absolute URIs, even though it was process URIs. It would see the absolute URI as a relative path and append it to the `resourceDirectory` which is obviously wrong. We end up with something like `c:\path\to\resources\file:\\\c:\full\path\to\the\image.png`

In addition we
  - Actually look at and test the `secure` flag.
  - Add `fileURLToPath` and `pathToFileURL` functions so we still work with older versions of node

The new process goes like this
- `resourceDirectory` if specified is made absolute right off the bat.
- In `readFile` we check if `resourceDirectory` is defined.
  - If it is, we convert it to a `URL` and resolve the resource uri with it
  - If it is not, we try to make the `URL` from just the resource uri. This fails if its not an absolute URL or absolute file path. We catch this and make it a friendly error.
  - Convert the URL to an absolute path, since it is now resolved correctly.
- If `resourceDirectory` was defined we now check if the file is outside this directory if `secure` is `true`.
- Then we read the file.


tfili:
@lilleyse Do we still need `secure`. @mramato said we may be able to remove it.

tfili:
Also, Node 8 has a `URL` class, so I'm not sure whats going on with the errors.

lilleyse:
> @lilleyse Do we still need secure. @mramato said we may be able to remove it.

I don't think it was ever even hooked up properly. If we don't need it it's fine by me to remove. There was  an old branch to fix it that never went anywhere which we can delete now: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/commits/fixSecureResources.

lilleyse:
> Also, Node 8 has a URL class, so I'm not sure whats going on with the errors.

`FileUrl.js` needs to include it too like `const { domainToUnicode, URL } = require('url');`

lib/FileUrl.js
```diff
@@ -0,0 +1,123 @@
...
+
+function fileURLToPath(path) {
+    Check.defined('path', path);
+
+    if (typeof path === 'string') {
+        path = new URL(path);
+    }
+
+    if (path.protocol !== 'file:') {
+        throw new DeveloperError('Expected path.protocol to start with file:');
```
lilleyse: This seems more like a runtime error.

lib/FileUrl.js
```diff
@@ -0,0 +1,123 @@
...
+    if (letter < CHAR_LOWERCASE_A || letter > CHAR_LOWERCASE_Z ||   // a..z A..Z
+        (sep !== ':')) {
+        throw new DeveloperError('file URL must be absolute');
+    }
+    return pathname.slice(1);
+}
+
+function getPathFromURLPosix(url) {
+    if (url.hostname !== '') {
+        throw new DeveloperError('Invalid platform');
```
lilleyse: Same comment.

specs/lib/readResourcesSpec.js
```diff
@@ -115,14 +115,27 @@ describe('readResources', () => {
 
     it('rejects when loading resource outside of the resource directory when secure is true', async () => {
         const gltf = readGltf(boxTexturedSeparate2Path);
-        gltf.images[0].uri = '../cesium.png';
+        gltf.images[0].uri = '../box-textured-binary-separate/cesium.png'; // Outside resourceDirectory
```
lilleyse: Why was the relative uri changed? Does `'../cesium.png';` still trigger an error?

lib/FileUrl.js
```diff
@@ -0,0 +1,123 @@
...
+const percentRegEx = /%/g;
+const backslashRegEx = /\\/g;
+const newlineRegEx = /\n/g;
+const carriageReturnRegEx = /\r/g;
+const tabRegEx = /\t/g;
+
+// The following function is copied from Node.js implementation of url.pathToFileURL
+// https://github.com/nodejs/node/blob/7237eaa3353aacf284289c8b59b0a5e0fa5744bb/lib/internal/url.js#L1345-L1383
+// url.pathToFileURL was added in 10.12 but pkg uses 10.4.1, and does not include this function.
+// Remove this function when pkg uses 10 LTS and includes this function.
```
lilleyse: Maybe the comment should be less about PKG and more about removing this once gltf-pipeline sets Node 10 to its minimum version.

specs/lib/readResourcesSpec.js
tfili: It was because the image didn't exist, so it would trigger a different error, as this test wasn't even testing the `secure` flag. It was throwing an error for not specifying a `resourceDirectory`.

I'm removing `secure` so this test will go away anyway.

lib/FileUrl.js
tfili: Copied from Swayze. Yep. Good catch.

tfili:
@lilleyse I think this is ready.

lilleyse:
Thanks @tfili

PR #468, opened 2019-03-21
Separate resources no longer hit Node buffer size limit

lilleyse:
Same idea as https://github.com/AnalyticalGraphicsInc/obj2gltf/pull/185.

Before merging buffers check to see if the merged buffer would exceed the Node buffer size limit (~2GB). If it does, save the buffers as separate .bin files instead of merging them into a single .bin file.

### DIFF
--- a/lib/mergeBuffers.js
+++ b/lib/mergeBuffers.js
@@ -1,4 +1,5 @@
 'use strict';
+const BUFFER_MAX_BYTE_LENGTH = require('buffer').constants.MAX_LENGTH;
 const Cesium = require('cesium');
 const ForEach = require('./ForEach');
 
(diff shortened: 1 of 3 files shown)


lib/mergeBuffers.js
```diff
@@ -23,12 +24,29 @@ function mergeBuffers(gltf, defaultName) {
...
         const buffer = gltf.buffers[bufferView.buffer];
         let mergedName = buffer.extras._pipeline.mergedBufferName;
         mergedName = defined(mergedName) ? baseBufferName + '-' + mergedName : baseBufferName;
+
+        if (splitBuffers) {
+            if (!defined(mergedNameCount[mergedName])) {
+                mergedNameCount[mergedName] = 0;
+            }
+            mergedName += '-' + mergedNameCount[mergedName]++;
+        }
```
likangning93: Looks like this block is ensuring that each buffer gets its own entry in `buffersToMerge` instead of getting lumped with other buffers.

I think `Buffer.concat` as used below will return a buffer referencing a separate chunk of memory even when called on an array of a single buffer. If we're worrying about fairly large buffers (such that concatenating would blow the buffer limit), could the `concat` below only happen when there's more than one buffer?

lilleyse: That's a good idea. Fixed.

PR #474, opened 2019-04-10
Fix: Don't duplicate buffer views that are referenced multiple times.

### LINKED ISSUE #416: Duplicates image data for images that originally share the same buffer view
Really any objects that share the same resources (shared buffer view, shared data uri) will end up duplicating that resource when the embedded glTF is written out.

Test model: [box-duplicate.gltf.txt](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/files/2288961/box-duplicate.gltf.txt)

Found while testing https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/302

freezy:
This PR adds a reference to each processed view buffer and doesn't re-add it when matched. It avoids resources being duplicated. This PR fixes #416.

The map is stored in the root's pipeline extras. It's an object, since I wasn't sure if using `Map` wouldn't introduce any issues with old browsers.

Note that this doesn't apply to buffer views who share the same segment in a buffer, those would still be duplicated.

### DIFF
--- a/lib/writeResources.js
+++ b/lib/writeResources.js
@@ -43,6 +43,9 @@ function writeResources(gltf, options) {
     options.separateShaders = defaultValue(options.separateShaders, false);
     options.dataUris = defaultValue(options.dataUris, false);
 
+    // remember which of the original buffer views have been written, so we can re-use them.
+    gltf.extras._pipeline.bufferViewMap = {};
+
     ForEach.image(gltf, function(image, i) {
         writeImage(gltf, image, i, options);
         ForEach.compressedImage(image, function(compressedImage) {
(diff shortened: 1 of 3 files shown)


lilleyse:
Thanks for this fix @freezy, I'll try to review later this week.

Can you send over a CLA if you haven't already?

https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla

freezy:
Oh, sorry I didn't know there was a CLA to sign. In this case I prefer maintaining my own fork.

lilleyse:
@freezy I opened https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/483 to fix the problem.

Issue #488, opened 2019-07-10
uniform variables error after updating model from gltf v1.0 to gltf v2.0


liukangkanglkk99:
I am trying to update model from gltf v1.0 to gltf v2.0.
The following may occur in a gltf file: 
`"techniques":{
 "a":{
     "uniforms":{
         "u_tex":"tex",
         "u_tex2":"bakingTex"
     }
 },
 "b":{
     "uniforms":{
         "u_tex":"bakingTex"
     }
 }
}
`
But in the `function moveTechniquesToExtension(gltf)` of moveTechniquesToExtension.js, 
the mappedUniforms variable is finally recorded as bakingTex -> u_tex and bakingTex -> u_tex2 information is missing.
This leads to errors in later use of uniform u_tex2.

lilleyse:
@liukangkanglkk99 Thanks for all the details, it should be quick for us to track this down. If you are able to share the glTF that would be helpful too.

lilleyse:
@liukangkanglkk99 I opened a fix here: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/490.

Would you mind trying it out with your gltf?

liukangkanglkk99:
With pleasure. I tested this and it does work.

PR #490, opened 2019-07-14
Fix techniques with same uniform name but different parameter

lilleyse:
Fixes https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/488

Had to map uniforms based on technique so that different techniques' uniforms would not override each other. Also fixed an issue where a program would get duplicated if two techniques referenced it.

### DIFF
--- a/lib/moveTechniquesToExtension.js
+++ b/lib/moveTechniquesToExtension.js
@@ -68,8 +68,6 @@ function moveTechniquesToExtension(gltf) {
             });
 
             if (!defined(seenPrograms[techniqueLegacy.program])) {
-                seenPrograms[techniqueLegacy.program] = true;
-
                 const programLegacy = gltf.programs[techniqueLegacy.program];
 
                 const program = {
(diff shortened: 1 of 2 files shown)


liukangkanglkk99:
Nice going! Thanks @lilleyse, I tested this and it does work.

lib/moveTechniquesToExtension.js
```diff
@@ -60,27 +61,35 @@ function moveTechniquesToExtension(gltf) {
...
             });
 
-            const programLegacy = gltf.programs[techniqueLegacy.program];
-            const program = {
-                name: programLegacy.name,
-                fragmentShader: undefined,
-                vertexShader: undefined,
-                glExtensions: glExtensions
-            };
+            if (!defined(seenPrograms[techniqueLegacy.program])) {
```
likangning93: Does this need to set `techniqueLegacy.program` in an `else` block?

lilleyse: Thanks, fixed!

lilleyse:
@likangning93 ready for another look

likangning93:
:+1:

PR #491, opened 2019-07-29
Add support for the basis compressed texture file extension

JordyvanDortmont:
When converting a `.gltf` file to a `.glb`, the `.basis` compressed texture file extension is not accepted as a valid file header. 

I added the file header, because soon basis will be supported in a GLTF (https://github.com/KhronosGroup/glTF/pull/1612).

### DIFF
--- a/lib/getImageExtension.js
+++ b/lib/getImageExtension.js
@@ -33,6 +33,8 @@ function getImageExtension(data) {
     } else if (webpHeaderRIFFChars.equals(Buffer.from([0x52, 0x49, 0x46, 0x46])) && webpHeaderWEBPChars.equals(Buffer.from([0x57, 0x45, 0x42, 0x50]))) {
         // See https://developers.google.com/speed/webp/docs/riff_container#webp_file_header
         return '.webp';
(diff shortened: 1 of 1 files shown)


lilleyse:
Based on the discussion in https://github.com/KhronosGroup/glTF/pull/1612 it seems like the extension will require basis images to be contained in a KTX2 file (`.ktx2` extension) not `.basis`.

Aside from that it would be good to add unit tests, one modifying `gets image extension from buffer` in `getImageExtensionSpec.js` and add two new ones in `processGltfSpec.js` mirroring the WebP tests. This will require adding a sample glTF and I think it's best to wait until the extension is finalized.

JordyvanDortmont:
Thank you for the feedback. This was also confirmed here: https://github.com/mrdoob/three.js/issues/17122#issuecomment-516199339.

I'll close the PR for now. Although there are, or were, quite some concerns around not directly including `.basis` files. We'll keep including them directly and if you'd like me to reopen and implement the units tests and add a sample glTF, I'll gladly do so.

PR #496, opened 2019-09-16
fix possible infinite loop in writeResources

likangning93:
Having a `gltf` with a plurality of images with the same `name` would cause this to infinite-loop.


lilleyse:
Good catch, thanks @likangning93

PR #498, opened 2019-09-19
Update build-cesium task for ES6

mramato:
This changes the Cesium build to output ES6 modules instead of AMD.  This is needed for Cesium's move to ES6.

@lilleyse are you the best person to review this?

We can hold off on merging until the Cesium PR is ready, but I wanted to get this opened  so I could reference it elsewhere.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -185,28 +185,22 @@ function amdify(source, subDependencyMapping) {
     // amdify source
     // indent
     outputSource = outputSource.replace(/\n/g, '\n    ');
+    outputSource = outputSource.replace(/'use strict';/g, '');
+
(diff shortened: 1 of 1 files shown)


lilleyse:
@mramato Should everything after the imports be unindented now? Besides that looks good.

unindent

mramato:
That's actually pretty tricky to do.  We don't actually touch any of the indented code, we just put back the string we get out of the source file. I might be able to do something about it, I'll see.

mramato:
Cool, thanks.  I stand corrected.

mramato:
@lilleyse this can be merged now that Cesium has merged ES6 support.

PR #516, opened 2020-01-31
Adds removal for unused textures, images, samplers in removeUnusedElements

sanjeetsuhag:


lilleyse:
@sanjeetsuhag the code looks good, just add some tests, probably here: https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/specs/lib/removeUnusedElementsSpec.js#L377-L378

sanjeetsuhag:
@lilleyse Added tests.

specs/lib/removeUnusedElementsSpec.js
```diff
@@ -391,6 +416,18 @@ describe('removeUnusedElements', () => {
...
+        expect(gltf.textures.length).toBe(1);
+    });
+
+    it('correctly removes/keeps samplers', () => {
+        expect(gltf.samplers.length).toBe(1);
+    });
+
+    it('correctly removes/keeps images', () => {
+        expect(gltf.samplers.length).toBe(1);
+    });
```
lilleyse: These tests could be improved.

* Incorporate these tests into https://github.com/AnalyticalGraphicsInc/gltf-pipeline/blob/master/specs/lib/removeUnusedElementsSpec.js#L377-L378 rather than being separate tests if possible
* `correctly removes/keeps textures` does not check for unused texture removal. Add two textures to the json, one used and one unused, and then check that the length is 1.
* `correctly removes/keeps images` is checking samplers instead of images. Like above, add an unused image to the json.

lilleyse:
Thanks @sanjeetsuhag!

Issue #532, opened 2020-03-27
Processing BrainStem.gltf removes nodes from scene.


prideout:
I tried the following command to compress the Khronos BrainStem model:

```
gltf-pipeline -i BrainStemp.gltf -o DracoBrainStem.gltf -d -s
```

The resulting asset has 21 nodes (just like the original) but the hierarchy has changed such that the 3rd node is orphaned (does not belong to the scene).

Is this intentional?  The 3rd node seems like it should belong to the scene since the first skinning joint points to the 4th node, which is a child of the 3rd node.

lilleyse:
@prideout I opened https://github.com/CesiumGS/gltf-pipeline/issues/532 to fix this. In the meantime you can pass in `--keep-unused-elements` to prevent gltf-pipeline from removing what it thinks are unused elements.

I may also open a PR to disable the `removeUnusedElements` stage by default since it has subtle bugs like this and not great forward compatibility for extensions.

prideout:
Nice, thanks!

PR #533, opened 2020-03-27
Don't delete nodes that have used descendants

lilleyse:
Fixes https://github.com/CesiumGS/gltf-pipeline/issues/532

Don't delete a node if it has a descendant that's used for skinning or animation. In the case of #532 a parent of a node in the skin's joints list was being removed. I confirmed the test passes in this branch but not in master.


likangning93:
:+1:

PR #538, opened 2020-04-15
Update to new Draco encode settings

FrankGalligan:


FrankGalligan:
These default values match the new default quantization settings in
version 1.3.6. The only difference is default quantization settings for
normals is 8. This will be the new default in the next version of Draco.
    
See release notes for more information:
https://github.com/google/draco#version-136-release

lilleyse:
Thanks @FrankGalligan!

PR #539, opened 2020-04-15
Fix min max for draco files

FrankGalligan:
This helps https://github.com/KhronosGroup/glTF-Sample-Models/pull/236
with files created from gltf-pipeline.


lilleyse:
Is it possible to compute the min max knowing the accessor min/max and the quantization bits? Or maybe it's more involved. Main reason I ask is the extra decode could hurt performance.

This branch:
```
(base) Seans-MacBook-Pro:gltf-pipeline slilley$ node bin/gltf-pipeline.js -i ~/Code/glTF-Sample-Models/2.0/BrainStem/glTF/BrainStem.gltf -o ~/Desktop/brainstem.gltf  -d
Total: 1272.551ms
```

Master:
```
(base) Seans-MacBook-Pro:gltf-pipeline slilley$ node bin/gltf-pipeline.js -i ~/Code/glTF-Sample-Models/2.0/BrainStem/glTF/BrainStem.gltf -o ~/Desktop/brainstem.gltf  -d
Total: 838.950ms
```

FrankGalligan:
So I tried some things, but there can always be some inconsistencies. Unless we adjust max and min by about 0.001. I think this might be fine. What do you think?

Also there is no guarantee that for some models 0.001 will be enough. It is for all of the gltf sample models.

FrankGalligan:
I think we will need to add getEncodedMinMax() to the encoder.

FrankGalligan:
I changed the code to not do the the decode. It works on all of the glTF sample models.

lilleyse:
@FrankGalligan The latest changes look good. The 0.001 tolerance should be ok. Is the main goal to run `--draco.uncompressedFallback` to get the exact values for https://github.com/KhronosGroup/glTF-Sample-Models/pull/236 and provide a reasonable backup for non `--draco.uncompressedFallback`?

FrankGalligan:
Yes if you run `--draco.uncompressedFallback` you will get exact. Otherwise you will get the tolerance based on the `accessor.min` and `accessor.max`.

In the future I think we can remove the tolerance. Filed an issue: https://github.com/google/draco/issues/599

lilleyse:
Cool, we'll follow https://github.com/google/draco/issues/599. In the meantime I think this change is fine.

PR #540, opened 2020-04-20
Fix quantized minmax for for some attributes.

FrankGalligan:
The minmax values for TANGENT, JOINTS_0, and WEIGHTS_0, could have been off.


lilleyse:
Thanks @FrankGalligan!

PR #552, opened 2020-07-17
Update docs (fixed file and folder path)

champ3oy:
Added '/' to indicate path and folders


lilleyse:
@champ3oy thanks for the suggestions, but this might be confusing because the directories now look like top level paths in Unix

champ3oy:
Okay, that's also right

PR #553, opened 2020-07-22
Fix CesiumJS build

lilleyse:
Fixes gltf-pipeline bundled in CesiumJS

* Add `.js` to the end of each import
* Changed formatting of imports in `forEachTextureMaterial` so that it's compatible with the `build-cesium` gulp task


YoussefV:
perfect. that fixes it

PR #562, opened 2020-11-11
Load draco asynchronously

FrankGalligan:
@lilleyse This PR works for me using Draco 1.3.6 (the current version)


lilleyse:
Based on discussion in https://github.com/CesiumGS/gltf-pipeline/pull/561, this PR actually doesn't work in Draco 1.3.6 and only 1.4.0

lilleyse:
Merging now that Draco 1.4.0 is released

PR #563, opened 2020-11-13
Lock draco to 1.3.6 temporarily

lilleyse:
Part of https://github.com/CesiumGS/gltf-pipeline/pull/561

gltf-pipeline initializes Draco modules in a way that will break once Draco 1.4.0 is released. This PR temporarily locks Draco to 1.3.6 so that projects that use gltf-pipeline can upgrade to gltf-pipeline 3.0.1 and not automatically pull Draco 1.4.0.

One annoyance is that a proposed fix https://github.com/CesiumGS/gltf-pipeline/pull/562 only works in Draco 1.4.0, not 1.3.6, for unknown reasons, so it complicates the order of events here.

The steps are:

* [ ] Merge this PR
* [ ] Publish gltf-pipeline 3.0.1
* [ ] Wait for Draco 1.4.0 release
* [ ] Merge https://github.com/CesiumGS/gltf-pipeline/pull/562 and add back caret symbol to `package.json`
* [ ] Publish gltf-pipeline 3.0.2

Projects will need to upgrade to gltf-pipeline 3.0.1 or 3.0.2 once they are released.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### 3.0.1 - 2020-11-14
+
+* Locked Draco npm dependency to version 1.3.6 to avoid module initialization errors in 1.4.0. [#563](https://github.com/CesiumGS/gltf-pipeline/pull/563)
+
 ### 3.0.0 - 2020-07-22
 
 * Breaking changes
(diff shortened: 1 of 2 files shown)


lilleyse:
Spoke to @mramato offline about this approach, he is ok with the plan.

PR #578, opened 2021-04-12
Delete uri property when parsing binary_glTF buffer (glTF 1.0)

lilleyse:
This PR deletes the uri property of the `binary_glTF` buffer, which is meant to be ignored because the binary chunk should be used instead. Without this fix the glTF 1.0 to 2.0 update step may produce an invalid buffer.

From the [KHR_binary_glTF](https://github.com/KhronosGroup/glTF/tree/master/extensions/1.0/Khronos/KHR_binary_glTF) glTF 1.0 extension:

> ...a runtime can ignore the `uri` property since the buffer refers to the Binary glTF `body` section, not an external resource. When a runtime encounters this buffer, it should use the Binary glTF `body` as the buffer.

Here's what a glTF like this looks like. A fake `uri` is used because [`buffer.schema.json`](https://github.com/KhronosGroup/glTF/blob/de6f8b5327959a61b77f280a95bf01b0c2681a6d/specification/1.0/schema/buffer.schema.json#L32) requires that `uri` exists. The `uri` would be deleted after this change.

```
  "buffers": {
    "binary_glTF": {
      "type": "arraybuffer",
      "byteLength": 13790,
      "uri": "data:,"
    }
  },
```

### DIFF
--- a/lib/parseGlb.js
+++ b/lib/parseGlb.js
@@ -77,6 +77,7 @@ function parseGlbVersion1(glb, header) {
         const binaryGltfBuffer = defaultValue(buffers.binary_glTF, buffers.KHR_binary_glTF);
         if (defined(binaryGltfBuffer)) {
             binaryGltfBuffer.extras._pipeline.source = binaryBuffer;
+            delete binaryGltfBuffer.uri;
         }
     }
     // Remove the KHR_binary_glTF extension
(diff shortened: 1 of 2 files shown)


ptrgags:
@lilleyse what would be the best way to test this? and is there a reason why there aren't any unit tests in this PR?

lilleyse:
@ptrgags thanks for the reminder, I meant to write a test but forgot

lilleyse:
@ptrgags ready

ptrgags:
Looks good, thanks @lilleyse!

PR #580, opened 2021-04-21
Fix name of spec gloss extension in addDefaults

lilleyse:
The name was `pbrSpecularGlossiness` and has been fixed to be `KHR_materials_pbrSpecularGlossiness`


ptrgags:
Works as advertised, thanks @lilleyse!

PR #590, opened 2021-06-15
rename hasExtension -> usesExtension to avoid confusion in CesiumJS

ptrgags:
In CesiumJS, we want to have a `hasExtension` function to check for extensions in 3D Tiles as well as glTF. This could get confused with the gltf-pipeline version, so we are renaming this.

This is based on this PR comment in cesium: https://github.com/CesiumGS/cesium/pull/9595#discussion_r651400226

@sanjeetsuhag @lilleyse

### DIFF
--- a/specs/lib/hasExtensionSpec.js
+++ b/specs/lib/hasExtensionSpec.js
@@ -2,7 +2,7 @@
 const usesExtension = require('../../lib/usesExtension');
 
 describe('usesExtension', () => {
-    it('has extension', () => {
+    it('uses extension', () => {
         const gltf = {
             extensionsUsed: [
                 'extension1',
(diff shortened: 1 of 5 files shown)


specs/lib/hasExtensionSpec.js
```diff
@@ -1,19 +1,19 @@
 'use strict';
-const hasExtension = require('../../lib/hasExtension');
+const usesExtension = require('../../lib/usesExtension');
 
-describe('hasExtension', () => {
+describe('usesExtension', () => {
     it('has extension', () => {
```
sanjeetsuhag, line 5:
```suggestion
    it('uses extension', () => {
```

sanjeetsuhag:
I made a minor change, but otherwise changes look good. Thanks @ptrgags

PR #593, opened 2021-06-21
Handle glTFs with EXT_meshopt_compression

lilleyse:
glTFs with [`EXT_meshopt_compression`](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/EXT_meshopt_compression) are now handled correctly when going through `gltf-pipeline`. The main fix was to support buffers that don't have a uri.

This PR also fixes a bug where a malformed glb would be created when using draco compression with the uncompressed fallback with embedded resources. `forceMergeBuffers` forces glb's to merge all buffers into one. The same problem showed up for meshopt models with the `fallback: true` option.

There was also a minor issue where the name of the merged buffer was the name of the last buffer in `gltf.buffers` rather than the first. The first buffer usually has the main contents, especially when loaded from a glb.


IanLilleyT:
We reviewed offline and added changes in https://github.com/CesiumGS/gltf-pipeline/pull/593/commits/e37f1ba9b76016834911aa42599dbcce7cad7087 and https://github.com/CesiumGS/gltf-pipeline/pull/593/commits/212ef55766c00c4b6c59b6f7ff8df6ae5a888a80

PR #600, opened 2021-08-03
Update build-cesium task to account for prettier formatting

lilleyse:
Updates the `build-cesium` task to account for gltf-pipeline being formatted with prettier now.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -130,7 +130,6 @@ function amdify(source, subDependencyMapping) {
   let requirePath;
 
   source = source.replace(/\r\n/g, "\n");
-  source = source.replace(/\b(let|const)\b/g, "var");
 
   let outputSource = source;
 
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -130,7 +130,6 @@ function amdify(source, subDependencyMapping) {
   let requirePath;
 
   source = source.replace(/\r\n/g, "\n");
-  source = source.replace(/\b(let|const)\b/g, "var");
```
lilleyse, line 133: No reason to use `var` anymore especially now that CesiumJS has dropped support for IE11.

ebogo1: does this also apply to CesiumJS source code?

ebogo1 (commented):
Looks good to me; I confirmed this works with CesiumJS 1.84 and that all specs pass just in case. Some questions - 

> It uses a different version of prettier than CesiumJS so I couldn't remove it from .prettierignore.

You mentioned this in https://github.com/CesiumGS/cesium/pull/9706; is there a reason why the prettier versions are different?

gulpfile.js
lilleyse: That's for us to decide. The `GltfPipeline` folder in CesiumJS is very self-contained so it's a good testing ground for things like this.

lilleyse:
> You mentioned this in CesiumGS/cesium#9706; is there a reason why the prettier versions are different?

Some of our other repos are also using the latest version of prettier. I thought it would be best to use the latest here.

We're free to update the prettier version in CesiumJS as well, it just means we might need to do the whole pre-prettier and post-prettier tags against depending on the extent of changes.

PR #615, opened 2022-06-28
Use JavaScript Array.fill in place of Cesium.arrayFill

jjhembd:
Several library functions and tests are generating pre-filled arrays using the polyfill method Cesium.arrayFill. This polyfill is no longer needed, since the JavaScript method [Array.fill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill) is implemented on all supported platforms. See https://github.com/CesiumGS/cesium/pull/10488.

This PR removes the calls to the polyfill, and replaces them with calls to [Array.fill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ### Next release
 
 - Fixed handling of glTF models with the [`EXT_mesh_features`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features) extension.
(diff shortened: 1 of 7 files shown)


lilleyse:
Thanks @jjhembd. I removed the `CHANGES.md` entry since this change is completely internal and doesn't impact the public API. Otherwise looks great.

lilleyse:
Oh wait, looks like there's an `eslint` error. Could you fix that?

jjhembd:
Thanks @lilleyse! Sorry I missed the `eslint` error. It's fixed now.

PR #620, opened 2022-08-04
Validate accessor min/max values

javagl:
Details discussed in https://github.com/CesiumGS/gltf-pipeline/pull/619


lilleyse:
One side effect of putting this in `updateVersion` is it will slow down glTF 1.0 to 2.0 conversion in CesiumJS.

But that's probably not a big concern now that glTF 1.0 assets are relatively rare.

PR #621, opened 2022-08-29
Fix parsing mapping with # in the file name.

IvanLi-CN:
```bash
gltf-pipeline -i model.glb -o outputs/model.gltf -t
gltf-pipeline -o model-16.glb -i outputs/model.gltf -b
```

image

Fixed the problem that when the file name of a model mapping has a "#" symbol, the content after it is parsed as URL Hash, resulting in a missing file address.


leonchen:
I think the fix should replace all the `#` characters so it would either use `replaceAll` or regexp.

PR #631, opened 2023-03-15
Basic CESIUM_primitive_outline support

javagl:
The `removeUnusedElements` family of functions considered the accessors that are used by `CESIUM_primitive_outline` to be "unused", and removed them. 

This PR is a small fix that causes these accessors to be preserved, and the `CESIUM_primitive_outline.indices` to be updated (no, not the accessor contents, only that accessor index). 

This is implemented in a somewhat copy-and-paste-ish style: It uses the same approach that was used for other extensions. But one could consider to generalize the approach there: Having to support each extension manually like this does not scale indefinitely....

This might fix https://github.com/CesiumGS/gltf-pipeline/issues/630

### DIFF
--- a/lib/removeUnusedElements.js
+++ b/lib/removeUnusedElements.js
@@ -111,6 +111,15 @@ Remove.accessor = function (gltf, accessorId) {
       if (defined(indices) && indices > accessorId) {
         primitive.indices--;
       }
+
+      const ext = primitive.extensions;
+      if (
+        defined(ext) &&
+        defined(ext.CESIUM_primitive_outline) &&
+        ext.CESIUM_primitive_outline.indices > accessorId
+      ) {
+        --ext.CESIUM_primitive_outline.indices;
+      }
     });
   });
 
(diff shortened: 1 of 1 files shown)


javagl:
Attached here is a ZIP that contains some stuff that may be useful for testing:

- A simple unit cube with the `CESIUM_primitive_outline` extension
- The compressed versions of that file, **BEFORE** this fix
- The compressed versions of that file, **AFTER** this fix
- One `tileset*.json` for each of them
- A Sandcastle that allows selecting and loading them (causing crashes for the **BEFORE** ones, of course!)

javagl:
A random note: The README at https://github.com/CesiumGS/gltf-pipeline/blob/dc213115d2ca12bd5afae00770e48c019ac7d2d5/README.md#command-line-flags says

`--draco.quantizePositionBits`... No, default 14

but the actual default seems to be 11, as of https://github.com/CesiumGS/gltf-pipeline/blob/dc213115d2ca12bd5afae00770e48c019ac7d2d5/lib/compressDracoMeshes.js#L493 

(I can fix that in this PR, if desired)

lilleyse:
Thank @javagl. The doc updates look good too.

PR #635, opened 2023-03-30
Set resource directory in example in README.md

javagl:
The example in the README.md for converting a `.gltf` file to a `.glb` file assumes that the `.gltf` file does not have any associated resources (i.e. no references to buffers or textures). This is not generally true, and when trying to process arbitrary `.gltf` files, people seem to stumble over the _"Error: glTF model references separate files but no resourceDirectory is supplied"_.

Maybe updating the example will help.

(Yes, I also liked the simplicity that was suggested with the example in its original form, but ... there's a trade-off...)

Fixes https://github.com/CesiumGS/gltf-pipeline/issues/623

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,14 @@ npm install -g gltf-pipeline
 const gltfPipeline = require("gltf-pipeline");
 const fsExtra = require("fs-extra");
 const gltfToGlb = gltfPipeline.gltfToGlb;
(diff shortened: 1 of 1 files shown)


javagl:
> Will need to investigate the CI issue separately.

Sorry, that may indeed have been a missing `prettier` run on my side (I didn't see the build result before it was merged, and ... am not used to prettier also affecting `.MD` files...)

PR #638, opened 2023-05-03
Dependencies updates

ankittrehan2000:
Only dev dependency updates:
1. Removes `open` library from the `package.json` and `gulpfile` since it is now an ESM only package and isn't very useful
2. Updates `prettier` to latest version


mramato:
Thanks, @ankittrehan2000

PR #648, opened 2023-11-08
Update dependencies

ankittrehan2000:
update github actions to use node 20 and update lint staged to 15 (only major change is that it requries node 18 now to run)

Only outdated dependency was lint staged: 14 -> 15.0.2 (major change was support dropped for node 16 and 18 is the minimum requirement: https://github.com/lint-staged/lint-staged/releases/tag/v15.0.0)

No code changes


mramato:
Thanks.

PR #652, opened 2024-01-02
Update prettier

ankittrehan2000:
Updates prettier to latest version. 

The only library pinned here is `mime` because it requires ESM


ankittrehan2000:
Created a new label called `dependency-updates` and opened https://github.com/CesiumGS/gltf-pipeline/issues/653 to track any stuck dependencies

PR #672, opened 2025-05-31
Remove use of `defaultValue`

javagl:
I'll open this despite the open question of [What's the future of `gltf-pipeline`?](https://github.com/CesiumGS/gltf-pipeline/issues/670) (or rather, to prevent the short-term answer to that "being broken"):

The `defaultValue` function has been deprecated in CesiumJS, and is scheduled for removal in a few months. The `gltf-pipeline` library is stil used in several projects (with one of them being CesiumJS itself...). This PR removes the use of `defaultValue` throughout the codebase, and replaces it with the nullish coalescing operator. 

(Note: This will likely cause merge conflicts with https://github.com/CesiumGS/gltf-pipeline/pull/669 . Depending on the timeline, we _will_ have to fix the `defaultValue` deprecation _first_. In doubt, it should probably be possible to create a "replace-all"-regex for the changes here. I just did them manually, because they have _just so_ been below the threshold where an automation seemed worthwhile, but that's subjective).

### DIFF
--- a/bin/gltf-pipeline.js
+++ b/bin/gltf-pipeline.js
@@ -11,7 +11,6 @@ const gltfToGlb = require("../lib/gltfToGlb");
 const processGlb = require("../lib/processGlb");
 const processGltf = require("../lib/processGltf");
 
-const defaultValue = Cesium.defaultValue;
 const defined = Cesium.defined;
 
 const defaults = processGltf.defaults;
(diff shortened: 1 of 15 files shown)


javagl:
There are still the unknowns around https://github.com/CesiumGS/gltf-pipeline/issues/670

Given the interconnection between CesiumJS and `gltf-pipeline`, the concept of "versions" (or even "semantic versioning") don't have so much meaning here. Right now, users might declare
`"cesium": "^1.103.0",`
`"gltf-pipeline": "^4.1.0",`
(as, for example, in the `3d-tiles-tools`). The `gltf-pipeline` _itself_ declares
`"cesium": "^1.86.1",`
I think that this will cause trouble when Cesium 1.134 comes out (without `defaultValue`). And there will be further steps necessary to actually _prevent_ this trouble. But I think that what has been done in this PR will be _necessary_, regardless of the exact further steps that will be taken.

PR #678, opened 2026-01-14
Fix Unresolved Reference for KHR_materials_transmission textures

Apidcloud:
Hello!

Some of our users reported that transmission textures (specifically `KHR_materials_transmission.transmissionTexture`) disappear after we process the models with gltf-pipeline. More than stripping the used texture, it causes an unresolved reference error when trying to load it to e.g., gltf viewer. 

I used the same file for one of the unit tests, but I'm also attaching it here if you want to quickly try it in the viewer. Specifically, there's an `"index": 4` that is unresolved.
[unresolved-reference-gltf.zip](https://github.com/user-attachments/files/24621086/unresolved-reference-gltf.zip)

image

The issue was that `forEachTextureInMaterial` did not consider the `KHR_materials_transmission` extension when building the set of referenced textures, so `removeUnusedElements` treated those textures as unused and removed them.

I added a few unit tests that fail without the patch (both for `removeUnusedElementsSpec.js`, and for `processGltfSpec.js`).

Please let me know if I should add or change something else.

Thanks!


lilleyse:
Hey @Apidcloud! This looks good, but before we merge could you sign the [Cesium CLA](https://github.com/CesiumGS/community/tree/main/CLAs)?

Apidcloud:
Thanks for the prompt reply, @lilleyse! I signed both, since I use my personal account for github.

Apidcloud:
@lilleyse can you confirm they got signed properly?

lilleyse:
Yes, confirmed. Sorry for the delay!

Apidcloud:
Hello! Sorry to ping again, but would it be possible to make a release with this patch?

lilleyse:
@Apidcloud just published a new version that includes this fix, thanks!

https://www.npmjs.com/package/gltf-pipeline

PR #680, opened 2026-06-23
Add support for KHR_meshopt_compression

henrikedstrom:
Adds support for the [`KHR_meshopt_compression`](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_meshopt_compression) extension, alongside the existing `EXT_meshopt_compression` support.

- Add KHR handling everywhere EXT was already handled (`removeUnusedElements.js`, `mergeBuffers.js`).
- Add KHR test data and tests.
- Fix `Remove.buffer` to guard the meshopt buffer index decrement on `> bufferId`, matching the plain `bufferView.buffer` decrement.

Related issue: https://github.com/CesiumGS/cesium/issues/13526


lib/removeUnusedElements.js
```diff
@@ -153,11 +153,16 @@ Remove.buffer = function (gltf, bufferId) {
...
 
-    if (
-      defined(bufferView.extensions) &&
-      defined(bufferView.extensions.EXT_meshopt_compression)
-    ) {
-      bufferView.extensions.EXT_meshopt_compression.buffer--;
+    const extensions = bufferView.extensions;
+    if (defined(extensions)) {
+      const ext = extensions.EXT_meshopt_compression;
+      if (defined(ext) && ext.buffer > bufferId) {
```
henrikedstrom, line 159: Added guard here, also for the EXT case. See the `bufferView.buffer` decrement above it.

lilleyse: Ah, good catch!

lilleyse:
Looks good, thanks @henrikedstrom.

